Issue #3, opened 2022-03-30
Configure frontend build


DavidBiddle:
This should include anything we're likely to need in both forms-admin and forms-runner, but not anything more specific. We already have a default Rollup build but this may also need a but of configuring.

This should include:
- [x] babel for transpiling
- [x] a browserslist configuration (start with a sensible default, we will adjust this later once we have a better idea of our user stats)
- [x] jest for unit testing
- [x] cypress for end-to-end testing
- [x] style and js linting
- [x] autoprefixer
- [x] dart-sass instead of ruby-sass
- [x] install `govuk-frontend`
- [x] style and js formatting

PR #9, opened 2022-04-01
Frontend build setup

DavidBiddle:
Implements #3


danielburnley (approved):
Looks good to me 🎉

0atman (approved):
Nice :ok_hand:

danielburnley:
Hm - Actually having just used this for the deploy, we'll need the `rails assets:precompile` command to work. As far as I can tell if you're not using import maps (which we're not) you'll need to use Yarn rather than NPM which means we'll want a `yarn.lock` instead of `package-lock.json` i think? (i'm rusty on this 😅 )

DavidBiddle:
The issue with `assets:precompile` should now be fixed 🤞

danielburnley (approved):
Looks good to me 🎉

PR #15, opened 2022-04-08
Add README and CONTRIBUTING documentation

DavidBiddle:
Follows the [GDS Way's README guidance](https://gds-way.cloudapps.digital/manuals/readme-guidance.html) on structure.

Would particularly appreciate comments on the 'Getting started' install instructions - I've tested them and think they work, but there might be a better way of doing this.


danielburnley (approved):
Seems like a sensible starting point to me 🎉

PR #17, opened 2022-04-11
Add FE build check to CI test job

DavidBiddle:


danielburnley (approved):
Makes sense to me 🎉 worst case can move it out if we need to change things in the future

PR #22, opened 2022-04-25
Remove gulp-sourcemaps package

DavidBiddle:
Sourcemap functionality is actually built in to Gulp 4, so we don't need this package.


danielburnley (approved):
Always a fan of trimming down packages

Issue #24, opened 2022-05-03
End-to-end form spike: Admin UI


DavidBiddle:
- [x] Forms list
- [ ] Form create page

Part of [Spike: Allow user to create a form with a title that sends email to GOV.UK Notify](https://trello.com/c/xewAy5s7/123-spike-allow-user-to-create-a-form-with-a-title-that-sends-email-to-govuk-notify)

PR #30, opened 2022-05-11
Api integration

MeliaBurnley:

Progress towards #24 

This connects to the API using `ActiveResource` and creates the basic pages and controllers for integration with the API for creation and listing the forms

#### Checklist

- [ ] I've written unit tests for these changes (if code change) - Coming in a subsequent PR so that we can work on #24 
    - [x] README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ yarn build & yarn build:css
 | ------------- | ------------- |
 | `DATABASE_URL` | The URL to the postgres instance (without the database at the end) |
 | `SENTRY_DSN` | The DSN provided by Sentry |
-| `API_BASE` | The base url for the API - E.g. `http://localhost:9090/api/v1` |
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -3,4 +3,7 @@
 
   # Defines the root path route ("/")
   root "home#index"
+
+  get "/form/new", to: "forms#new"
+  post "/form/new", to: "forms#create"
```
DavidBiddle, line 8: Out of interest, is there a reason for setting these routes manually rather than using 
`resources :forms, only: [:new, :create]`?

DavidBiddle (approved):

config/routes.rb
danielburnley: hah, no particular reason, just ended up how i wrote it as I went 😄 can always change it as we go forward

PR #38, opened 2022-05-19
Move setup steps into Rails setup script

DavidBiddle:
By default Rails provides a setup file which can be used for initial setup and update tasks. Previouslty we weren't really using this, so this PR moves some of our setup tasks into this file. 
Tem members and contributors can now run `bin/setup` when first cloning the repo or when pulling changes to make sure that their dependencies, FE assets and database are up to date.

#### Checklist

    - [x] README.md


danielburnley (approved):
oh nice, looks good 🎉

PR #44, opened 2022-05-24
Test rails session with 'last seen' message

0atman:
Confirm our understanding of rails session cookies.

Screenshot from 2022-05-24 13-52-46

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -24,6 +24,7 @@ FileUtils.chdir APP_ROOT do
 
   puts "\n== Preparing database =="
   system! "bin/rails db:prepare"
+  system! "RAILS_ENV=test bin/rails db:setup"
 
   puts "\n== Removing old logs and tempfiles =="
   system! "bin/rails log:clear tmp:clear"
--- a/config/application.rb
+++ b/config/application.rb
@@ -20,5 +20,8 @@ module FormsAdmin
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
     #
+    config.session_store :redis_session_store, 
+      servers: ENV['REDIS_URL'],
+      key: '_app_session_key'
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,6 +1,7 @@
 require_relative "boot"
 
 require "rails/all"
+require "active_support/core_ext/integer/time"
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
--- a/config/application.rb
+++ b/config/application.rb
@@ -20,10 +20,5 @@ module FormsAdmin
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
     #
-
-    config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL", "redis://localhost:6379/1") }
-    config.public_file_server.headers = {
-      "Cache-Control" => "public, max-age=#{2.days.to_i}",
-    }
   end
 end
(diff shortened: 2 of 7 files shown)


bin/setup
```diff
@@ -24,6 +24,7 @@ FileUtils.chdir APP_ROOT do
 
   puts "\n== Preparing database =="
   system! "bin/rails db:prepare"
+  system! "RAILS_ENV=test bin/rails db:setup"
```
0atman, line 27: @danielburnley just checking this is the best way to dothis

danielburnley: I always forget `bin/setup` exists - seems like a sensible thing to do yea 👍🏻

config/application.rb
```diff
@@ -18,5 +19,11 @@ class Application < Rails::Application
     #
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+    #
+
+    config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL", "redis://localhost:6379/1") }
+    config.public_file_server.headers = {
+      "Cache-Control" => "public, max-age=#{2.days.to_i}",
```
0atman, line 26: So 2 days was the default, but obvs we could chose something more suitible

0atman:
@danielburnley so redis backend isn't too tricky, if we want to use it? I think we sidestep many issues by using it.

config/application.rb
danielburnley: yea we can change that easily enough

danielburnley:
@0atman yea that makes sense. I think you're right we'll probs side step a lot of issues, and at least the redis instance will be locked to gov paas. I'd be interested in quickly seeing if you can add some kinda user-id based encrypting of data on top to make it encrypted per user, and if you _really_ wanted we could try model moj forms and do per user per form (assign each form a random guid on creation and sign it with that too)

danielburnley:
@0atman you shouldn't need to worry about generating the session key itself - that's done automagically. I think we can handle the encrypting of user data (the value in the cache) separately in the future. For now just getting it to use a redis session store instead of the cookie based store should be all good 👍🏻

0atman:
```
$ redis-cli KEYS '*'
1) "b6f00f9c56e71b2b675c31dbc45a381c"
2) "b82e831c86a218bcd083155f9e583b5a"
```

@alice-carr and I confirmed that both our sessions were being stored in our shared redis :ok_hand:

config/environments/development.rb
```diff
@@ -17,21 +17,10 @@
...
-      "Cache-Control" => "public, max-age=#{2.days.to_i}",
-    }
-  else
-    config.action_controller.perform_caching = false
-
-    config.cache_store = :null_store
-  end
+  # config.session_store :cache_store, key: '_app_session_key'
+  config.session_store :redis_session_store, 
+    servers: 'redis://localhost:6379/1',
```
danielburnley, line 22: 👀 

I think this should be `ENV['REDIS_URL']`? though this is only in development rather than the application config

alice-carr: good catch, just fixed that

0atman:
@danielburnley thanks for picking this up. So while this does save the session in redis. It ALSO seems to still be saved in cookies.

danielburnley:
@alice-carr it looks like it's not doing cookie storage for me at least? I cleared out the cookies and i just get the session key stored in `_app_session_key` and if Redis is down it can't access the session

danielburnley:
It has just occurred to me that this is the admin, not the runner 😅 the admin probably doesn't need a session (for now?) but the runner will for user data. 

I'll repeat this work on the runner and reference this PR there

DavidBiddle:
@0atman Based on Daniel's comment above I feel like we can probably close this for now but let me know if you think otherwise

0atman:
ha!

Issue #43, opened 2022-05-24
Create initial form overview screen


danielburnley:
Both #41 and #42 require a form overview screen to link to, we can create a simple one to begin with that contains an edit link to our existing edit page (editing the title/submission email) 

As this will be iterated and changed in the future, we don't need to worry about making this too designed and can just be a simple link to the edit details page that exists already

To do
- [ ] Create form overview screen
- [ ] Form overview screen contains link to existing edit screen for "Edit form"

PR #45, opened 2022-05-26
Align form builder home screen with the prototype

### LINKED ISSUE #42: Align form builder home screen with the prototype
Trello card: https://trello.com/c/2mA8sYr5/262-create-form-builder-home-screen

Prototype screen: https://forms-prototypes.london.cloudapps.digital/form-designer/form-home

To do:

- [ ] Update form home screen create button to match the prototype
- [ ] Update form list display to match the prototype
  - Note: In Dev we can keep the link to the runner for now as a link underneath the form name or something similar
- [ ] Edit links take you to the form overview screen (depends on #43 )
- [ ] Create request and frontend specs for this

DavidBiddle:
Closes #42 

I've added a basic overview page so that I have something to point to - currently I'm using the show view for this and the edit view for what will probably eventually be the edit title page, I'm sure we'll want to change this eventually but i think it works for now.
 
I'm not entirely convinced the cypress tests I've written are adding much at the moment - I'm open to suggestions on better things to test here, but these will at least be enough to work with in our CI pipeline (see #46).

- [x] Update form home screen create button to match the prototype
- [x] Update form list display to match the prototype
  - Note: In Dev we can keep the link to the runner for now as a link underneath the form name or something similar
- [x] Edit links take you to the form overview screen (depends on #43 )
- [x] Create request and frontend specs for this 

#### Checklist

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -9,8 +9,6 @@ class FormsController < ApplicationController
 
     form.save!
 
-    flash[:message] = form.id
-
     flash[:message] = "Successfully created!"
     redirect_to action: "show", id: form.id
   rescue StandardError
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -9,11 +9,17 @@ class FormsController < ApplicationController
 
     form.save!
 
+    flash[:message] = form.id
+
     flash[:message] = "Successfully created!"
-    redirect_to :root
+    redirect_to action: "show", id: form.id
   rescue StandardError
     flash[:message] = "Unsuccessful"
     render :new
+  end
+
+  def show
+    @form = Form.find(params[:id])
   end
 
   def edit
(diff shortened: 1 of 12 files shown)


app/controllers/forms_controller.rb
```diff
@@ -9,13 +9,19 @@ def create
 
     form.save!
 
+    flash[:message] = form.id
```
danielburnley, line 12: is this overwritten on the following line?

danielburnley (approved):
Not sure if the line i commented on is intentionally left in, but seems good! Feel free to delete that line and merge if need be

It might be worth creating an issue for running the cypress tests in the pipeline and we can grab it at some point soon (I'm happy to try if you're unsure how to do it), but it can be separate to this PR

app/controllers/forms_controller.rb
DavidBiddle: Ah, that's a line I added for testing - have removed!

PR #47, opened 2022-05-26
Add cypress step to CI pipeline

### LINKED ISSUE #46: Add Cypress tests to CI pipeline
We should run the Cypress tests on every PR to to catch any issues early. Ideally we should add this to our existing CI workflow.

DavidBiddle:
Closes #46 

#### Checklist

    - [n/a] README.md


DavidBiddle:
Pausing this for now. The Cypress action itrself seems to work pretty smoothly, but I'm not currently able to get the rails server running in the action.
The error I currently see is: `RuntimeError: GDS-SSO running in mock mode and no test user found. Normally we'd load the first user in the database. Create a user in the database`, which is odd as the `db:seed` task  should be running (it's part of `db:prepare`, which is run as part of `bin/setup`.
Once the DB is working we'll also need to add something for the API - either access to the dev endpoint or a stub of some form.
This task may end up being quite long-running when we've got a few more tests, so we might want to consider running it in parallel with build-and-test task rather than as part of it.

thomasiles (approved):
Awesome!

PR #48, opened 2022-05-27
43 create initial form overview screen

### LINKED ISSUE #43: Create initial form overview screen

DavidBiddle:
Closes #43  
Includes an initial version of the overview page which is largely the same as the equivalent page in the prototype in style and structure terms. There are some slight tweaks:
- It doesn't contain an 'Add question' button yet, since this feature doesn't exist yet
- It does include delete button (i've moved this from the edit page) - we don't have a concept of deleting forms in the prototype yet
- It lists the submission email (this felt like a useful addition for the time being)

#### Checklist

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -41,7 +41,7 @@ class FormsController < ApplicationController
 
   def destroy
     form = Form.find(params[:id])
-    form.destroy!
+    form.destroy
 
     flash[:message] = "Successfully deleted #{form.name}"
     redirect_to root_path, status: :see_other
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -35,7 +35,7 @@ class FormsController < ApplicationController
     flash[:message] = "Successfully updated!"
     redirect_to action: "show", id: form.id
   rescue StandardError
-    flash[:message] = "Unsuccessful"
+    flash[:message] = "Update unsuccessful"
     redirect_to :edit_form, id: params[:id]
   end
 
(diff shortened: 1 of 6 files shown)


app/controllers/forms_controller.rb
```diff
@@ -35,19 +35,21 @@ def update
...
     redirect_to :edit_form, id: params[:id]
   end
 
   def destroy
     form = Form.find(params[:id])
-    form.destroy!
-
-    flash[:message] = "Successfully deleted #{form.name}"
-    redirect_to root_path, status: :see_other
+    if form.destroy
```
DavidBiddle, line 44: For some reason `form.destroy!` throws an undefined method error, and Rubocop doesn't like form.destroy without the if statement, which is why I've gone for this solution

danielburnley: Lol ok then - I would say it's cos ActiveResource probably doesn't define a `destroy!` method 🤔 which is because it does them all loud by default, but then rubocop will complain yeah cos it expects it to return true/false. Oof - we could disable that linting if it turns up again in the future with the reason that active resource doesn't work like that I guess

danielburnley (approved):
nice nice 🎉 looks good!

Issue #52, opened 2022-06-08
Create page for adding a question to a form


danielburnley:
We should be able to add a question page to the form. The page itself could be made ahead of the API work, however the page itself will not be able to be persisted until that work has been completed

- [x] form builder can add the question text
- [x] form builder can add the optional short name
- [x] form builder can optionally add hint text
- [x] form builder can select the sort of answer it is from list:
  - [x] single line of text
  - [x] address
  - [x] date
  - [x] email address
  - [x] National Insurance number 
  - [x] phone number
- [x] user can save this page
- [x] saves the question type as appropriate in the saved from structure
----
**Related issues**
https://github.com/alphagov/forms-api/issues/25

PR #54, opened 2022-06-09
Add page models to forms

danielburnley:

This is some work to prepare for what will be needed in #52 alongside developing the API. I added these to check that that API works as expected and figured it was worth committing them as I was here.

**Note**

People will need to change the `API_BASE` values as a result of this change to remove the `api` from the route

#### Checklist

    - [x] README.md


DavidBiddle (approved):
Looks good 🎉

PR #61, opened 2022-06-23
Set next page parameter

danielburnley:

We need to set the next page param when new pages are created, this is for the runner to be able to navigate from page to page and in the future to allow for branching and page re-ordering

#### Checklist

- **N/A** I've updated the documentation in (If any documentation requires updating)


DavidBiddle (approved):
Looks good to me. I think that `previous_last_page` function will come in handy when we add page deletion too

PR #62, opened 2022-06-27
Add initial user organisation code

0atman:
Also driveby update ruby version

Co-authored-by: alice-carr <alice-carr@users.noreply.github.com>

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


danielburnley:
This will need the github action updating FYI

app/controllers/forms_controller.rb
```diff
@@ -7,6 +7,7 @@ def create
     form = Form.new({
       name: params[:name],
       submission_email: params[:submission_email],
+      org: current_user.organisation_slug,
```
danielburnley, line 10: Just a sense check that we want to use org here rather than "organisation" - either works for me just wanting to make sure we're cool with it being a shortened version

danielburnley (commented):
I can approve this once we've fixed the github actions if this is where you want this to sit to begin with - There'll want to be a follow up where the forms list on the home page no longer just queries for "all" and instead queries for all scoped to the organisation, which can just be a get param on the forms endpoint

danielburnley:
Also before you merge this please make sure you let the team know in slack because it requires a ruby version update

PR #65, opened 2022-07-13
Add org parameter to form list request

DavidBiddle:
Adds the user's organisation to the forms request on the homepage. Combined with the [corresponding API PR by Tris](https://github.com/alphagov/forms-api/pull/39#issuecomment-1181958019) this will result in the user only seeing forms from their organisation.

I've edited the DB seed script to include the user's `organisation_slug`, but this won't work for existing local users, you'll have to update or destroy your existing user manually. I can add something to automate this if people think it's necessary.

#### Checklist


DavidBiddle:
Security lint is failing because activerecord has a vulnerability - will be resolved by [bumping to 7.0.3.1](https://github.com/alphagov/forms-admin/pull/66)

thomasiles (approved):

PR #86, opened 2022-08-10
Switch summary list markup to govuk-component

aliuk2012:

`govuk-component` gem provides us with a summary list component.
Theres a couple of benefits here in that the govuk-component version
should always match the implementation of govuk-frontend nunjucks version.
This means that we don't have to worry about manually updating our version
if govuk-frontend updates their components and ends up breaking our one.

Apart for the key values being made bold and the actions being on one line,
I think govuk-component implementation does a better job with the actions
because now the screenreader would alert the user to the fact that there
is a list of actions rather than not realising that there is more actions
without navigation through the page.

#### Before 
Screenshot 2022-08-10 at 08 52 33

#### After
Screenshot 2022-08-10 at 08 51 58

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/views/home/index.html.erb
+++ b/app/views/home/index.html.erb
@@ -10,18 +10,13 @@
 <%= govuk_start_button(text: "Create a form", href: new_form_path) %>
 
 <h2 class="govuk-heading-m">Your forms</h2>
-<dl class="govuk-summary-list">
-  <% @forms.each do |form| %>
-    <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key govuk-!-width-full govuk-!-font-weight-regular">
-        <p class="govuk-body"><%= form.name %></p>
-      </dt>
-      <dd class="govuk-summary-list__actions">
-        <%= link_to form_path(form.id), class: "govuk-link govuk-link--no-visited-state" do %>
-          Edit <span class="govuk-visually-hidden"> <%= form.name %></span>
-        <% end %><br>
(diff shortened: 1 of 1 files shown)


app/views/home/index.html.erb
```diff
@@ -10,6 +10,7 @@
 <%= govuk_start_button(text: "Create a form", href: new_form_path) %>
 
 <h2 class="govuk-heading-m">Your forms</h2>
+
```
DavidBiddle, line 13: Doesn't look like this actually includes the summary list change - think maybe something has gone wrong with a history rewrite?

cypress/e2e/home.cy.js
```diff
@@ -9,7 +9,7 @@ describe('Home screen', function () {
   })
 
   it('allows the user to edit an existing form', function () {
-    cy.contains('Edit').click()
+    cy.contains('edit').click()
```
DavidBiddle, line 12: I think we probably want to keep the Edit / View in Runner links capitalised but can check with Hannah if we're not sure

app/views/home/index.html.erb
aliuk2012: 🤦  Yeah, I had commited the reverted changes to take screenshots. Should now see the changes

DavidBiddle:
Nice one - just the one comment from me about the change of case on these action links, but otherwise looks really good and seems like a sensible change!

DavidBiddle (approved):

cypress/e2e/home.cy.js
aliuk2012: Good suggestion about the copy. I had a chat with Hannah and I've reverted those changes.

PR #104, opened 2022-08-18
Add tasklist

thomasiles:
# WIP: add new tasklist component
Add basic tasklist component.

This is a work in progress and the code shouldn't pass any tests - I've
basically plonked a new component right where the previous options are
when a user creates a form. To complete this, we would need to make a
new page for adding and editing pages, to match the prototype journey.

A bare bones tasklist component has been added to get things started
using view_components.

The original plan was to take the code for a tasklist component straight
from the DfE project at:

git@github.com:DFE-Digital/register-trainee-teachers.git

This didn't quite work out:

- there have been changes to the `govuk-components` and
  `view_components` gems which stopped my copying it directly.
- That project uses webpack and webpacker, which means the way we
  include CSS/JS is slightly different.
- There are differences in the design of the tasklist which have to be
  accounted for; we have sections and numbers.

The code from that project was used as the starting point and there are
many pieces of it still here.

There are some decisions to be made before the rest of the work to
complete the tasklist is done.

- The scss and js would be included directly in
  `app/assets/stylesheets/application.scss` and
  `app/javascript/application.js` for it to be included in the bundle.

  The DfE project referenced could do this using webpack configuration.
  I don't think moving to webpack would be worth it just for this
  convenience. There might be other ways we could achieve it but I
  think I could live with listing components directly. Is there anything
  else we miss by not having this? I'm open to suggestions.

- Does the layout make sense? There seems to be a few ways to structure
  view_components - sticking them in their own folders (sidecar?) or
  having single files all in a single directory. I think there are other
  questions, what is worth making into a component? when to use inner
  classes/procs? should we use helpers? can we call the tasklist with an
  array of options etc.

- Is there an easy way to copy components from other projects - maybe it
  would be worth matching other projects config just to re-use their
  work?

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


thomasiles:
I've rebased and built another branch as this one has fallen behind and the imlpementation changed.

PR #105, opened 2022-08-19
Create a new pages index page

aliuk2012:
This new page is needed because the pages list on the form show
page will no longer be shown. Instead on the forms show page
there will be a single link to take the user to the list of pages
where they can be created/edited.
#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/views/forms/show.html.erb
+++ b/app/views/forms/show.html.erb
@@ -11,8 +11,6 @@
       <p class="govuk-body"><%= flash[:message] %></p>
     <% end %>
 
-    <%= link_to t("forms.form_overview.add_question"), new_page_path(@form), class: "govuk-button govuk-!-margin-bottom-3 govuk-!-margin-top-3", 'data-module':"govuk-button", draggable:"false", role: "button" %>
-
     <h2 class="govuk-heading-m govuk-!-margin-top-7 govuk-!-margin-bottom-0">
       <%= t("forms.form_overview.form_name") %>
     </h2>
@@ -30,23 +28,17 @@
       </div>
     </dl>
 
-    <% if @pages.any? %>
-      <h2 class="govuk-heading-m govuk-!-margin-top-7 govuk-!-margin-bottom-0"><%= t("forms.form_overview.your_questions") %></h2>
-      <% @pages.each do |page| %>
-        <dl class="govuk-summary-list">
-          <div class="govuk-summary-list__row">
-            <dt class="govuk-summary-list__key govuk-!-font-weight-regular">
-              <%= page.question_text %>
-            </dt>
-            <dd class="govuk-summary-list__actions">
-              <%= link_to edit_page_path(@form,page), class: "govuk-link govuk-link--no-visited-state" do %>
-              <%= t("forms.form_overview.edit") %> <span class="govuk-visually-hidden"><%= page.question_text %></span>
-              <% end %>
-            </dd>
-          </div>
-        </dl>
-      <% end %>
-    <% end %>
+    <h2 class="govuk-heading-m govuk-!-margin-top-7 govuk-!-margin-bottom-0"><%= t("forms.form_overview.your_questions") %></h2>
+
+    <dl class="govuk-summary-list">
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key govuk-!-font-weight-regular">
+          <%= govuk_link_to t("forms.form_overview.add_or_edit_questions"), form_pages_path(@form) %>
+        </dt>
+
+      </div>
+    </dl>
+
 
     <h2 class="govuk-heading-m govuk-!-margin-top-7 govuk-!-margin-bottom-0">
       Submission email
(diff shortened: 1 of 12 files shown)


app/views/forms/show.html.erb
```diff
@@ -30,23 +28,17 @@
...
-          </div>
-        </dl>
-      <% end %>
-    <% end %>
+    <h2 class="govuk-heading-m govuk-!-margin-top-7 govuk-!-margin-bottom-0"><%= t("forms.form_overview.your_questions") %></h2>
+
+    <dl class="govuk-summary-list">
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key govuk-!-font-weight-regular">
+          <%= govuk_link_to t("forms.form_overview.add_or_edit_questions"), form_pages_path(@form) %>
```
DavidBiddle, line 36: So app/views/forms/show.html.erb is going to become the task list right?

(Not a blocking comment, just making sure I understand properly!)

app/views/pages/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-l"><%= @form.name %></span>
+    <h1 class="govuk-heading-l"><%= t("pages.index.title") %></h1>
+
+    <% if flash[:message] %>
+      <p class="govuk-body"><%= flash[:message] %></p>
+    <% end %>
+
+    <%= link_to t("pages.index.add_question"), new_page_path(@form), class: "govuk-button govuk-!-margin-bottom-3 govuk-!-margin-top-3", 'data-module':"govuk-button", draggable:"false", role: "button" %>
```
DavidBiddle, line 14: Could this be a `govuk_button_link_to`?
e.g.
```suggestion
    <%= govuk_button_link_to t("pages.index.add_question"), new_page_path(@form), class: "govuk-!-margin-bottom-3 govuk-!-margin-top-3"%>
```

app/views/pages/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    <% if @pages.any? %>
+      <h2 class="govuk-heading-m govuk-!-margin-top-7 govuk-!-margin-bottom-0"><%= t("forms.form_overview.your_questions") %></h2>
+      <% @pages.each do |page| %>
+        <dl class="govuk-summary-list">
+          <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key govuk-!-font-weight-regular">
+              <%= page.question_text %>
+            </dt>
+            <dd class="govuk-summary-list__actions">
+              <%= link_to edit_page_path(@form,page), class: "govuk-link govuk-link--no-visited-state" do %>
```
DavidBiddle, line 25: Similar to above - think this could be a `govuk_link_to`:
```suggestion
              <%= govuk_link_to edit_page_path(@form,page), no_visited_state: true do %>
```

cypress/e2e/edit-page.cy.js
```diff
@@ -10,8 +10,8 @@ describe('Edit a form page', function () {
       .should('have.value', 'What is your work address?')
   })
 
-  it('allows the user to navigate back to the form overview page', function () {
-    cy.contains('Go to form overview').click()
-    cy.url().should('eq', `${Cypress.config().baseUrl}/forms/${this.formId}`)
+  it('allows the user to navigate back to the forms  page', function () {
```
DavidBiddle, line 13: Might help to name this the same as above for consistency:
```suggestion
  it('allows the user to navigate back to the form pages', function () {
```

DavidBiddle:
Looks good - raised a couple of comments where I think you could use the govuk-components stuff (I think in code i originally wrote! 😁) and one minor test naming thing.

app/views/forms/show.html.erb
aliuk2012: yeah that is correct

app/views/pages/index.html.erb
aliuk2012: 🤔 why do we want to turn off the visited state, is that not an accessibility issue if thats done because users might not realise that they have already visited that page before

DavidBiddle: Bit of a grey area - it's based on this [design system guidance on links](https://design-system.service.gov.uk/styles/typography/#links):

> Use the govuk-link--no-visited-state modifier class where it is not helpful to distinguish between visited and unvisited states, for example when linking to pages with frequently-changing content such as the dashboard for an admin interface.

I think this is in there because it's content that will potentially change pretty frequently as a form goes through review processes etc, so the state a user has previously seen it in won't necessarily be the current state. But I think that you're probably right that we should get rid of the `no_visited_state: true` - if we find that users are getting conused, we can change it back, but it makes sense to keep the visited state by default.

aliuk2012: Design/Content [decided](https://gds.slack.com/archives/C030K5706TX/p1661155640382409) we should keep the default link behaviour.

app/views/pages/index.html.erb
aliuk2012: 👍

DavidBiddle (approved):
Looks good!

PR #113, opened 2022-08-31
Add make live

thomasiles:
#### Add a live journey to the admin

In order to make forms live, a new route /make_live has been added. This shows a confirmation screen with some information and allows the form creator to set the live_at property of forms. 

If the form is already live, the confirmation screen is skipped and the user is redirected to the live_screen. This should allow them to get the URL for the service if the form is already live, if they were to hit this route. There are other options here - e.g. 404, but this seems reasonable.

Relates to: https://trello.com/c/rcu9p97d/366-add-new-designs-to-forms-admin-that-fake-making-a-form-live-and-getting-the-link-to-the-already-live-form-should-have 

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/views/forms/make_live/confirmation.html.erb
+++ b/app/views/forms/make_live/confirmation.html.erb
@@ -1,0 +1,1 @@
+<h1 class="govuk-heading-xl">its live!</h1>
--- a/app/views/forms/make_live/confirmation.html.erb
+++ b/app/views/forms/make_live/confirmation.html.erb
@@ -1,7 +1,9 @@
+<% set_page_title(t('page_titles.your_form_is_live')) %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <span class="govuk-caption-l"><%= @form.name %></span>
-    <h1 class="govuk-heading-l">Your form is live</h1>
+    <h1 class="govuk-heading-l"><%= t('page_titles.your_form_is_live') %></h1>
 
     <p class="govuk-body">
       The URL for your form is:
@@ -11,13 +13,10 @@
       </span>
     </p>
 
-    <p class="govuk-body">
-      The form will not be indexed by search engines, so people will not be able to find it easily.
-      Contact your GOV.UK publishing team to publish a link to your form on GOV.UK so people can find it.
-    </p>
+    <%= t("make_live.confirmation.body_html") %>
 
     <p class="govuk-body">
-      <%= govuk_link_to "Continue to form details", form_path(@form.id) %>
+      <%= govuk_link_to t("make_live.confirmation.continue_link"), form_path(@form.id) %>
     </p>
   </div>
 </div>
--- a/app/views/forms/make_live/confirmation.html.erb
+++ b/app/views/forms/make_live/confirmation.html.erb
@@ -7,7 +7,7 @@
       The URL for your form is:
 
       <span class="govuk-!-display-block govuk-!-margin-top-6 govuk-!-padding-bottom-6">
-        <%= govuk_link_to "#{ENV['RUNNER_BASE']}/form/#{@form.id}","#{ENV['RUNNER_BASE']}/form/#{@form.id}", { target:"_blank", rel:"noopener noreferrer"} %>
+        <%= govuk_link_to link_to_runner(ENV["RUNNER_BASE"], @form.id ), link_to_runner(ENV["RUNNER_BASE"], @form.id ), { target:"_blank", rel:"noopener noreferrer"} %>
       </span>
     </p>
 
--- a/app/views/forms/make_live/confirmation.html.erb
+++ b/app/views/forms/make_live/confirmation.html.erb
@@ -1,1 +1,23 @@
-<h1 class="govuk-heading-xl">its live!</h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-l"><%= @form.name %></span>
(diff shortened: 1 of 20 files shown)


app/views/forms/make_live/confirmation.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-l"><%= @form.name %></span>
+    <h1 class="govuk-heading-l">Your form is live</h1>
+
+    <p class="govuk-body">
+      The URL for your form is:
+
+      <span class="govuk-!-display-block govuk-!-margin-top-6 govuk-!-padding-bottom-6">
+        <%= govuk_link_to "#{ENV['RUNNER_BASE']}/form/#{@form.id}","#{ENV['RUNNER_BASE']}/form/#{@form.id}", { target:"_blank", rel:"noopener noreferrer"} %>
```
DavidBiddle, line 10: You could optionally consider putting `#{ENV['RUNNER_BASE']}/form/#{@form.id}` in a variable somewhere to avoid the repetition.

app/views/forms/make_live/new.html.erb
```diff
@@ -0,0 +1,52 @@
...
+        Make your form live
+      </h1>
+
+      <p class="govuk-body">
+        When you make your form live you will get a URL for the form. The form
+        will not be indexed by search engines, so people will not be able to
+        find it easily. Contact your GOV.UK publishing team to publish a link to
+        your form on GOV.UK so people can find it.
+      </p>
+      <p></p>
```
DavidBiddle, line 24: Think we should delete this empty paragraph

app/views/forms/make_live/new.html.erb
```diff
@@ -0,0 +1,52 @@
...
+    <%= form_with(model: @make_live_form, url: make_live_create_path) do |f| %>
+      <% if @make_live_form&.errors&.any? %>
+        <%= f.govuk_error_summary %>
+      <% end %>
+
+
+      <h1 class="govuk-heading-l">
+        <span class="govuk-caption-l"><%= @make_live_form.form.name %></span>
+        Make your form live
+      </h1>
```
DavidBiddle, line 16: Elsewhere we put the caption outside the heading, think we should probably do the same here, i.e.:
```suggestion
      <span class="govuk-caption-l"><%= @make_live_form.form.name %></span>
      <h1 class="govuk-heading-l">Make your form live</h1>
```

app/views/forms/make_live/confirmation.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="govuk-grid-row">
```
DavidBiddle, line 1: Should this file and `new.html.erb` be localised? I think we're still using Rails i18n everywhere else

app/views/forms/make_live/new.html.erb
aliuk2012: At the moment the sub-heading is outside the h1 so for screenreaders they wouldn’t have context as to what the particular page relates too

> If the caption should be considered part of the page heading, you can also nest the caption within the <h1>.


https://design-system.service.gov.uk/styles/typography/#headings-with-captions

I had a chat with @thomasiles about this earlier in the week and we have updated the main form overview page already

DavidBiddle: Makes sense!

app/views/forms/make_live/new.html.erb
aliuk2012: Good spot

app/views/forms/make_live/confirmation.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <span class="govuk-caption-l"><%= @form.name %></span>
+    <h1 class="govuk-heading-l"><%= t('page_titles.your_form_is_live') %></h1>
+
+    <p class="govuk-body">
+      The URL for your form is:
+
+      <span class="govuk-!-display-block govuk-!-margin-top-6 govuk-!-padding-bottom-6">
+        <%= govuk_link_to link_to_runner(ENV["RUNNER_BASE"], @form.id ), link_to_runner(ENV["RUNNER_BASE"], @form.id ), { target:"_blank", rel:"noopener noreferrer"} %>
+      </span>
+    </p>
```
aliuk2012, line 14: I'm planning on stripping this out into a proper view component in another piece of work where we will be playing back this content on other pages.

DavidBiddle: Good idea!

config/locales/forms/make_live.yml
```diff
@@ -0,0 +1,49 @@
...
+        <p class="govuk-notification-banner__heading">
+          Any changes you make to a live form will be updated in the form
+          immediately.
+        </p>
+        <p class="govuk-body">
+          This could have an impact on people who are filling in the form at
+          the same time. They may lose any answers they have already provided
+          and may need to start again.
+        </p>
+  helpers:
```
DavidBiddle, line 37: Think this is what's causing your test failure - for reasons I don't understand `i18n-tasks` doesn't like it when you have your own translations and the `helpers`/`activemodel` translations the govuk components gem uses in the same file. Should be resolved by moving either your translations or the activemodel ones into a separate file

aliuk2012: fixed these now :-)

DavidBiddle (approved):

PR #114, opened 2022-08-31
Tweak preview links 

aliuk2012:
Relates to https://trello.com/c/CrfvO2xj/70-make-it-easier-to-preview-view-the-form-youre-making-in-admin
#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/views/forms/show.html.erb
+++ b/app/views/forms/show.html.erb
@@ -4,8 +4,10 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <span class="govuk-caption-l"><%= @form.name %></span>
-    <h1 class="govuk-heading-l"><%= t("forms.form_overview.title") %></h1>
+    <h1 class="govuk-heading-l">
+      <span class="govuk-caption-l"><%= @form.name %></span><span class="govuk-visually-hidden"> - </span>
+      <%= t("forms.form_overview.title") %>
+    </h1>
 
     <% if flash[:message] %>
       <p class="govuk-body"><%= flash[:message] %></p>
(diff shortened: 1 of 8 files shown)


app/views/forms/show.html.erb
```diff
@@ -4,8 +4,10 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <span class="govuk-caption-l"><%= @form.name %></span>
-    <h1 class="govuk-heading-l"><%= t("forms.form_overview.title") %></h1>
+    <h1 class="govuk-heading-l">
+      <span class="govuk-caption-l"><%= @form.name %></span><span class="govuk-visually-hidden"> - </span>
```
thomasiles, line 8: nice!

spec/views/forms/show.html.erb_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+describe "forms/show.html.erb" do
+  around do |example|
+    ClimateControl.modify RUNNER_BASE: "api-host" do
```
thomasiles, line 5: I guess api-host might be better as runner-host in this case? not suggesting changing it. Would it be worth adding a helper or config set by the env var to keep setting the value in one place in teh code somewhere? Don't know if the benefits outway the cost, just a thought

thomasiles (approved):

spec/views/forms/show.html.erb_spec.rb
aliuk2012: 🤦  good catch, should have been `runner-host` not `api-host`

PR #123, opened 2022-09-09
Live form UI changes

DavidBiddle:
Adds various conditional UI elements when the form is live:
- Adds live form warning to the add questions page
- Adds live form URL to the task list page
- Changes the heading on the task list page to 'Your form'

Trello card: https://trello.com/c/8o2Z1IBf/1-update-forms-admin-with-new-designs-for-editing-a-live-form-should-have

#### Checklist

    - [n/a] README.md

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -29,8 +29,4 @@ module ApplicationHelper
   def link_to_runner(runner_url, form_id)
     "#{runner_url}/form/#{form_id}"
   end
-
-  def form_is_live(form)
-    form.live_at.present? && form.live_at < Time.zone.now
-  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -29,4 +29,8 @@ module ApplicationHelper
   def link_to_runner(runner_url, form_id)
     "#{runner_url}/form/#{form_id}"
   end
+
+  def form_is_live(form)
+    form.live_at.present? && form.live_at < Time.zone.now
+  end
 end
--- a/app/views/forms/show.html.erb
+++ b/app/views/forms/show.html.erb
@@ -6,14 +6,14 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-l">
       <span class="govuk-caption-l"><%= @form.name %></span><span class="govuk-visually-hidden"> - </span>
-      <%= @form_is_live ? t("forms.form_overview.live_title") : t("forms.form_overview.title") %>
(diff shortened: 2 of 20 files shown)


app/helpers/application_helper.rb
```diff
@@ -29,4 +29,8 @@ def govuk_back_link_to(url = :back, body = "Back")
   def link_to_runner(runner_url, form_id)
     "#{runner_url}/form/#{form_id}"
   end
+
+  def form_is_live(form)
+    form.live_at.present? && form.live_at < Time.zone.now
+  end
```
aliuk2012, line 35: I don't think this is needed we already have this method on the model https://github.com/alphagov/forms-admin/blob/main/app/models/form.rb#L14. 

Perhaps for the time being we dont have to check the live_at is in the past. If you need we should have it to save future work then could you make the changes to the model method?

app/views/forms/show.html.erb
```diff
@@ -6,13 +6,22 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-l">
       <span class="govuk-caption-l"><%= @form.name %></span><span class="govuk-visually-hidden"> - </span>
-      <%= t("forms.form_overview.title") %>
+      <%= form_is_live(@form) ? t("forms.form_overview.live_title") : t("forms.form_overview.title") %>
```
aliuk2012, line 9:
```suggestion
      <%= @form.live? ? t("forms.form_overview.live_title") : t("forms.form_overview.title") %>
```

app/views/forms/show.html.erb
```diff
@@ -6,13 +6,22 @@
...
     </h1>
 
     <% if flash[:message] %>
       <p class="govuk-body"><%= flash[:message] %></p>
     <% end %>
 
+    <% if form_is_live(@form) %>
```
aliuk2012, line 16:
```suggestion
    <% if @form.live? %>
```

app/views/pages/index.html.erb
```diff
@@ -4,6 +4,11 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
+
+    <% if form_is_live(@form) %>
```
aliuk2012, line 8:
```suggestion
    <% if @form.live? %>
```

app/helpers/application_helper.rb
DavidBiddle: You're right - will get rid of the helper and move the check to the model. I think it's worth keeping the time check, easy to remove if we need to but also seems like a sensible bit of future-proofing / guard against weird stuff happening with our data.

aliuk2012 (approved):
:shipit:

PR #125, opened 2022-09-15
Stop page titles being double html escaped

thomasiles:
See https://trello.com/c/vBKqVumH/127-bug-title-text-isnt-properly-encoded

When editing pages, the title of the page uses question_text. When the question_text contains some characters, such as ', rails escapes them twice, which produces titles with hex codes in.

This commit stops that happening and adds a view test to the edit page view. There are other tests which could be added - testing the title helper etc. but the view test seemed like the most reliable way to test the escaping behaviour. This means rendering using the application layout, which breaks out of the unit testing model a bit.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -7,7 +7,7 @@ module ApplicationHelper
   end
 
   def set_page_title(title)
-    content_for(:title) { title }
+    content_for(:title) { title.html_safe }
   end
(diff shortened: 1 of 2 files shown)


DavidBiddle (approved):
Nice, looks good! 🎉

PR #127, opened 2022-09-16
Remove validation for "What happens next" to make it optional

aliuk2012:
We have decided that "What happens next" should be optional for the time being. We will need to get users to manually add in this information to any existing forms. By removing this validation we also enable them to "delete/clear" a previous saved content relating to what happens next if they decide not to have any copy at all.

Related to https://trello.com/c/4mCMqpda/356-add-what-happens-next-to-the-confirmation-page-should-have
#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/forms/forms/what_happens_next_form.rb
+++ b/app/forms/forms/what_happens_next_form.rb
@@ -4,7 +4,7 @@ class Forms::WhatHappensNextForm
 
   attr_accessor :form, :what_happens_next_text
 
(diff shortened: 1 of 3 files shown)


danworth (approved):
Looks good to me, removes the mandatory condition but keeps the char limit🥇

thomasiles (approved):
Looks good 👍

PR #129, opened 2022-09-20
Allow runner URL helper to create live links and add copy to clipboard button

DavidBiddle:
- Adds the correct live URL to the form overview and the make live confirmation 
- Makes the live form URL plain text, not a hyperlink (we want people to copy it, rather than click on it)
- Adds a button to copy the URL to the clipboard

Trello card: https://trello.com/c/Jdrh6XZ6/154-bug-preview-link-shown-in-live-form-url-component

#### Checklist

    - [n/a] README.md
    - [x] Elsewhere (marked up new JS functions with JSDoc syntax)

### DIFF
--- a/app/components/form_url_component/view.html.erb
+++ b/app/components/form_url_component/view.html.erb
@@ -1,7 +1,9 @@
-<p class="govuk-body">
-  <%= t("form_url_component.url_is") %>
+<div data-module="copy-to-clipboard" data-copy-button-text="Copy URL to clipboard">
+  <p class="govuk-body">
+    <%= t("form_url_component.url_is") %>
+  </p>
 
-  <span class="govuk-!-display-block govuk-!-margin-top-6 govuk-!-padding-bottom-6">
-    <%= govuk_link_to @runner_link, @runner_link, { target:"_blank", rel:"noopener noreferrer"} %>
-  </span>
-</p>
+  <p class="govuk-body" data-copy-target>
+    <strong><%= @runner_link %><strong>
+  </p>
+</div>
--- a/app/components/form_url_component/view.html.erb
+++ b/app/components/form_url_component/view.html.erb
@@ -1,4 +1,4 @@
-<div data-module="copy-to-clipboard" data-copy-button-text="Copy URL to clipboard">
+<%= tag.div "data-module": "copy-to-clipboard", "data-copy-button-text": t("form_url_component.copy_to_clipboard") do %>
   <p class="govuk-body">
     <%= t("form_url_component.url_is") %>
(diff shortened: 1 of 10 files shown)


app/helpers/application_helper.rb
```diff
@@ -26,7 +26,8 @@ def govuk_back_link_to(url = :back, body = "Back")
     )
   end
 
-  def link_to_runner(runner_url, form_id)
-    "#{runner_url}/preview-form/#{form_id}"
+  def link_to_runner(runner_url, form_id, live = false)
+    mode_segment = live ? "form" : "preview-form"
+    "#{runner_url}/#{mode_segment}/#{form_id}"
   end
```
aliuk2012, line 32: I was thinking maybe pass in the form and let the helper get access to the info it needs to work out what link to use
```suggestion
  def link_to_runner(runner_url, form)
    mode_segment = "preview-" if form.live?
    "#{runner_url}/#{mode_segment}form/#{form.id}"
  end
```

aliuk2012: 🤦  sorry meant to use the invert of `.live?`

DavidBiddle: Yeah, that could work and it's quite neat. I suppose what we do depends on what we expect the various 'Preview this form' links to do for a live form - should they go to the live form URL or to the preview version?

DavidBiddle: @aliuk2012 to be clearer on that - I think if we do it by inverting .live, then all of the 'Preview this form' links will point to the live URL rather than the preview. whereas if we pass a 'live' boolean to the helper we can pick and choose which ones point to live routes and which point to the preview links. I'm not actually sure which of those we want here

app/components/form_url_component/view.html.erb
```diff
@@ -1,7 +1,10 @@
...
+  </p>
 
-  <span class="govuk-!-display-block govuk-!-margin-top-6 govuk-!-padding-bottom-6">
-    <%= govuk_link_to @runner_link, @runner_link, { target:"_blank", rel:"noopener noreferrer"} %>
-  </span>
-</p>
+  <p class="govuk-body" data-copy-target>
+    <strong><%= @runner_link %><strong>
+  </p>
+</div>
```
aliuk2012, line 9:
```suggestion
```

aliuk2012: Stray closing div tag

aliuk2012 (approved):
Great work, @DavidBiddle. 
Only one stray div in the markup but the rest looks ace.

PR #133, opened 2022-09-23
Form status tag

aliuk2012:

- Add new component for the form status tag
- Add form status tag to the forms details page & questions page
- Add form status to the list of forms (Home page) and removes the edit/preview links on the home page

Refers to https://trello.com/c/csc7PAeM/4-add-draft-or-live-tags-to-forms-admin-on-create-or-edit-a-form-page-and-add-or-edit-questions-should-have

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


app/views/forms/show.html.erb
```diff
@@ -7,8 +7,11 @@
     <h1 class="govuk-heading-l">
       <span class="govuk-caption-l"><%= @form.name %></span><span class="govuk-visually-hidden"> - </span>
       <%= @form.live? ? t("forms.form_overview.live_title") : t("forms.form_overview.title") %>
+      <span class="govuk-!-display-block"><%= render FormStatusTagComponent::View.new(status: @form.status) %></span>
```
DavidBiddle, line 10: Not sure this should be inside the H1

DavidBiddle, line 10: You may have already tried this and found an issue, but instead of styling a `span` into a block element we could use a `div`here

aliuk2012: @DavidBiddle would a div inside a `h1` be valid markup?  https://caninclude.glitch.me/caninclude?child=div&parent=h1

aliuk2012: TBH, i think these headings could be converted to proper view components to keep them consistent.  I placed it inside to give it a bit more context otherwise its kind of a floating "LIVE"/"DRAFT" in the middle of no where with no context

DavidBiddle: Yeah, see what you mean about wanting to give it some context, I just worry that we're starting to overstuff these H1s a bit.

Wonder if something like this would work:
```erb
<h1><%= whatever_h1_content %></h1>
<dl>
    <dt class="govuk-visually-hidden">Status</dt>
    <dd><%= render FormStatusTagComponent::View.new(status: @form.status) %></dd>
</dl>
```

That `dl` approach would be consistent with how Content Publisher and Whitehall mark up statuses, and we could probably add more metadata to that area in future if the design called for it.

DavidBiddle: This is moot now anyway 😄

PR #144, opened 2022-09-26
Only show preview links on forms with pages

DavidBiddle:
Moves the preview link into a view component, and only shows it when the form has pages.

Trello card: https://trello.com/c/EHyVdh6L/170-hide-preview-link-from-form-admin-users-if-their-form-does-not-have-any-questions

#### Checklist

    - [n/a] README.md

### DIFF
--- a/app/components/preview_link_component/view.html.erb
+++ b/app/components/preview_link_component/view.html.erb
@@ -1,0 +1,3 @@
+<% if @pages.any? %>
+  <%= govuk_link_to t("home.preview"), @link, { target:"_blank", rel:"noopener noreferrer"} %>
+<% end %>
--- a/app/components/preview_link_component/view.html.erb
+++ b/app/components/preview_link_component/view.html.erb
@@ -1,3 +1,1 @@
-<% if @pages.any? %>
-  <%= govuk_link_to t("home.preview"), @link, { target:"_blank", rel:"noopener noreferrer"} %>
-<% end %>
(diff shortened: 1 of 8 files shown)


app/components/preview_link_component/view.html.erb
```diff
@@ -0,0 +1,3 @@
+<% if @pages.any? %>
```
aliuk2012, line 1: theres a better way to do this sort of thing, let me dig up the docs for it

app/controllers/forms_controller.rb
```diff
@@ -5,6 +5,7 @@ class FormsController < ApplicationController
 
   def show
     @form = Form.find(params[:id])
+    @pages = @form.pages
```
aliuk2012, line 8: why do we need a new instance var? when we can use `@forms.pages` in the view itself?

app/components/preview_link_component/view.html.erb
aliuk2012: Remove this condition in the markup and in the `view.rb` add the following method

```
def render?
   @pages.any
end
```

`render?` method should return a Boolean and based on that results ViewComponent will either render the component or not

docs: https://viewcomponent.org/api.html#render--boolean
example : https://www.honeybadger.io/blog/ruby-view-components/#conditional-rendering

aliuk2012 (approved):

app/components/preview_link_component/view.html.erb
DavidBiddle: Very handy!

PR #146, opened 2022-09-29
Restrict users from just making an incomplete form live

aliuk2012:
- introduces factory_bot and faker
- adds method to form model to check if the form as a whole is ready to be made live
- refactors the forms task list by abstraction the code from controller into a new service called FormTaskListService

Users will now either see "Make your form live" as either text (if form is not ready to be made live), as link (if all the required parts of the form are completed) or not at all (if the form is live)

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


DavidBiddle (approved):
Nice, looks good to me 🎉

PR #147, opened 2022-09-29
Add Dockerfile based upon Alpine Ruby

danworth:
Enables us to build forms-admin into a docker image. This will help with local development based around docker-compose but also prepares for a migration from PaaS.

Based upon Alpine base image and uses non-root user as per GDS way guidance.

### Testing
We will add github action to build the Dockerfile on the different architectures we're using within the team. For now the image builds correctly on my M1 mac and runs as expected within a larger docker-compose set-up including runner and api (future PR shortly).

The ruby process is correctly running as pid 1 so processing TERM signals should work as expected, for example:

```
GDS11321:forms-admin dan.worth$ docker exec local_forms-admin_1 ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
ruby         1  1.8  7.7 259732 157624 ?       Ssl  14:38   0:04 puma 5.6.5 (tcp://0.0.0.0:3000) [app]
```

This builds up on the work that @thomasiles did, thanks Tom 🏅 

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)


thomasiles (approved):
nice one 👍

PR #150, opened 2022-10-03
update text on make live page

alice-carr:
Originally it was thought that we may have to stop users from editing a live form’s name because it would be needed for the form slug and changing the title would break links.

But our URLs now rely on the form ID, so changing the name and slug will not break links.

So we should update the warning content on the ‘Make your form live’ page to remove the wording about not being able to change the name of a live form.

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -113,7 +113,7 @@ en:
         </p>
 
         <p class="govuk-body">
-        After you have made your form live, completed forms will be sent to %{submission_email}.
+          After you have made your form live, completed forms will be sent to %{submission_email}.
         </p>
   not_found:
     body: |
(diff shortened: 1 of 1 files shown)


aliuk2012 (approved):

config/locales/en.yml
```diff
@@ -113,13 +113,8 @@ en:
         </p>
 
         <p class="govuk-body">
-          Make sure you are happy with the form before you make it live. After you
-          have made your form live:
+        After you have made your form live, completed forms will be sent to %{submission_email}.
```
DavidBiddle, line 116: Optionally you could indent this for readability
```suggestion
          After you have made your form live, completed forms will be sent to %{submission_email}.
```

DavidBiddle (approved):
Looks good - just one very minor suggestion from me

PR #155, opened 2022-10-06
Rename :id to :form_id for consistency

DavidBiddle:
In the pages controller we're using `form_id` as the URL param that idenitifes the form. In the form controllers, we're using the much more generic `:id`. This means it's inconsistent across the admin app. It's also inconsistent with the runner, where we use `form_id` across the board.

Changing this also allows us to organise `routes.rb` a bit more readably.

#### Checklist

    - [n/a] README.md


config/routes.rb
```diff
@@ -8,33 +8,38 @@
...
-  scope "forms/:form_id/pages" do
-    get "/" => "pages#index", as: :form_pages
-    get "/:page_id/edit" => "pages#edit", as: :edit_page
-    patch "/:page_id/edit" => "pages#update", as: :update_page
-    get "/new" => "pages#new", as: :new_page
-    post "/new" => "pages#create", as: :create_page
-    get "/:page_id/delete" => "forms/delete_confirmation#delete", as: :delete_page
-    delete "/:page_id/delete" => "forms/delete_confirmation#destroy", as: :destroy_page
+  scope "forms/:form_id" do
+    # resources :forms, only: %i[show]
```
aliuk2012, line 13: do we need this comment?

PR #156, opened 2022-10-06
Restrict form viewing by org

DavidBiddle:
For pages that request a form from the API, this will check if the form is from the user's organisation. If not, it renders a 403 page.

Trello card: https://trello.com/c/gvazDV8a/149-add-basic-authorisation-to-protect-forms-from-being-changed-by-other-organisations

#### Checklist

    - [n/a] README.md

### DIFF
--- a/spec/factories/models/forms.rb
+++ b/spec/factories/models/forms.rb
@@ -3,6 +3,7 @@ FactoryBot.define do
     sequence(:name) { |n| "Form #{n}" }
     submission_email { Faker::Internet.email(domain: "example.gov.uk") }
     privacy_policy_url { Faker::Internet.url(host: "gov.uk") }
+    org { "test-org" }
     live_at { nil }
     missing_sections { nil }
     support_email { nil }
--- a/spec/factories/models/forms.rb
+++ b/spec/factories/models/forms.rb
@@ -38,5 +38,9 @@ FactoryBot.define do
       support_url { Faker::Internet.url(host: "gov.uk") }
       support_url_text { Faker::Lorem.sentence(word_count: 1, random_words_to_add: 4) }
     end
+
+    trait :from_other_org do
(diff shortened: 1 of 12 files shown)


thomasiles (approved):
awesome, looks good to me

spec/factories/models/forms.rb
```diff
@@ -37,5 +38,9 @@
       support_url { Faker::Internet.url(host: "gov.uk") }
       support_url_text { Faker::Lorem.sentence(word_count: 1, random_words_to_add: 4) }
     end
+
+    trait :from_other_org do
+      org { "another-org" }
+    end
```
aliuk2012, line 44: Why do we need this trait? if we need it is to set this in a test, it is possible to just pass it in when we build it

spec/requests/forms_spec.rb
```diff
@@ -52,6 +53,28 @@
         expect(ActiveResource::HttpMock.requests).to include pages_request
       end
     end
+
+    context "with a form from another organisation" do
+      let(:form) do
+        build :form, :from_other_org, id: 2
```
aliuk2012, line 59:
```suggestion
        build :form, org: "another-org", id: 2
```

spec/requests/pages_spec.rb
```diff
@@ -51,6 +52,28 @@
         expect(ActiveResource::HttpMock.requests).to include pages_request
       end
     end
+
+    context "with a form from another organisation" do
+      let(:form) do
+        build :form, :from_other_org, id: 2
```
aliuk2012, line 58:
```suggestion
        build :form, org: "another-org", id: 2
```

spec/factories/models/forms.rb
DavidBiddle: Didn't realise we could do that, great!

PR #157, opened 2022-10-07
Allow optional questions

aliuk2012:
- Add config gem to allow us to start using config settings/feature flags
- Adds feature flag capabilities to our app (Only works for local dev, if production settings are used then all instances in PaaS will use those settings. We need to config our app to use different environments when on PaaS.
- Allow users to mark questions as optional.
 
 
 
image

References: https://trello.com/c/QiFK6P12/164-forms-admin-add-the-ability-for-form-creators-to-mark-questions-as-optional
#### Checklist

    - [x] README.md
    - [ ] Elsewhere (please link)


spec/models/page_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+describe Page do
+  describe "#convert_is_optional_to_boolean" do
+    context "when make_question_optional feature is disabled", feature_make_question_optional: false do
+      it "returns nil" do
+        expect(described_class.new.convert_is_optional_to_boolean).to be nil
+      end
+    end
+
+    context "when make_question_optional feature is disabled", feature_make_question_optional: true do
```
aliuk2012, line 11: `feature_make_question_optional: true` setup the feature flag to be true for this test scenarios

thomasiles (approved):
Looks good! 👍

PR #160, opened 2022-10-11
add longer text format to admin

alice-carr:
Adds longer text answer format

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -63,6 +63,6 @@ private
   end
 
   def answer_types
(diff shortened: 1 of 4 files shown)


aliuk2012 (changes requested):
Work looks good but we need to test this fully with notify before merging. (Happy for my review to be dismissed once the team are happy with it having been tested fully with notify)

We should test this with notify before merging it. There might be issues with notify templating if the user created new line breaks and then starts them with markdown syntax.

thomasiles:
I rebased this branch to catch up with current main, it was reviewed and merged, so I'm closing this one.

PR #175, opened 2022-10-28
Make what happens next a required section

DavidBiddle:

- Makes the 'What happens next' section mandatory for going live
- Makes the 'What happens next' section mandatory when the form is already live
- 'What happens next' remains optional when the form is still in Draft.

Also includes:
- Making the Contact details mandatory for going Live - previously the 'make form live' link wasn't active if contact details weren't filled in, but a user could bypass this by going to the make live URL.
- Updating a few tests to use FactoryBot instead of manually creating form data.

#### Checklist

    - [n/a] README.md


thomasiles (approved):
nice! Good fix on the contact details.

PR #182, opened 2022-11-03
Mark questions completed

DavidBiddle:
Adds the ability for a user to mark a form's questions as complete. 

Trello card: https://trello.com/c/pd1oQb8M/186-forms-admin-allow-form-admin-users-to-mark-the-question-task-as-completed

Found organising this a bit tricky, but my approach here was:
- move the existing 'index' controller action into a new file to stop the pages controller getting too bloated, and add the new action for handling the form into the same controller.
- add a relatively thin MarkCompleteComponent to cover the UI. With some small changes this should be relatively easy to reuse if we want to add this 'mark completion' functionality to other tasks in the list later on.
- added a new feature flag - `task_list_statuses`. This feature doesn't really make sense without the rest of the task list work, so this feature flag should cover all of the task list status work.

#### Checklist

    - [n/a] README.md


config/settings.yml
```diff
@@ -1,3 +1,4 @@
 # Used to add feature flags in the app to control access to certain features.
 features:
   make_question_optional: true
```
aliuk2012, line 3: We should have a ticket to remove this feature flag  and any conditions

spec/components/mark_complete_component/view_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+RSpec.describe MarkCompleteComponent::View, type: :component do
+  let(:mark_complete_form) { Forms::MarkCompleteForm.new(form:).assign_form_values }
+
+  context "when the form has pages" do
+    let(:form) { build(:form, :with_pages, id: 2) }
+
+    context "when the task status feature is enabled" do
+      before do
+        allow(FeatureService).to receive(:enabled?).with(:task_list_statuses).and_return(true)
+      end
```
aliuk2012, line 12:
```suggestion
    context "when the task status feature is enabled", feature_task_list_statuses: true do
```

spec/components/mark_complete_component/view_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        render_inline(described_class.new(form.pages, mark_complete_form, "/"))
+
+        expect(page.text).to have_text("Have you finished editing your questions?")
+      end
+    end
+
+    context "when the task status feature is disabled" do
+      before do
+        allow(FeatureService).to receive(:enabled?).with(:task_list_statuses).and_return(false)
+      end
```
aliuk2012, line 24:
```suggestion
    context "when the task status feature is disabled", feature_task_list_statuses: false do
```

spec/requests/page_list_spec.rb
```diff
@@ -0,0 +1,108 @@
+require "rails_helper"
+
+RSpec.describe "Pages", type: :request do
+  before do
+    User.create!(email: "user@example.com")
+  end
```
aliuk2012, line 6: Why is this needed?

app/controllers/page_list_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+private
+
+  def mark_complete_options
+    [OpenStruct.new(value: "true"), OpenStruct.new(value: "false")]
+  end
+
+  def mark_complete_form_params
+    params.require(:forms_mark_complete_form).permit(:mark_complete)
+  end
+end
```
aliuk2012, line 41: Could we maybe make this controller a bit more generic? I like the fact that the html for the component has been made generic but this controller feels very specific to pages. Maybe a "MarkSectionCompleted" controller?

aliuk2012 (commented):
Just a couple comments/suggestions but otherwise looks really good. let me know if you want to jump on a call to work through these together.

spec/requests/page_list_spec.rb
DavidBiddle: Not sure actually - this was just moved over from the existing pages controller tests. I'll see if removing it breaks anything

config/settings.yml
DavidBiddle: Have made one for that here: https://trello.com/c/RsRgO2kq/187-remove-optional-questions-feature-flag

app/controllers/page_list_controller.rb
DavidBiddle: @aliuk2012 
I'm in two minds about this - while it would be nice to make it more generic the notion of marking the pages task as complete feels like it should be quite closely coupled to the pages list. I guess the way I was imagining it is that the MarkCompleteComponent might appear on multiple pages, but it would post back to different controllers depending on what route was passed into it.

Does that make sense? This might be somewhere where my lack of knowledge re: Rails patterns is letting me down, but I couldn't work out a way of separating out this logic cleanly.

aliuk2012: ok,  lets leave it the way you did it. If we find we need it for MarkSectionComplete for other tasks then we can revisit it. Rule of 3 and keeping things DRY :wink:

aliuk2012 (approved):

PR #183, opened 2022-11-07
Improve request logging

thomasiles:
https://trello.com/c/0COanWVj/274-implement-basic-event-logging-for-forms-admin

Currently the logs sent to splunk are in rails default output format - spread across multiple lines and in text format.

This commit does the following:
- adds the lograge gem
- changes the request logging format to json
- outputs a single line of json for each request/response
- adds request_id, user details and form id to request logs where available
Only production logs are changed, in development they are left as they were.

I looked at a few approaches to this:
- using rails logging without additional gems
- using lograge
- using semantic_logger
- using https://github.com/alphagov/rack-logstasher

Rails offers support for tagging logs and changing the output format but doesn't make it easy to change the standard request format without doing what libraries do anyway.

Semantic logger is a more comprehensive solution used by DfE and offers more control and options and covers a wider use case than just changing the request logs.

The alphagov rack-logstasher is aimed at Sinatra/rack applications which don't have rails request logging already.

Lograge was chosen because it's focussed on changing the standard request logging and nothing else. We may want to change the format of other logs e.g. `logger.info "log something specific" to be in json format (like the runner) and adding some standard keys in those too.

I also looked at passing the request_id to the API and changing the API logs to json. This means changing at lot of the tests in the project so will split that into another PR.

To keep this commit self-contained only the request logs have been altered.

```
Started GET "/forms/8/pages/new" for 172.20.0.1 at 2022-11-07 12:39:52 +0000
Cannot render console from 172.20.0.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
Processing by PagesController#new as HTML
  Parameters: {"form_id"=>"8"}
Authenticating with mock_gds_sso strategy
  User Load (0.7ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1]]
  Rendering layout layouts/application.html.erb
  Rendering pages/new.html.erb within layouts/application
  Rendered pages/_form.html.erb (Duration: 10.3ms | Allocations: 8156)
  Rendered pages/new.html.erb within layouts/application (Duration: 24.7ms | Allocations: 10399)
  Rendered layout layouts/application.html.erb (Duration: 136.8ms | Allocations: 19186)
Completed 200 OK in 208ms (Views: 141.3ms | ActiveRecord: 0.7ms | Allocations: 25100)
```

Example of logs after the change:
```
{"method":"GET","path":"/forms/8/pages/new","format":"html","controller":"PagesController","action":"new","status":200,"duration":413.99,"view":370.6,"db":0.95,"host":"localhost","user_id":1,"user_email":"example@example.com","user_organisation_slug":"government-digital-service","request_id":"08832656-bfad-4aaf-a237-b0c9f6c37f4c","form_id":"8"}
```

#### Checklist

    - [ ] README.md
    - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -10,11 +10,6 @@ class FormsController < ApplicationController
     @task_list = FormTaskListService.call(form: @form).all_tasks
   end
 
-  def append_info_to_payload(payload)
-    super
-    payload[:form_id] = params[:form_id]
-  end
-
 private
 
   def form_params
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -10,6 +10,11 @@ class FormsController < ApplicationController
     @task_list = FormTaskListService.call(form: @form).all_tasks
   end
 
+  def append_info_to_payload(payload)
+    super
+    payload[:form_id] = params[:form_id]
+  end
+
 private
 
   def form_params
--- a/Gemfile
+++ b/Gemfile
@@ -62,6 +62,9 @@ gem "gds-sso"
 gem "govuk-components", "3.2.2"
 gem "govuk_design_system_formbuilder", "3.1.2"
 
+# For structured logging
+gem "lograge"
+
 # Use validate_url so we don't have to write custom URL validation
 gem "validate_url"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -216,6 +216,11 @@ GEM
       railties (>= 6.0.0)
     json (2.6.2)
     jwt (2.5.0)
+    lograge (0.12.0)
+      actionpack (>= 4)
+      activesupport (>= 4)
+      railties (>= 4)
+      request_store (~> 1.0)
     loofah (2.19.0)
       crass (~> 1.0.2)
       nokogiri (>= 1.5.9)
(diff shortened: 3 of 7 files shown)


app/controllers/forms_controller.rb
```diff
@@ -10,6 +10,11 @@ def show
     @task_list = FormTaskListService.call(form: @form).all_tasks
   end
 
+  def append_info_to_payload(payload)
+    super
+    payload[:form_id] = params[:form_id]
+  end
```
aliuk2012, line 16: I think we can structure this a bit better. 

Its odd that we have a "FormsController" and and "FormsBaseController". We had this issue with forms-runner and refactored it to have a single forms base controller that things would inherit from.

thomasiles: yes - could I make the append_info_to_payload in application_controller add form_id if it's in params - then it doesn't need to be repeated? Do you think it's ok to take it from the params like this, rather than the From model from API?

I think the FormsController" and "FormsBaseController" is unfortunate naming in this case - they are not related. The FormsController handles showing the `tasklist`, `FormsBaseController` is the base class for the forms pattern, maybe we can rename the forms pattern controllers into something which doesn't clash with form?

aliuk2012 (approved):

PR #184, opened 2022-11-07
Add signout and profile links to header

DavidBiddle:
Adapts the header so that it:
- can provide the user's name as a link to their profile in Signon
- contains a sign out link
- is a component

In some cases (e.g. in Tests or component preview) we won't have access to the `@current_user` variable won't be available, so in those cases the header falls back to the previous version without login features.

Trello card: https://trello.com/c/KFF7AYq9/251-forms-admin-add-signout-link

#### Checklist

    - [n/a] README.md


app/components/header_component/view.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if @current_user %>
+  <%= govuk_header navigation_classes: ["govuk-header__navigation--end"] do |header|
+    header.product_name(name: "Forms")
```
SamJamCul, line 3: It could be nice to have the name in the locales rather than here

spec/components/header_component/view_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe HeaderComponent::View, type: :component do
+  describe "default status" do
+    before do
+      render_inline(described_class.new(nil))
+    end
+
+    it "contains the service name" do
+      expect(page).to have_text("Forms")
```
SamJamCul, line 10: And then the tests could match against the locales instead

SamJamCul (approved):
Looks good, just a suggestion about the hard coded strings

app/components/header_component/view.html.erb
DavidBiddle: Good spot - have updated!

PR #185, opened 2022-11-08
Fix broken profile link

DavidBiddle:
The previous PR used the gds-sso gem's user `uid` field to generate a direct link to the user's profile. Unfortunately, `uid` doesn't do what I thought it did - it's a UUID rather than the number that appears in the profile link. This means that the profile link 404s, and it also means that the GDS-SSO gem doesn't actually provide us with any way to link to the user profile.

In this PR I've made the link point to the Signon home page. This isn't ideal behaviour as it makes the link text confusing, but it's consistent with the other apps that use Signon. We're also about to review our use of Signon anyway so this may be a temporary fix.

Trello card: https://trello.com/c/KFF7AYq9/251-forms-admin-add-signout-link

#### Checklist

    - [n/a] README.md

### DIFF
--- a/app/components/header_component/view.rb
+++ b/app/components/header_component/view.rb
@@ -6,7 +6,7 @@ module HeaderComponent
       super
       @current_user = current_user
       @show_profile_link = current_user&.name.present?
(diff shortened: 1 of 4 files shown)


alice-carr (approved):
looks good

PR #194, opened 2022-11-10
Add Email confirmation

thomasiles:
# Add email confirmation when setting form submission email

Trello cards: [frontend](https://trello.com/c/d4bc6tiC/344-front-end-build-the-confirmation-code-loop-for-setting-an-email-address-for-completed-forms-to-go-to-should-have) and [backend](https://trello.com/c/acUPDM7v/498-back-end-build-the-loop-for-setting-an-email-address-for-completed-forms-to-go-to-moving-to-github)

This PR adds a new method of setting the submission email for a form. It's a fairly large feature and this PR doesn't implement all functionality (no emails are sent!). It's behind a feature flag in production, which still uses the current journey.

It adds:
 - database table and model to store unconfirmed email addresses along with a six digit code and some auditing data
 - new routes to allow a user to generate a new unconfirmed email address
 - new routes to allow a user to confirm an email address by entering a code. If the code matches, we know the email is controlled by the user and so the form is updated to set the email
 
Points to note:
- This PR doesn't implement the notify integration which sends the code to users
- Only one unconfirmed email address can be created per form. If the user changes it, the old one is lost
- There is no way to resend an email, the user would have to change the email again to get the code emailed again
- Confirmed emails are not remembered - if a users wants to change the submission email to one which was previously used, they must send and enter a code again
- There are no request tests yet. These will be added in another PR to keep the size of this one controllable

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/forms/forms/submission_email_form.rb
+++ b/app/forms/forms/submission_email_form.rb
@@ -8,8 +8,8 @@ class Forms::SubmissionEmailForm
   GOVUK_EMAIL_REGEX = /\.gov\.uk\z/i
 
   validates :temporary_submission_email, presence: true
-  validates :temporary_submission_email, format: { with: EMAIL_REGEX, message: :invalid_email }, if: -> { temporary_submission_email.present? }
-  validates :temporary_submission_email, format: { with: GOVUK_EMAIL_REGEX, message: :non_govuk_email }, if: -> { temporary_submission_email.present? }
+  validates :temporary_submission_email, format: { with: EMAIL_REGEX, message: :invalid_email }
+  validates :temporary_submission_email, format: { with: GOVUK_EMAIL_REGEX, message: :non_govuk_email }
 
   EMAIL_CODE_REGEX = /[0-9]{6}/
   validates :email_code, format: { with: EMAIL_CODE_REGEX, message: :invalid_email_code }, if: -> { email_code.present? }
--- a/app/forms/forms/submission_email_form.rb
+++ b/app/forms/forms/submission_email_form.rb
@@ -1,0 +1,94 @@
+class Forms::SubmissionEmailForm
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  attr_accessor :form, :temporary_submission_email, :email_code, :confirmation_code, :current_user
+
+  EMAIL_REGEX = /.*@.*/
+  GOVUK_EMAIL_REGEX = /\.gov\.uk\z/i
+
+  validates :temporary_submission_email, presence: true
+  validates :temporary_submission_email, format: { with: EMAIL_REGEX, message: :invalid_email }, if: -> { temporary_submission_email.present? }
+  validates :temporary_submission_email, format: { with: GOVUK_EMAIL_REGEX, message: :non_govuk_email }, if: -> { temporary_submission_email.present? }
+
+  EMAIL_CODE_REGEX = /[0-9]{6}/
+  validates :email_code, format: { with: EMAIL_CODE_REGEX, message: :invalid_email_code }, if: -> { email_code.present? }
+  validate :email_code_matches_confirmation_code
+
+  def submit
+    return false if invalid?
+
+    confirmation_code = generate_confirmation_code
+
+    form_submission_email = FormSubmissionEmail.find_by_form_id(form.id)
+
+    form_processed = if form_submission_email.nil?
+                       create_submission_email_record(confirmation_code)
+                     else
+                       update_submission_email_record(confirmation_code, form_submission_email)
+                     end
+
+    if form_processed
+      # send notify email?
+      Rails.logger.info "Email sent to #{temporary_submission_email} with code #{confirmation_code}"
+      true
+    else
+      false
+    end
+  end
+
+  def confirm_confirmation_code
+    return false if invalid?
+
+    # Update the submission email in the form
+    form.submission_email = temporary_submission_email
+    form.save!
+    mark_submission_email_as_confirmed
+  end
+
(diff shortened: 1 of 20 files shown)


app/forms/forms/submission_email_form.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  def mark_submission_email_as_confirmed
+    self.confirmation_code = nil
+    form_submission_email = FormSubmissionEmail.find_by_form_id(form.id)
+    form_submission_email.update!(confirmation_code: nil, updated_by_name: current_user.name, updated_by_email: current_user.email)
+  end
+
+  def assign_form_values
+    # Need to handle when the record doesn't exist
+    # return self if FormSubmissionEmail.find_by!(form_id: form.id).nil?
```
aliuk2012, line 56: Can be removed.

config/locales/forms/submission_email_form.yml
```diff
@@ -0,0 +1,21 @@
...
+            temporary_submission_email:
+              blank: Enter an email address for the form
+              invalid_email: Enter an email address in the correct format, like name@example.com
+              non_govuk_email: Enter an email address ending in .gov.uk
+            email_code:
+              # blank: Enter the confirmation code
+              invalid_email_code: The code should be 6 characters long
+              email_code_mismatch: This code is not correct - double check the code or send a new one
+
+
```
aliuk2012, line 21: remove extra lines

config/routes.rb
```diff
@@ -15,6 +15,15 @@
     post "/change-name" => "forms/change_name#update"
     get "/change-email" => "forms/change_email#new", as: :change_form_email
     post "/change-email" => "forms/change_email#create"
+
+    get "/submission-email" => "forms/submission_email#new", as: :submission_email_form
+
```
aliuk2012, line 20:
```suggestion
```

db/schema.rb
```diff
@@ -10,10 +10,24 @@
...
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_07_13_111847) do
+ActiveRecord::Schema[7.0].define(version: 2022_11_03_162118) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "form_submission_emails", force: :cascade do |t|
+    t.integer "form_id"
+    t.string "submission_email"
```
aliuk2012, line 19:
```suggestion
```
No longer needed

app/forms/forms/submission_email_form.rb
```diff
@@ -0,0 +1,94 @@
+class Forms::SubmissionEmailForm
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  attr_accessor :form, :temporary_submission_email, :email_code, :confirmation_code, :current_user
+
+  EMAIL_REGEX = /.*@.*/
+  GOVUK_EMAIL_REGEX = /\.gov\.uk\z/i
+
+  validates :temporary_submission_email, presence: true
```
DavidBiddle, line 10: In the existing email form, the `presence` validation has an `if: -> { form.live? }` condition on it so that the user can put an empty value in there if the form is in draft. Is that behaviour being intentionally removed?

app/forms/forms/submission_email_form.rb
```diff
@@ -0,0 +1,94 @@
...
+  validates :temporary_submission_email, format: { with: EMAIL_REGEX, message: :invalid_email }, if: -> { temporary_submission_email.present? }
+  validates :temporary_submission_email, format: { with: GOVUK_EMAIL_REGEX, message: :non_govuk_email }, if: -> { temporary_submission_email.present? }
```
DavidBiddle, line 12: Not sure these validations make sense together - think the `presence: true` validation means that the  `if: -> { temporary_submission_email.present? }` in the other two doesn't do anything. That condition makes sense in the existing email form because there it's sometimes valid to have a blank email.

config/settings.yml
```diff
@@ -2,3 +2,4 @@
 features:
   make_question_optional: true
   task_list_statuses: false
+  submission_email_confirmation: false
```
DavidBiddle, line 5: Could set this to `true` unless we're waiting on other work or want to do this in a separate PR

DavidBiddle (commented):
Looks good overall, just a couple of questions around validation on the email

app/forms/forms/submission_email_form.rb
thomasiles: Good spot! I've removed them

config/settings.yml
thomasiles: I think we need to wait until the Notify integration and end-2-end tests are in place.

DavidBiddle: Makes sense!

DavidBiddle (approved):

app/views/forms/submission_email/submission_email_confirmed.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1 class="govuk-heading-l">
+  <span class="govuk-caption-l"><%= @submission_email_form.form.name %></span>
+  <%= t("page_titles.confirm_email_success") %>
+</h1>
+
+<%= t('email_code_success.body_html', submission_email: @submission_email_form.form.submission_email) %>
+
+<%= govuk_link_to(t('email_code_success.continue'), form_path) %>
```
aliuk2012, line 8: We need to wrap `<p>` around this link otherwise the font size of it will be smaller than the text above.

PR #204, opened 2022-11-16
Email confirmation tasklist

thomasiles:
## Add tasklist status for the email confirmation loop

This includes two small change to copy:
- adding a break between the text and email address in the email hint text on the tasklist
- adding a missing paragraph tag to the email code sent page

It also add's the email confirmation loop task status.

There are two email confirmation tasks which need status:
- sending a code
- confirming a code

Forms which already have a submission email set have both status set to
completed.

The second task status is dependent on the first. If the first task is
complete, the confirmation tasks is active and marked as not started.

When the confirmation task is done, both are marked as complete.

If the user tries to change the email to an address which matches the
current submission email address, both tasks are marked as complete.

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/spec/models/form_spec.rb
+++ b/spec/models/form_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 describe Form do
-  let(:form) { described_class.new(name: "Form 1", org: "Test org", live_at: "") }
+  let(:form) { described_class.new(id: 1, name: "Form 1", org: "Test org", live_at: "", submission_email: "") }
 
   describe "#live?" do
     context "when form is draft" do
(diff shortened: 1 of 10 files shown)


spec/models/form_spec.rb
```diff
@@ -75,4 +75,31 @@
...
+    it "with FormSubmissionEmail with no code returns :confirmed" do
+      create :form_submission_email, form_id: form.id, temporary_submission_email: "test@example.gov.uk", confirmation_code: ""
+      expect(form.email_confirmation_status).to eq(:confirmed)
+    end
+
+    it "with FormSubmissionEmail with code and email matches forms returns :confirmed" do
+      form.submission_email = "test@example.gov.uk"
+      create :form_submission_email, form_id: form.id, temporary_submission_email: "test@example.gov.uk", confirmation_code: "123456"
+      expect(form.email_confirmation_status).to eq(:confirmed)
+    end
```
aliuk2012, line 103: Do all these `create...` need to be created, could we not use build instead?

aliuk2012: We agreed, that we would leave the implementation as is and work on refactoring the method thats calling `save` as its doing a lot for one method.

DavidBiddle (approved):
Looks good to me! 🎉

aliuk2012 (approved):

PR #205, opened 2022-11-16
Run Migrations on Start Up

danworth:
The database migrations have been running when the app starts on PaaS. For consistency run them on ECS when the app starts.

Ensures the database migrations run when the app starts on ECS as they currently do on PaaS. Also note that the forms-api runs its migration each time it starts.

### Notes
This has been deployed to our AWS development environment and its running. It also ran the migrations and created the necessary tables:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,4 +62,4 @@ COPY --chown=ruby:ruby . .
 
 EXPOSE 3000
 
-CMD ["rails", "s", "-b", "0.0.0.0"]
+CMD ["/bin/sh", "-c", "rake db:migrate && rails s -b 0.0.0.0"]


thomasiles (approved):
This seems to still handle signals ok?

danworth:
>This seems to still handle signals ok?

Yeah its a great point to raise Tom, and happy if you want to validate my checking. I have sent a `SIGTERM` to the container running locally and from the forms-admin logs it does appear to receive and stop gracefully.

```
GDS11321:local dan.worth$ docker kill --signal=SIGTERM local_forms-admin_1
local_forms-admin_1
```
The app logs
```
forms-admin_1   | * Listening on http://0.0.0.0:3000
forms-admin_1   | Use Ctrl-C to stop
forms-admin_1   | - Gracefully stopping, waiting for requests to finish
forms-admin_1   | Exiting
forms-admin_1   | Shutting down background worker
forms-admin_1   | Killing session flusher
local_forms-admin_1 exited with code 1
```
Also I stopped the container running on ECS and the logs show
```
Exiting
Gracefully stopping, waiting for requests to finish
```
Looking at the processes within the running container...

When it first starts `pid 1` is the shell command which spawns the migration command
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
ruby         1  0.1  0.0   1680     4 ?        Ss   08:43   0:00 /bin/sh -c rake db:migrate && rails s -b 0.0.0.0
ruby         8 17.8  4.3 161996 89132 ?        Sl   08:43   0:01 /usr/local/bin/ruby /usr/local/bin/rake db:migrate
```
after that runs and the puma server is started it becomes `pid 1` so I presume `rails server` is using something like `exec` when spawning the puma process.
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
ruby         1  8.7  5.4 204736 111076 ?       Ssl  08:43   0:01 puma 6.0.0 (tcp://0.0.0.0:3000) [app]
```

All in all I think the SIGTERM is correctly being processed by the rails command.

PR #216, opened 2022-11-23
315-ask-answer-type-first-when-creating-question-in-admin

aliuk2012:

Trello card: https://trello.com/c/N78PUM5K/315-ask-answer-type-first-when-creating-question-in-admin

This PR introduces a new flow for creating a question for a form. This includes a page where the user selects the answer type for their new question, and then sets other question details. To do this, the answer type is stored in session, which is new for this application. 

**Add a question**

To test, add a new question to a form:

image

You should then be able to select an answer type:

image

And on submission, you should be able to review the answer type that's been selected:

image

**Edit an existing question**

Click on edit for an existing question:

image

You are able to see the answer type in the summary, with a change link:

image

Clicking on the change link will bring you to the pre-populated answer type list:

image

Selecting a new answer type will save this to the API, and return to the Edit page.

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


app/controllers/pages_controller.rb
```diff
@@ -19,12 +20,15 @@ def create
 
   def edit
     @page = Page.find(params[:page_id], params: { form_id: @form.id })
+
+    # if session[:page]["id"] == @page.id
+    #   session[:page]["answer_type"]
```
DavidBiddle, line 25: Can these comments be removed?

SamJamCul: Gone!

app/views/pages/_form.html.erb
```diff
@@ -1,27 +1,38 @@
...
+
+  <% end %>
+
+  <%= govuk_summary_list do |summary_list|
+    summary_list.row do |row|
+      row.key(text: "Answer type")
+      row.value(text: t("helpers.label.page.answer_type_options.names.#{page_object.answer_type}"))
+      row.action(text: "Change",
+                 href: change_answer_type_path,
+                 visually_hidden_text: t("helpers.label.page.answer_type_options.names.#{page_object.answer_type}")
```
DavidBiddle, line 28: I think  we want the hidden text to include 'answer type' here - that's how it is in the prototype. Think the logic here is that someone using speech recognition is probably more likely to say 'Click change answer type text' or 'Click change answer type' than 'Click change text'.

```suggestion
                 visually_hidden_text: "Answer type" + t("helpers.label.page.answer_type_options.names.#{page_object.answer_type}")
```

(there's probably a nicer way of concatenating these strings than that)

DavidBiddle (commented):
Looks good overall, have just raised an issue with some visually hidden text

app/controllers/pages_controller.rb
```diff
@@ -35,8 +36,8 @@ def update
 
 private
 
-  def page_params(form_id)
-    params.require(:page).permit(:question_text, :question_short_name, :hint_text, :answer_type, :is_optional).merge(form_id:)
+  def page_params
+    params.require(:page).permit(:question_text, :question_short_name, :hint_text, :answer_type, :is_optional).merge(form_id: @form.id).merge(session:)
```
thomasiles, line 40: Do we need to merge the whole session here or just add `session[:answer_type]`? I think the session has other stuff in it - e.g. csrf token etc.

app/views/pages/_form.html.erb
SamJamCul: Good spot - after some messing around I couldn't find a better way of interpolating within an interpolation, so the `+` seems like the easiest way of doing it.

DavidBiddle (approved):

app/forms/forms/type_of_answer_form.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    session[:page] = { answer_type: }
+  end
+
+  def number(form)
+    # If this page is in form, return the position, else it must be new so
+    # return the number if it was inserted at the end
+    index = form.pages.index(self)
+    (index.nil? ? form.pages.length : index) + 1
+  end
```
thomasiles, line 20: `number(form)` is now in two places. - /models/Page.rb and here. Could we put it somewhere better? would it make more sense in /models/Form.rb and taking page as an argument?

thomasiles (approved):
Impressive work!

app/forms/forms/type_of_answer_form.rb
SamJamCul: This has now been refactored into `/models/Form.rb`, as well as flushing out a bug with the page number caption for editing the answer type!

PR #220, opened 2022-11-28
Add reorder page

thomasiles:
# Add Page reordering to the page list

Page reordering has been added to the admin, though hidden behind a feature flag for now.

'Move up' and 'Move down' buttons are added to the current page_list component. The component may be better off using a list or table element but that change can be implemented in another ticket.

The API changes required for actually moving pages are in [https://github.com/alphagov/forms-api/pull/116](in this PR)

Trello card: https://trello.com/c/vUh413Cv/296-add-reordering-questions-to-admin

image

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/components/page_list_component/_index.scss
+++ b/app/components/page_list_component/_index.scss
@@ -1,0 +1,3 @@
+.govuk-button-group-right {
+  justify-content: end;
+}
--- a/app/components/page_list_component/_index.scss
+++ b/app/components/page_list_component/_index.scss
@@ -1,3 +1,9 @@
-.govuk-button-group-right {
+.app-page-list {
+  .govuk-summary-list:not(:last-child) {
+    margin-bottom: 0;
+  }
+}
+
+.app-page-list__button-group {
   justify-content: end;
 }
--- a/app/components/page_list_component/view.html.erb
+++ b/app/components/page_list_component/view.html.erb
@@ -1,14 +1,50 @@
-<% @pages.each do |page| %>
-  <dl class="govuk-summary-list">
-    <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key govuk-!-font-weight-regular">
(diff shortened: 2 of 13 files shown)


app/components/page_list_component/view.html.erb
```diff
@@ -1,14 +1,50 @@
...
+          <dt class="govuk-summary-list__key govuk-!-font-weight-regular">
+            <%= question_text_with_optional_suffix(page) %>
+          </dt>
+
+          <dd class="govuk-summary-list__actions govuk-!-padding-bottom-6">
+
+            <div class="govuk-button-group form-action-group govuk-button-group-right" }">
+              <% if show_up_button(index) %>
+                <%= f.govuk_submit t("forms.form_overview.move_up"), secondary: true, name: 'move_direction[up]', value: page.id %>
+                <span class="govuk-visually-hidden"><%= page.question_text %></span>
```
DavidBiddle, line 16: Think the hidden text needs to be inside the button or it won't be part of the accessible name. You can probably do this by moving the `<span class="govuk-visually-hidden"></span>` into the translation and then passing the question name into it from this view

thomasiles: Good spot and fix! 👍

app/components/page_list_component/_index.scss
```diff
@@ -0,0 +1,3 @@
+.govuk-button-group-right {
```
aliuk2012, line 1: Is this overriding govuk design system styling?

Should we not perhaps add our own "app" specific override?

thomasiles: yes good point! I've updated the CSS, based off the same convention we used in the task-list.

thomasiles:
I've changed the spacing on page-list component so there is margin below the the last section is kept.

image

DavidBiddle (approved):
Looks good 🎉

PR #225, opened 2022-11-30
Set default builder GOVUKDesignSystemFormBuilder

SamJamCul:
Set the default form builder in the application config.

This reduces the number of places we need to specify it and matches the runner config.

More work can be done to remove it, for example from applicationController and view tests.

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


0atman (approved):
Makes sense to me!

PR #231, opened 2022-12-02
Capybara integration tests

aliuk2012:

We want to move away from cypress and use a combination of Capybara and view spec to gain confidence in our testing strategy. 

Add integration tests for the following journeys:

**From the Home page:** 
  - Creating a new draft form
  - Editing an existing forms name

**From the Form Task list page:**
- Create a new question for each answer type for a completely new form
- Create a new question for each answer type for a form that has existing questions

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


SamJamCul (approved):
Looks good to me! I think it'll get interesting once our answer types start having significantly different workflows, but this is a good way to start things

PR #236, opened 2022-12-08
Add selection questions

DavidBiddle:

Trello cards:
- https://trello.com/c/9AKruIBG/321-add-selection-question-type-to-admin-radio-buttons-only
- https://trello.com/c/2p8Qn2Ns/223-add-selection-question-type-to-admin-checkboxes-only

#### Checklist

  - [x] README.md
  - [x] Elsewhere (please link)


thomasiles (approved):
This is incredibly impressive! ⭐ ⭐ ⭐ ⭐ ⭐ Nice job on on such a huge feature. Approve!

I'd be interested in your thoughts about adding more structure around answer_settings in the future. I guess as we add more possible values as questions get more complicated it might be useful when saving/loading them to the session and displaying different options etc.

app/views/pages/_form.html.erb
```diff
@@ -11,24 +11,49 @@
...
   <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible" />
 
-  <%= f.govuk_check_boxes_fieldset :is_optional, multiple: false, legend: { text: t("pages.add_rules_legend") , size: 'm' } do %>
-    <%= f.govuk_check_box :is_optional, true,0, multiple: false,
-                          small: true,
-                          label: { text: t("pages.make_optional_text")  },
-                          hint: { text: t("pages.make_optional_hint") } %>
-
+  <% if page_object.answer_type == 'selection' %>
+    <%= f.hidden_field :is_optional %>
```
aliuk2012, line 15: Why is this a hidden field? 
Wouldn't we want to reset the value if they user edits an existing question (that is set to a type that allows it to be optional) and changes the type to "selection"  and therefore might not be able to mark it optional?

config/routes.rb
```diff
@@ -53,6 +55,8 @@
         scope "/edit" do
           get "/type-of-answer" => "pages/type_of_answer#edit", as: :type_of_answer_edit
           post "/type-of-answer" => "pages/type_of_answer#update", as: :type_of_answer_update
+          get "/selections_settings" => "pages/selections_settings#edit", as: :selections_settings_edit
+          post "/selections_settings" => "pages/selections_settings#update", as: :selections_settings_update
```
aliuk2012, line 59:
```suggestion
          get "/selections-settings" => "pages/selections_settings#edit", as: :selections_settings_edit
          post "/selections-settings" => "pages/selections_settings#update", as: :selections_settings_update
```

config/routes.rb
```diff
@@ -45,6 +45,8 @@
       scope "/new" do
         get "/type-of-answer" => "pages/type_of_answer#new", as: :type_of_answer_new
         post "/type-of-answer" => "pages/type_of_answer#create", as: :type_of_answer_create
+        get "/selections_settings" => "pages/selections_settings#new", as: :selections_settings_new
+        post "/selections_settings" => "pages/selections_settings#create", as: :selections_settings_create
```
aliuk2012, line 49:
```suggestion
        get "/selections-settings" => "pages/selections_settings#new", as: :selections_settings_new
        post "/selections-settings" => "pages/selections_settings#create", as: :selections_settings_create
```

app/views/pages/_form.html.erb
DavidBiddle: @aliuk2012 `is_optional` behaves slightly differently for the checkbox and radio inputs, we're using it as the way to decide whether the 'none of the above' option appears in the runner. So for this input we're setting the value on the selections_settings view, where it's called `Include an option for ‘None of the above’` to make it clearer what it does. 

So for a user editing an existing question, the flow would be:
- go into the edit question page
- click 'Change' on answer type
- select 'Selection'
- progress to the selection_settings view, where `Include an option for ‘None of the above’` will be populated with the existing value of `is_optional` but they can edit if that's no longer correct
- get returned to the edit question page, where the is_optional checkbox no longer appears. 

The hidden input is just on this page to make passing the data around easier.

aliuk2012 (approved):

PR #251, opened 2022-12-21
Autofill date of birth

DavidBiddle:

Trello card: https://trello.com/c/sDrOCmAp/383-dob-implementation-of-autofill-work-in-production

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/controllers/pages/date_settings_controller.rb
+++ b/app/controllers/pages/date_settings_controller.rb
@@ -1,0 +1,50 @@
+class Pages::DateSettingsController < PagesController
+  def new
+    input_type = session.dig(:page, "answer_settings", "input_type")
+    @date_settings_form = Forms::DateSettingsForm.new(input_type:)
+    @date_settings_path = date_settings_create_path(@form)
+    @back_link_url = type_of_answer_new_path(@form)
+    render "pages/date_settings"
+  end
+
+  def create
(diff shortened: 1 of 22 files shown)


SamJamCul (approved):

app/controllers/pages/date_settings_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    if @date_settings_form.submit(session)
+      redirect_to new_page_path(@form)
+    else
+      render "pages/date_settings"
+    end
+  end
+
+  def edit
+    @page = Page.find(params[:page_id], params: { form_id: @form.id })
+    input_type = @page&.answer_settings&.input_type
```
aliuk2012, line 24: Really don't like our usage of `&` to fail gracefully. We should try and make sure all questions have at least answer_settings set

aliuk2012:
Looks alright to me, only suggestion might be that we should add a migration for existing date types to set the answer_settings so we dont end up having a mishmash of date types legacy vs new version

app/controllers/pages/date_settings_controller.rb
DavidBiddle: Good point - have made a placeholder card for this so we don't forget: https://trello.com/c/RDzeLJqz/425-write-migrations-for-changed-autofill-fields

PR #260, opened 2023-01-03
Add logging user_ip in production

thomasiles:
We currently don't log the ip address of the user accesssing the admin.

This commit adds a new entry in the logs for user_ip, which in turn uses the rails remote_ip call. The client ip is picked from the X-forwarded-for header, which is passed through the PaaS router to the application. This may or may not be the right thing to log - it's not clear from running the code on localhost that the correct ip is picked up.

We'll merge this and check the logs in staging to check we have the correct value.

See documentation from MDN and rails for more details.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

https://api.rubyonrails.org/v5.1/classes/ActionDispatch/Request.html#method-i-remote_ip-3D

Trello card: https://trello.com/c/bgLy2njq/355-implement-basic-protective-monitoring-of-govuk-forms

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,6 +26,7 @@ class ApplicationController < ActionController::Base
       payload[:user_organisation_slug] = current_user.organisation_slug
     end
     payload[:request_id] = request.request_id
+    payload[:remote_ip] = request.remote_ip
     payload[:form_id] = params[:form_id] if params[:form_id].present?
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,7 +26,7 @@ class ApplicationController < ActionController::Base
       payload[:user_organisation_slug] = current_user.organisation_slug
     end
     payload[:request_id] = request.request_id
-    payload[:remote_ip] = request.remote_ip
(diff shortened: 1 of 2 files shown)


danworth (approved):
Looks like its worth testing. I'm not sure how you configure Rails to pick out the correct client IP as the one to the left of the trusted proxy (and their docs don't seem to provide much info). In Express you can specify the number of trusted proxies or a CIDR block to match on. Worth experimenting because as per that link it's easy to spoof if you just take the far left IP (a slight hole I found in a previous team i worked on). The number of proxies should remain as 1 after migration but worth keeping in mind.

app/controllers/application_controller.rb
```diff
@@ -26,6 +26,7 @@ def append_info_to_payload(payload)
       payload[:user_organisation_slug] = current_user.organisation_slug
     end
     payload[:request_id] = request.request_id
+    payload[:remote_ip] = request.remote_ip
```
danworth, line 29: Just approved it but now I'm wondering why we call it `payload.remote_ip` here and then `user_ip` below, can it be consistent?

PR #262, opened 2023-01-06
Fix issue with input type playback

DavidBiddle:
Editing a question page with an exisitng date field (i.e. one without an answer settings field) will result in an error, since the page_settings_summary component is expecting the input type to be present. The same will be true of address fields if that it merged before this change.

This PR adds a check for this. This should be a temporary solution - the proper solution is to migrate the existing date (and address) fields. There is a ticket for this in the backlog which should be done in the next sprint.

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/components/page_settings_summary_component/view.html.erb
+++ b/app/components/page_settings_summary_component/view.html.erb
@@ -28,22 +28,24 @@
                     visually_hidden_text: t("selections_settings.include_none_of_the_above")) end %>
       <% end %>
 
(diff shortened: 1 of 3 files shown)


aliuk2012:
Good catch, should there maybe be a test to ensure this sort of thing doesn't happen again unexpectedly?

DavidBiddle:
@aliuk2012 Good point, have added

aliuk2012 (approved):

PR #263, opened 2023-01-06
Add request id forwarding to logs

thomasiles:
# Make the log config consistent with runner and API

- Do config in production.rb
- add request_id to API calls

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


app/controllers/application_controller.rb
```diff
@@ -33,4 +34,12 @@ def append_info_to_payload(payload)
   def clear_questions_session_data
     session.delete(:page) if session[:page].present?
   end
+
+  def set_request_id
+    if Rails.env.production?
+      [Form, Page].each do |active_resource_model|
+        active_resource_model.headers["X-Request-ID"] = request.request_id
```
danworth, line 41: From [the docs](https://api.rubyonrails.org/classes/ActionDispatch/RequestId.html) it looks like this will either take the existing `x-request-id` header field from the request or generate one if not available which seems ideal (the aws ALB should add `x-request-id` I think, not sure about the PaaS Route) 👍

danworth (approved):
Nice 🏅

PR #266, opened 2023-01-09
Disable sentry transactions

danworth:
    Disable Sentry Transactions

    Sentry Transactions can be used to monitor application performance by
    sampling transaction data. We are not using this aspect of Sentry and
    have decided to turn it off.

### Notes
Setting this to `0.0` or removing it all together will disable transaction data: https://docs.sentry.io/platforms/ruby/guides/rails/configuration/options/#tracing-options

I've opted to set it to `0.0` so there is some on-going record for the future that we have intentionally disabled this feature.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@
 config/settings.local.yml
 config/settings/*.local.yml
 config/environments/*.local.yml
+
+# Vim swap files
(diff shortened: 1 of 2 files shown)


alice-carr (approved):
👍

PR #269, opened 2023-01-10
Migrate Sentry Config to Settings

danworth:
Use the Settings gem and pattern for setting the Sentry configuration.

## Notes
Uses the same pattern as forms-api: https://github.com/alphagov/forms-api/blob/main/config/settings.yml#L8


danworth:
No longer necessary.

PR #272, opened 2023-01-11
Remove obsolete db tables

aliuk2012:

These were some how added to schema.db back in August last year but we shouldn't be using them directly. 
The only two tables forms-admin uses are as follows

- Users, to store signed in signon users 
- form_submission_emails, used when confirming form submission email addresses and checking if they have been verified.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


danworth (approved):
Looks good. Where is `db/schema.rb` used out of interested. I've just checked forms-admin on aws rds and those tables don't exist so I presume they're not part of the db migration setup?

PR #276, opened 2023-01-13
Change IP logging in production

thomasiles:
# Change PaaS client_ip logging logic

Client IP logging for protective monitoring currently uses the value returned from rails `request.remote_ip` which uses the rails configuration for trusted proxies to work out the most reasonable value for client IP based on the `X-forwarded_for` headers. The logic for  getting the clients real IP isn't as simple as using the socket IP as in PaaS clients are always connecting through a router and we also  can't trust the value of `X-forwarded_for` as the client may be spoofing the value.

https://github.com/rack/rack/blob/main/lib/rack/request.rb

We don't set any values for trusted_proxies - we would set these to amazon cloud-front values, for example see the following gist:

https://gist.github.com/mrk21/f25b22528d69d0cfdd2426e8470a8ee6

In this commit instead of tackling this, we simply log the first value set in `X-forwarded_for` after basic sanitisation to protect against malicious input.

This commit also picks up request_id using the PaaS specific value, to help debugging and observing the above behaviour.

Trello card: https://trello.com/c/bgLy2njq/355-implement-basic-protective-monitoring-of-govuk-forms

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,9 +1,11 @@
+require "resolv"
+
 class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
+  before_action :set_request_id
   before_action :authenticate_user!
   before_action :set_user_instance_variable
   before_action :check_service_unavailable
-  before_action :set_request_id
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
   before_action :clear_questions_session_data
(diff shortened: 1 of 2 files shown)


DavidBiddle:
@thomasiles Code looks good to me, just checking I understand the commit message properly - is this a PaaS-specific change that we plan to swap for using `trusted_proxies` when we do the AWS switchover?

thomasiles:
> @thomasiles Code looks good to me, just checking I understand the commit message properly - is this a PaaS-specific change that we plan to swap for using `trusted_proxies` when we do the AWS switchover?

The code itself is not PaaS specific and can be deployed to PaaS or AWS. I think the topic of protective monitoring may be looked at again when we switch to AWS as there may be better solutions when that happens. 

This commit is also a bit of a test - if we find this is enough to log IP addresses we might not need to do anything more complicated. If it doesn't give us what we need it's back to the drawing board - that might mean adding cloudfront IP's to trusted proxies (even in PaaS it's the same IPs because it uses AWS underneath).

DavidBiddle (approved):
@thomasiles that makes sense - approved!

PR #278, opened 2023-01-16
Fix status counter for live forms

aliuk2012:

BUG: We were counting all tasks needed to make form live, but we didn't mark the "make form live" task as completed once the form went live so it meant that live forms ended up reporting that the user completed 8 out of 9 tasks.
Designers agreed that we should just hide this text once a form is live and this is a temporary solution because eventually when a form is live, it will become read-only and the user will not longer see this task list.

- Mark "make form live task" completed when its made live.
- hide summary text when completed & total count match (i.e form is live)

#### Before
image

#### After
image

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/spec/components/task_list_component/view_spec.rb
+++ b/spec/components/task_list_component/view_spec.rb
@@ -95,6 +95,25 @@ RSpec.describe TaskListComponent::View, type: :component do
       it "does render the summary text at the top of the task list" do
         expect(page).to have_selector(".app-task-list__summary", text: "You've completed 23 of 34 tasks.")
       end
+
+      context "when all tasks completed then hide the summary text (i.e form is live)" do
+        before do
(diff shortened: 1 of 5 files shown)


spec/components/task_list_component/view_spec.rb
```diff
@@ -95,6 +95,25 @@
...
+                { task_name: "task a", path: "#", status:, active: },
+              ] },
+            { title: "section 2",
+              rows: [
+                { task_name: "task d", path: "#", status:, active: },
+              ] },
+          ]))
+        end
+
+        it "does render the summary text at the top of the task list" do
```
SamJamCul, line 113: Should be the other way around, `does not render the...`

SamJamCul (approved):

PR #283, opened 2023-01-18
Add confirmation panel to 'your form is live' confirmation page

lfdebrux:
Make the 'your form is live' confirmation page match the GOV.UK Design System [confirmation pages](https://design-system.service.gov.uk/patterns/confirmation-pages/) pattern.

We also add some specs for this view.

Trello card: https://trello.com/c/lZBRotsv/364-add-green-panel-to-form-live-confirmation-page

Note that this PR doesn't change the form URL component, we'll do this in a separate PR.

##### Screenshot

Screenshot of 'your form is live' page

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/spec/views/forms/make_live/confirmation.html.erb_spec.rb
+++ b/spec/views/forms/make_live/confirmation.html.erb_spec.rb
@@ -1,0 +1,27 @@
+require "rails_helper"
+
+describe "forms/make_live/confirmation.html.erb" do
+  around do |example|
+    ClimateControl.modify RUNNER_BASE: "runner-host" do
+      example.run
+    end
+  end
+
+  before do
+    assign(:form, OpenStruct.new(id: 1, name: "Form 1", form_slug: "form-1"))
+    render template: "forms/make_live/confirmation"
+  end
+
+  it "contains page heading and sub-heading" do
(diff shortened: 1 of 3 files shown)


spec/views/forms/make_live/confirmation.html.erb_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+describe "forms/make_live/confirmation.html.erb" do
+  around do |example|
+    ClimateControl.modify RUNNER_BASE: "runner-host" do
+      example.run
+    end
+  end
+
+  before do
+    assign(:form, OpenStruct.new(id: 1, name: "Form 1", form_slug: "form-1", status: "draft", pages: nil))
```
aliuk2012, line 11: why does it need `status` and `pages`?

lfdebrux: Good catch. I copied this line from the tests for `show.html`, except replacing `pages:` with `nil`, but you're right, it shouldn't need to be there.

It is surprising though that the view would work with `status: draft`, but I guess the routing logic is what takes care of that?

aliuk2012 (approved):

PR #293, opened 2023-01-26
Autofill cleanup tasks

DavidBiddle:
Resolves some divergences from the design, namely:
On the answer type page:
- makes the hint text for each input match the design
- hides the name input behind the feature flag (this was wrongly made visible)
On the answer settings pages:
- Adds a ‘Go to your questions’ link at the bottom of the page
On the name settings page:
- Change title list to `Mr, Mrs, Miss, Ms, Mx`

Trello card: https://trello.com/c/nPbS36yK/483-autofill-cleanup-tasks

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -4,7 +4,7 @@ class Page < ActiveResource::Base
   self.include_format_in_path = false
   headers["X-API-Token"] = ENV["API_KEY"]
 
-  ANSWER_TYPES = %w[name single_line number address date email national_insurance_number phone_number long_text selection organisation_name text].freeze
+  ANSWER_TYPES = %w[name organisation_name email phone_number national_insurance_number address date selection number text single_line long_text].freeze
 
   COMPLEX_ANSWER_TYPES = %w[selection text date address name].freeze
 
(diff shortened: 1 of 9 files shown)


app/models/page.rb
```diff
@@ -4,7 +4,7 @@ class Page < ActiveResource::Base
   self.include_format_in_path = false
   headers["X-API-Token"] = ENV["API_KEY"]
 
-  ANSWER_TYPES = %w[name single_line number address date email national_insurance_number phone_number long_text selection organisation_name text].freeze
+  ANSWER_TYPES = %w[name organisation_name email phone_number national_insurance_number address date selection number text single_line long_text].freeze
```
aliuk2012, line 7: @DavidBiddle do you you think now might be a good time to do the refactor, we spoke about this morning? Splitting this out into simple/complex and then this constant would just be a merge of simple/complex so we have a complete set of answer types?

aliuk2012:
Just a suggestion about refactoring the answer type constant but otherwise looks good

app/models/page.rb
DavidBiddle: @aliuk2012 I think I'd prefer to do that when we remove the feature flag, as I think there's some related work involved there. But I've renamed `COMPLEX_ANSWER_TYPES` to `ANSWER_TYPES_WITH_SETTINGS` to at least make that clearer here

aliuk2012: Thank you thats much clearer

aliuk2012 (approved):

PR #297, opened 2023-01-31
Fixed incorrect apostrophes

hannahkc:

I have:

- fixed several apostrophes that were using a left/opening single quote mark rather than a right/closing single quote mark. 
- fixed a couple of straight apostrophes. 
- changed: "Your form’s contact details" to "Your form’s contact details for support" because it's more specific.


aliuk2012 (approved):
Nice change 👍

PR #301, opened 2023-02-06
Implement Basic Auth Strategy

danworth:
The user-research environment requires basic auth using a single set of credentials. All other non-local environments use Signon.

It is necessary to disable Warden's intercept_401 behaviour to prevent it from interfering with the basic auth flow.

Trello card: https://trello.com/c/al6fxy57/459-add-login-part-to-the-user-research-environment

#### Notes
Warden's `intercept_401` behaviour isn't really documented but disabling this stops its failure strategy from interrupting Rail's basic auth flow. This is only disabled when the use basic auth setting is enabled.

When we get to abstracting and removing Signon we'll probably want to put these auth methods into a separate concern/service but for now this keeps the change minimal and simple to understand.

This can be tested by running locally and toggling between basic and sigon authentication. @DavidBiddle has some other suggestions from the branch which involved removing the signon gem completely so we'll bring those things over when he's back (I started a new branch because it seemed simpler to start with signon gem being installed rather than putting it back in...).


app/controllers/application_controller.rb
```diff
@@ -3,30 +3,51 @@
...
     super
     payload[:host] = request.host
-    if current_user.present?
-      payload[:user_id] = current_user.id
-      payload[:user_email] = current_user.email
-      payload[:user_organisation_slug] = current_user.organisation_slug
+    if @current_user.present?
+      payload[:user_id] = @current_user.id
+      payload[:user_email] = @current_user.email
+      payload[:user_organisation_slug] = @current_user.organisation_slug
```
aliuk2012, line 50: This is why we need to be careful with using instance vars. Are we calling a `current_user` object or an instance var sort of confusion. The two could be very different

danworth: The `ApplicationController` before this change was always setting the instance var `@current_user = current_user`. This won't work for basic auth because it doesn't effect the result of `current_user`. Therefore we set `@current_user` depending on the auth method and then make consistent use of it. The change here from `current_user` to `@current_user` is intentional.

The `HomeController` was calling the `current_user` method rather than the instance var which won't work because basic auth doesn't effect it. I think the change in this PR aligns on the use of the instance var which seems the Rail convention, especially for flows through controllers (HomeController is inheriting from the ApplicationController) and to their views. Personally I think using the instance var is easier to understand because it relies upon a variable we've set in our code and not the results of `current_user` which we have less visibility of how it's being set. Rails convention seems to be not to use instance vars for components because they're shared between views and instance var make less sense.

Happy to chat, but would like to keep the scope of this change limited to the auth strategies, greater refactoring of instance vars should be done separately I feel. I think this improves the consistent use of the `@current_user` v `current_user`.

app/controllers/application_controller.rb
```diff
@@ -3,30 +3,56 @@
 class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
   before_action :set_request_id
-  before_action :authenticate_user!
-  before_action :set_user_instance_variable
+  before_action :authenticate
   before_action :check_service_unavailable
```
lfdebrux, line 7: I realise this isn't something that was changed in this PR, but I just noticed that with the order of these before actions as currently written, users will need to log in before they see the service unavailable page. Is this the behaviour we want?

app/controllers/application_controller.rb
```diff
@@ -3,30 +3,56 @@
...
+      signon_auth
     end
   end
 
-  def set_user_instance_variable
+  def basic_auth
+    request.env["warden"].manager.config.intercept_401 = false
+    authenticate_or_request_with_http_basic do |name, password|
+      name == Settings.basic_auth.username && password == Settings.basic_auth.password
+    end
```
lfdebrux, line 24: Any reason we're not using `http_basic_authenticate_or_request_with` here?

https://rubydoc.info/github/rails/rails/ActionController%2FHttpAuthentication%2FBasic%2FControllerMethods:http_basic_authenticate_or_request_with

```suggestion
    http_basic_authenticate_or_request_with (
      name: Settings.basic_auth.username,
      password: Settings.basic_auth.password,
    )
```

spec/controller/application_controller_spec.rb
```diff
@@ -20,4 +20,48 @@
...
     end
   end
+
+  controller do
+    def index
+      render status: :ok, json: {}
+    end
+  end
+
+  context "when authenicating a user" do
```
lfdebrux, line 30:
```suggestion
  context "when authenticating a user" do
```

lfdebrux (commented):
Tested basic auth locally, seemed to work well! Couple of comments on the code, feel free to ignore though.

app/components/header_component/view.rb
```diff
@@ -2,10 +2,10 @@
 
 module HeaderComponent
   class View < ViewComponent::Base
-    def initialize(current_user)
+    def initialize(user)
```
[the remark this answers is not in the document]
DavidBiddle: It's because `current_user`is also the name of the signon method we are/were using - when we were changing from `current_user` to `@current_user` elsewhere we found it a bit confusing that we were also using `current_user` as the parameter name in this class. You're right though, we should add something about this to the commit message

app/controllers/application_controller.rb
DavidBiddle: No, think that should work

app/controllers/application_controller.rb
DavidBiddle: Good spot - I think `check_service_unavailable` should probably the first thing that runs, so I'll move it to the top.

app/controllers/application_controller.rb
```diff
@@ -3,30 +3,51 @@
 class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
```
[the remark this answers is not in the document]
DavidBiddle: I think this suggestion seems sensible. I quite like the idea of only including this if we're not using basic auth:

```suggestion
  include GDS::SSO::ControllerMethods unless Settings.basic_auth.enabled
```

app/controllers/application_controller.rb
danworth: yeah I noticed this as well, I would like to do it in its own PR, one to keep the scope of this tight to basic auth and the other is because if we need to back this out we know that we're only backing out the basic auth and not undoing other good work which then gets forgotten. Thanks for the suggestion Laurence.

aliuk2012: Also `check_service_unavailable` could be updated to use setting config yaml file too.

spec/components/header_component/view_spec.rb
```diff
@@ -40,4 +40,22 @@
       expect(page).to have_link(I18n.t("header.sign_out"), href: "/auth/gds/sign_out")
     end
   end
+
+  context "when baic auth is enabled" do
```
lfdebrux, line 44:
```suggestion
  context "when basic auth is enabled" do
```

app/controllers/application_controller.rb
0atman: That's a really good realisation!

DavidBiddle: That's a very good spot Al, think it's another good candidate to go in the refactor PR

lfdebrux (approved):

PR #308, opened 2023-02-08
Add "Create a draft" call to action button

aliuk2012:

This takes the used back to the form task list page.

Trello card: https://trello.com/c/jWdf1F56/514-add-create-a-draft-to-edit-button-if-form-doesnt-already-have-a-draft-or-edit-the-draft-of-this-formif-form-does-have-an-existin

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/spec/views/live/show_form.html.erb_spec.rb
+++ b/spec/views/live/show_form.html.erb_spec.rb
@@ -116,7 +116,13 @@ describe "live/show_form.html.erb" do
     end
   end
 
-  xit "contains a link to edit the form" do
-    expect(rendered).to have_link("Edit this form", href: form_path(form.id, edit: true))
+  it "does not contain a link to create a new draft" do
+    expect(rendered).not_to have_link(t("show_live_form.draft_create"), href: form_path(form.id))
+  end
+
+  context "when draft/live feature enabled", feature_draft_live_versioning: true do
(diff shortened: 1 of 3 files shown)


spec/views/live/show_form.html.erb_spec.rb
```diff
@@ -116,7 +116,13 @@
     end
   end
 
-  xit "contains a link to edit the form" do
-    expect(rendered).to have_link("Edit this form", href: form_path(form.id, edit: true))
+  it "does not contain a link to create a new draft" do
+    expect(rendered).not_to have_link(t("show_live_form.draft_create"), href: form_path(form.id))
+  end
```
SamJamCul, line 121: I'm not sure if it's necessary to have the negative test? We could explicitly turn the feature flag off for this test so that it doesn't fail when the flag switches

SamJamCul (approved):

spec/views/live/show_form.html.erb_spec.rb
aliuk2012: I did think about explicitly  turning off the feature flag but that wouldn't have caught the scenario where someone accidentally removes the feature flag clause on the page which then renders the button. regardless whether the flag is removed or turned off we shouldn't be expecting to see the button.

PR #326, opened 2023-02-17
Refactor controllers

aliuk2012:

Review individual commits, mostly refactoring controller actions to make them more restful.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


app/controllers/page_list_controller.rb
```diff
@@ -23,11 +16,11 @@ def update
         raise StandardError, "Save unsuccessful"
       end
     else
-      render :edit, status: :unprocessable_entity
+      render :index, status: :unprocessable_entity
```
danworth, line 19: Just curious and surely some convention I'm unaware of, before it was referencing `:edit` template within the pages_list controller https://github.com/alphagov/forms-admin/blob/main/app/views/page_list/edit.html.erb. Now we seem to want it to render the `index` view of `pages` which doesn't exist https://github.com/alphagov/forms-admin/tree/main/app/views/pages. The Rails docs also seem to imply that if you want one controller to render the template from another controller you have to give it more information https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-an-action-s-template-from-another-controller so I would expect something like `render "pages/index". 

Since it builds and CI passes I assume it does and it's using some convention that I don't get yet, but also no tests were updated in this commit so perhaps its not being tested? Anyway thought it worth asking :)

danworth (commented):
The granular commits and their messages are brilliant and made it simple to see the changes and why they're being made (I think a good example or why merge commits are useful because preserving that will be handy) 🏅 It seems you want David or Sam to review so I'll just leave a comment about some Rails convention that I probably don't get yet 😄

app/controllers/page_list_controller.rb
aliuk2012: Oh thats a great catch... 👏 

I was using RubyMines "Refactor" feature when you move/rename files it tries to be helpful which is great in this instance BUT we obviously dont seem to be testing the unhappy path which is why tests are all passing. 

See code coverage report as proof that we aren't testing unhappy path...
image

DavidBiddle (approved):
Looks good - very sensible refactors!

spec/requests/forms_spec.rb
```diff
@@ -171,4 +135,39 @@
       end
     end
   end
+
+  describe "#index - Viewing all forms for an organisation" do
+    describe "Given a form" do
```
lfdebrux, line 140: This doesn't seem quite right as a description... do we even need the extra describe block here? Could we just have the test directly in the `#index ...` group?

config/routes.rb
```diff
@@ -43,7 +43,7 @@
     scope "/pages" do
       get "/" => "page_list#edit", as: :form_pages
       post "/" => "page_list#update"
-      post "/move-page" => "page_list#move_page", as: :move_page
+      post "/move-page" => "pages#move_page", as: :move_page
```
lfdebrux, line 46: It's not obvious to me why we're making this page, would you mind adding more detail to the commit message?

app/controllers/pages_controller.rb
```diff
@@ -77,4 +96,8 @@ def handle_submit_action
   def answer_types
     @answer_types = Page::ANSWER_TYPES
   end
+
+  def mark_complete_options
+    [OpenStruct.new(value: "true"), OpenStruct.new(value: "false")]
+  end
```
lfdebrux, line 102: Does this need to be defined in both `pages_controller` and `page_list_controller`?

app/controllers/page_list_controller.rb
lfdebrux: I'm confused generally about the difference between `pages_controller` and `page_list_controller`.

It looks like from the routes file in the main branch that the pages controller handles all changes that might be done via the page list form, including reordering and marking the questions as complete, which makes sense to me.

However in this PR you're moving most of the functionality from page list controller to pages, leaving just marking as complete ... I personally think it makes sense to either go the whole way and remove the page list controller (put the confirmation update method in the pages controller with the others?), or just leave it as it is and keep the split. Having the functionality for the pages list form split across two controllers seems a little strange to me.

aliuk2012: @danworth see 2ebfcb9c515afedff7355a65a026e334ab726698, basically, I moved some of the functionality out of the controller to simplify it. By abstracting this bit of code and getting the controller to only be responsible to managing which page the user sees next makes it easier to test. 

"Thin controllers are always better than fat controllers"


100% now for both controller specs and the model 
image

image

aliuk2012: @lfdebrux 100% agree, this is something we were debuting about when we make the changes to move update action. I've now completely got rid of `page_list_controller` which didn't make sense as we dont have a pages_list resource. done as part of 8c301bf705288d547fd485f9fbe88d793fee0932

app/controllers/pages_controller.rb
aliuk2012: Good spot, I looked into this and it turned out we didn't even need this or the instance variables which were calling this. These values are actually set inside a view component which expects specific values. a196337b4dec47ecc4e2251e03b9d91fdb76e566 removes the instance var and this method

config/routes.rb
aliuk2012: 759376e825804e37cc382871ffd5e74e89c1d815

spec/requests/forms_spec.rb
aliuk2012: Fixed this now

app/controllers/pages_controller.rb
lfdebrux: Ah, you mean https://github.com/alphagov/forms-admin/blob/main/app/components/mark_complete_component/?

lfdebrux: > These values are actually set inside a view component which expects specific values.

Might be worth noting this in the commit message

app/controllers/forms_controller.rb
```diff
@@ -22,24 +22,17 @@ def mark_pages_section_completed
...
 
-    if @mark_complete_form.valid?
-      @form.question_section_completed = @mark_complete_form.mark_complete
-      if @form.save
-        redirect_to form_path(@form)
-      else
-        raise StandardError, "Save unsuccessful"
-      end
+    if @mark_complete_form.mark_section
+      redirect_to form_path(@form)
```
lfdebrux, line 25: Nice!

lfdebrux (approved):

PR #331, opened 2023-02-23
Update content for editing a live form journey

aliuk2012:
## What problem does the pull request solve?

We want it to be clear to form creators when they are editing an existing form and when they are creating a new form. This PR updates the content in the editing an existing form journey.

Trello card: https://trello.com/c/VTr26Wi1

Some of these changes are behind a feature flag, but most are not, as we want the content to be different whether their changes are going into a draft or going live immediately.

Also, note that this PR removes the GET route for the 'make your form/changes live' confirmation page; we agreed after discussion that we wanted to redirect users away from the confirmation page if they came back to it, as the information is available elsewhere and this is different to a full service where the transaction ends forever with the confirmation page [[1]]. Removing the separate `#confirmation` endpoint means we don't have to add lots of extra logic around the form state and last action, we can just render the page immediately with the POST. What this means in practice for users is that if they make their form live and then try and refresh the page, the browser will ask to confirm form resubmission (tested in latest Chrome on macOS). If they bookmark the confirmation page URL, when they try and visit the URL at a later date they will see the form asking whether they want to make their changes live.

[1]: https://trello.com/c/VTr26Wi1/518-make-changes-live-in-forms-admin-and-update-task-list-content#comment-63f8c88229014bc89f6646d0

<details>
<summary><h2>Screenshots</h2></summary>

### Make your form live

Screenshot of page at /forms/2/make-live when form has not been made live before and draft/live versioning feature is enabled

### Edit your form

Screenshot of page at /forms/1/ when form has been made live and draft/live versioning feature is enabled

### Make your changes live

Screenshot of page at /forms/1/make-live when form has been made live before and draft/live versioning feature is enabled

### Your changes are live

Screenshot of page at /forms/1/live-confirmation when form had been made live at least once before and draft/live versioning feature is enabled

</details>


app/controllers/forms/make_live_controller.rb
```diff
@@ -1,8 +1,11 @@
 module Forms
   class MakeLiveController < BaseController
     def new
-      redirect_to live_confirmation_url if current_form.live? && !FeatureService.enabled?(:draft_live_versioning)
-      @make_live_form = MakeLiveForm.new(form: current_form)
+      if current_form.live? && !FeatureService.enabled?(:draft_live_versioning)
+        redirect_to live_confirmation_url
+      else
+        render template: "forms/make_live/new", locals: { make_live_form: MakeLiveForm.new(form: current_form) }
+      end
```
lfdebrux, line 8: Am I right in thinking this code is still doing the same thing after this change, just more explicitly?

aliuk2012: Once draft/live is finished and we remove the flag I dont think there will be a redirect at all. 
We could have a redirect in this controller action if all the required fields are completed or there aren't any changes but lets keep it simple for the time being and extend on.

app/views/forms/make_live/make_your_form_live.html.erb
```diff
@@ -16,7 +16,9 @@
 
       <%= t("make_live.new.body_html", submission_email: @make_live_form.form.submission_email) %>
 
-      <%= render LiveFormWarningComponent::View.new() %>
+      <% if !FeatureService.enabled?(:draft_live_versioning) %>
```
aliuk2012, line 19:
```suggestion
      <% unless FeatureService.enabled?(:draft_live_versioning) %>
```

lfdebrux (approved):

PR #332, opened 2023-02-24
Fix current_form requesting the same form multiple times

aliuk2012:
I discovered while looking at the request that forms-admin were sending to forms-api and noticed every time we called `current_form`, it would make a request for the form instead of memorizing it and reusing the initial response
Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


thomasiles (approved):
nice test 🏅

PR #334, opened 2023-02-25
Switch more env settings

aliuk2012:

Moves RUNNER_BASE, API_BASE, API_KEY and SERVICE_UNAVAILABLE env vars to settings.yml.
Successfully deployed this branch to staging, see forms-deploy PR https://github.com/alphagov/forms-deploy/pull/140

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


config/settings.yml
```diff
@@ -3,6 +3,16 @@ features:
   reorder_pages: true
   draft_live_versioning: false
 
+forms_api:
+  # Authentication key to authenticate forms-runner to forms-api
+  auth_key: changeme
```
danworth, line 8: Just wondering if it's safer to not have a default in the production config just to avoid it accidentally being used if there's some other accident in setting the necessary ENV var override. Feels like it might be safer to not have a default and add a check that a value must be provided?

danworth (approved):
Looks sensible and this is great to move us towards the common way of setting these env vars 👍  I'm not sure whether we should have a default for the API key in production but this moves us in the right direction and we can adjust later if necessary 🏅 🏅

config/settings.yml
aliuk2012: Agreed, I've created a trello card to take a look at forms-api during firebreak and see if we can customise this further. 
I just wanted to replace like for like at the moment.

danworth: Ok cool, the recent deployment failure of this PR because it was merged ahead of the corresponding forms-deploy PR is an example where defaults potentially cause problems. forms-admin deployed without the env var overrides and used the defaults in settings.yml which then broke the environment (my preference would be to fail to deploy and the existing version keep running). Its all fixed and no big deal but shows it'll be a worthwhile firebreak activity to investigate removing the defaults 👍

PR #337, opened 2023-02-28
Add routes to page list view

DavidBiddle:
As part of basic routing we want to play back the routes users have made.
- Adds a feature flag so we can merge this before the API work is finished
- Adds a new component for displaying routes on a draft
- Updates the existing PageOptionsService to show the routes on a live form

Limitations:
- for MVP we're only allowing a single condition per page (see the Trello card for conversation about this).
- The edit links don't go anywhere because we haven't built the add/edit flow yet.
- there's no way to create routes within the app yet, so if you want to test this locally you'll need to:
  - enable the `basic_routing` feature flag in forms-admin/config/settings/development.yml
  - Create conditions manually in the Rails console: 
    - in forms-api, run `bin/rails console` to open the console
    - `Condition.create(check_page: Page.find(15), routing_page: Page.find(15), goto_page: Page.find(17), answer_value: "Wales")` to create a condition

Trello card: https://trello.com/c/Wblu8EV6/539-display-routes-in-page-list

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/components/page_list_component/view.html.erb
+++ b/app/components/page_list_component/view.html.erb
@@ -31,6 +31,30 @@
 
             </dd>
           </div>
+
+          <% if FeatureService.enabled?(:basic_routing) %>
+            <% page.routing_conditions.each do |condition| %>
+              <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key govuk-summary-list__key app-page-list__key">
+                  <%= t("page_conditions.condition_name", page_index: index + 1) %>
+                </dt>
+
+                <dd class="govuk-summary-list__value">
+                  <%= t('page_conditions.condition_html',
+                    check_page_text: question_text_for_page(condition.check_page_id),
+                    answer_value: condition.answer_value,
+                    goto_page_text: question_text_for_page(condition.goto_page_id)
+                  ) %>
+                </dd>
+
+                <dd class="govuk-summary-list__actions govuk-!-padding-bottom-6">
+                  <%= govuk_link_to "#" do %>
+                    <%= t("forms.form_overview.edit_with_visually_hidden_text_html", visually_hidden_text: t("page_conditions.condition_name", page_index: index + 1)) %>
+                  <% end %>
+                </dd>
+              </div>
+            <% end %>
+          <% end %>
         <% end %>
       </dl>
     <% end %>
--- a/app/components/page_list_component/view.rb
+++ b/app/components/page_list_component/view.rb
@@ -15,5 +15,9 @@ module PageListComponent
     def show_down_button(index)
       index != @pages.length - 1
     end
+
+    def question_text_for_page(id)
+      @pages.find { |page| page.id == id }.question_text
+    end
   end
 end
--- a/app/views/live/show_pages.html.erb
+++ b/app/views/live/show_pages.html.erb
@@ -12,7 +12,7 @@
     <%= render FormStatusTagDescriptionComponent::View.new(status: :live) %>
 
     <% form.pages.each_with_index do |page, index| %>
-      <%= render(SummaryCardComponent::View.new( **PageSummaryCardDataService.call(page:).build_data)) %>
+      <%= render(SummaryCardComponent::View.new( **PageSummaryCardDataService.call(page:, pages: form.pages).build_data)) %>
     <% end %>
 
     <p class="govuk-body">
--- a/spec/components/page_list_component/view_spec.rb
+++ b/spec/components/page_list_component/view_spec.rb
@@ -2,52 +2,117 @@ require "rails_helper"
 
 RSpec.describe PageListComponent::View, type: :component do
   let(:pages) { [] }
+  let(:routing_conditions) { [] }
+  let(:page_list_component) { described_class.new(pages:, form_id: 0) }
 
-  before do
-    render_inline(described_class.new(pages:, form_id: 0))
-  end
+  describe "rendering component" do
+    before do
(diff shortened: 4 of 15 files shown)


hannahkc (approved):

app/components/page_conditions_component/view.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      ) %>
+    </dd>
+
+    <dd class="govuk-summary-list__actions govuk-!-padding-bottom-6">
+      <%= govuk_link_to "#" do %>
+        <%= t("forms.form_overview.edit_with_visually_hidden_text_html", visually_hidden_text: t("page_conditions.condition_name", page_index: @index)) %>
+      <% end %>
+    </dd>
+  </div>
+<% end %>
```
aliuk2012, line 21: Could we get rid of the instance vars in the component and pass the values through from `view.rb`?

app/components/page_list_component/view.html.erb
```diff
@@ -31,6 +31,30 @@
 
             </dd>
           </div>
+
+          <% if FeatureService.enabled?(:basic_routing) %>
```
aliuk2012, line 35:
```suggestion
          <% if FeatureService.enabled?(:basic_routing) && page.routing_conditions.any? %>
```

aliuk2012: I was originally trying to think if there would be an issue with older form pages that might not have routing_conditions and if `page.routing_conditions` was `nil` for whatever. Happy for you to ignore the suggestion if you feel theres no chance of this happening and would be pointless.

spec/components/page_list_component/view_spec.rb
```diff
@@ -49,5 +50,39 @@
...
+    context "when conditions are enabled", feature_basic_routing: true do
+      let(:pages) do
+        [OpenStruct.new(id: 1, question_text: "What country do you live in?", routing_conditions:),
+         OpenStruct.new(id: 2, question_text: "What is your name?", routing_conditions:),
+         OpenStruct.new(id: 3, question_text: "What is your pet's name?", routing_conditions:)]
+      end
+
+      context "when there are no conditions" do
+        it "conditions section is blank" do
+          expect(page).not_to have_text("Question 1's route")
```
aliuk2012, line 63: Is there another way we could tag conditions eg. css to html structure? The reason I'm say that is that if for some reason the `t("page_conditions.condition_name")` changed then this test would still pass because the page doesn't contain this specific tests.

app/components/page_list_component/view.rb
```diff
@@ -15,5 +15,9 @@ def show_up_button(index)
     def show_down_button(index)
       index != @pages.length - 1
     end
+
+    def question_text_for_page(id)
+      @pages.find { |page| page.id == id }.question_text
+    end
```
aliuk2012, line 21: Could we maybe add some unit tests for this method and the others above? These should be easier and faster to test than having to try find the results amongst the html thats rendered

app/views/live/show_pages.html.erb
```diff
@@ -12,7 +12,7 @@
     <%= render FormStatusTagDescriptionComponent::View.new(status: :live) %>
 
     <% form.pages.each_with_index do |page, index| %>
-      <%= render(SummaryCardComponent::View.new( **PageSummaryCardDataService.call(page:).build_data)) %>
+      <%= render(SummaryCardComponent::View.new( **PageSummaryCardDataService.call(page:, pages: form.pages).build_data)) %>
```
aliuk2012, line 15: Should we maybe have an example of the Summary Card component displaying routing options?

app/components/page_list_component/view.html.erb
DavidBiddle: Ah, that makes sense - in this case I'm fairly confident there shouldn't be `nil` values since the relevant API PRs have been merged

spec/components/page_list_component/view_spec.rb
DavidBiddle: I don't think there's a CSS/HTML way to do it (the markup is the same for the conditions as for the pages themselves), but can improve this by calling the translation directly instead of hard coding it

aliuk2012: Sure we could improve the tests by using the translations but the thing I'm trying to avoid here is our usage of `.not_to have...` because theres more chance of some not being on the page than there would be if it was expected on the page.

What if the test was to only have 1 question, we could check the page only has one `.govuk-summary-list__row` or actually in this test because there are three questions and they dont have any routes we would expect there to only be 3 `.govuk-summary-list__row` 

```
expect(page).to have_css('.govuk-summary-list__row',count: 3)
```

DavidBiddle: Oh, I see what you mean - thought you were just looking for something more specific than the string for the selector. This does sound more robust, could probably even do:
```
expect(page).to have_css('.govuk-summary-list__row',count: pages.length)
```
to avoid the risk of the counts getting out of sync. Will try locally

DavidBiddle: Done!

app/components/page_list_component/view.rb
DavidBiddle: Done!

app/views/live/show_pages.html.erb
DavidBiddle: Done!

aliuk2012 (approved):
# 🚢 🏅

PR #341, opened 2023-03-03
Refactor form status tag components

lfdebrux:
Rewrites the status tag components to get the tag text from the translation file. This lets us simplify the model a little.

I found I needed to do this as part of working towards the ticket [Add draft/live status to the forms list page](https://trello.com/c/845sNCtW), but I decided to split it out to keep the coming PR neater.

### DIFF
--- a/app/components/form_status_tag_component/view.rb
+++ b/app/components/form_status_tag_component/view.rb
@@ -1,17 +1,21 @@
 module FormStatusTagComponent
   class View < ViewComponent::Base
-    attr_accessor :status
-
-    def initialize(status: "DRAFT")
+    def initialize(status: :draft)
       super
-      @status = status.upcase
+      @status = status.to_sym
     end
 
     def status_colour
       {
         draft: "purple",
         live: "blue",
-      }[status.downcase.to_sym]
+      }[@status]
+    end
+
+    def status_text
+      # i18n-tasks-use t('form_statuses.draft')
+      # i18n-tasks-use t('form_statuses.live')
+      I18n.t("form_statuses.#{@status}")
     end
   end
 end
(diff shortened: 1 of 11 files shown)


app/components/form_status_tag_component/view.rb
```diff
@@ -1,17 +1,21 @@
 module FormStatusTagComponent
   class View < ViewComponent::Base
-    attr_accessor :status
-
-    def initialize(status: "DRAFT")
+    def initialize(status: :draft)
       super
-      @status = status.upcase
+      @status = status.to_sym
```
aliuk2012, line 5: Could we add some tests to show that we convert the status to a symbol? Just wondering what would happened if `to_sym` was accidentally removed and some call to this component passed in a string somewhere

app/components/form_status_tag_component/view.rb
```diff
@@ -1,17 +1,21 @@
...
       {
         draft: "purple",
         live: "blue",
-      }[status.downcase.to_sym]
+      }[@status]
+    end
+
+    def status_text
+      # i18n-tasks-use t('form_statuses.draft')
+      # i18n-tasks-use t('form_statuses.live')
```
aliuk2012, line 17:
```suggestion
```

app/models/form.rb
```diff
@@ -26,9 +24,7 @@ def draft?
   end
 
   def status
-    return STATUSES[:live] if live?
-
-    STATUSES[:draft]
+    live? ? :live : :draft
```
aliuk2012, line 27: Love this refactor. we can do one better in the future and actually calculate it in status in forms-api

app/components/form_status_tag_component/view.rb
lfdebrux: Unfortunately we need these [hints](https://github.com/glebm/i18n-tasks#fine-tuning) :/ I found that without them the `i18n-tasks missing` test didn't warn us if one of these translations was deleted.

app/components/form_status_tag_component/view.rb
lfdebrux: Added

aliuk2012 (approved):

PR #349, opened 2023-03-17
Move frontend build to Vite

DavidBiddle:
## What problem does this PR solve?
This PR makes various improvements to the FE build:
- Removes a few unused dependencies
- no longer relies on Yarn (so if you're calling any of the NPM scripts directly you should now use `npm run test` instead of `yarn test` etc)
- Moves from Gulp/Rollup/Sprockets and the Rails CSS/JSBundling gems to vite-rails, which offers us:
  - fast dev server startup (we don't have a lot of frontend code yet so this isn't a huge issue for us, but it's nice going forward)
  - hot module reloading (i.e. when you change a JS or CSS file it'll update almost immediately on the site without a full page refresh)
    - including SCSS and JS files in the components folder, as long as they're directly or indirectly imported from the entrypoints files
- What I think is a better folder structure (we now have a 'app/frontend' folder, which will include our entry points and any shared JS/CSS, rather than having separate `app/javascript` and `app/assets/stylesheets folders`.
- We're no longer transpiling our JS for older browsers.

## Things worth noting
  - This means it's theoretically possible to for your dev server to do something different from the prod build - I expect this is most likely to happen when you're looking at a browser-specific bug. In these cases instead of running `bin/dev` you'll want to run `bin/vite build` to build the assets and then start a local rails server with `bin/rails s`.
- We're running the JS and CSS linters at the start of every dev build and every time a module gets updated. It's easy enough to take this out if people find it annoying and want to run it manually, as we do with rubocop.

There's more information about this change in the accompanying ADR PR: https://github.com/alphagov/forms/pull/72/files

### Browser support
I've tested in our supported browsers and it seems to be working as expected. Also tested in IE11, where as expected JS doesn't run but that doesn't cause any issues.

### Page speed
I couldn't measure this in ppagespeed insights (because of our Signon integration), but a local Lighthouse audit against the dev server suggested that this has led to no change in our web vitals (for context we were already scoring '100' for performance):

- First Contentful Paint: 0.5 s -> 0.5 s
- Time to Interactive: 0.5 s -> 0.5 s
- Speed Index: 0.6 s -> 0.5 s
- Total Blocking Time: 0 ms -> 0 ms
- Largest Contentful Paint: 0.5 s -> 0.5 s
- Cumulative Layout Shift: 0.001 -> 0.001

### Build time
I measured build time by running `time bin/setup` on both main and the fe-build-changes branches. To make the test as fair as possible in both cases I:
- deleted node_modules and the assets output folder
- changed the bin/setup script to run `bundle install --redownload` instead of `bundle install`

The build time improved slightly from 1:53.31 to 1:30.06.

### Package health
Following this change, the number of direct dependencies flagged as 'Need review' in Snyk advisor has dropped from 7 to 2 (in all cases these are devDependencies, not production dependencies, so they're lower risk anyway).
- before: https://snyk.io/advisor/check/npm/49bda85e-81e5-44ed-a32c-6d50129abd73/needReview
- after: https://snyk.io/advisor/check/npm/47c302e5-a889-45a8-98d7-44c6bddd8720/needReview

Trello card: https://trello.com/c/J9FyujS0/604-improve-frontend-build

#### Checklist

  - [x] README.md


config/vite.json
```diff
@@ -0,0 +1,18 @@
+{
+  "all": {
+    "sourceCodeDir": "app/frontend",
+    "entrypointsDir": "entrypoints",
+    "watchAdditionalPaths": [],
+    "additionalEntrypoints": ["node_modules/govuk-frontend/govuk/assets/**/*"],
```
aliuk2012, line 6: Should the components css/js be included here?

DavidBiddle: Not as entrypoints - they're already included as long as they're being imported in the respective css/js entrypoints.

danworth:
This all seems well considered, I don't know enough to offer an expert review though. Has this been deployed to PaaS dev and or AWS dev (deploying to AWS dev would ensure the docker image is built correctly and runs as expected when the environment is set to 'production' ). If not and you think it's necessary I'm happy to run through how to use the AWS pipelines to deploy dev branches to AWS dev environment, just let me know.

DavidBiddle:
@danworth It's been up to PaaS dev without any issues, I'm up for checking it on AWS dev as well if that's useful. I _think_ I can do that myself now but can shout if I get stuck

lfdebrux:
Note that one might need to run `bin/setup` to get Vite to work (I did)

lfdebrux (approved):

lfdebrux:
Amazing work 🎉

PR #360, opened 2023-03-29
Update forms show and live form show pages for draft/live

lfdebrux:

Trello card: https://trello.com/c/845sNCtW

<details>
<summary><h4>Screenshots</h4></summary>

Screenshot of forms homepage with draft live tags

Screenshot of show live form page with draft edits

</details>

### DIFF
--- a/app/views/forms/index.html.erb
+++ b/app/views/forms/index.html.erb
@@ -15,7 +15,7 @@
 
     <%= table.head do |head|
        head.row do |row|
-        row.cell(header: true, text: t("home.form_name_heading"), width: "full")
+        row.cell(header: true, text: t("home.form_name_heading"))
         row.cell(header: true, text: t("home.form_status_heading"), numeric: true)
       end
     end %>
@@ -31,7 +31,10 @@
             end
           end
           row.cell(numeric: true) do
-            render FormStatusTagComponent::View.new(status: form.status)
+            html = ""
+            html << render(FormStatusTagComponent::View.new(status: :draft)) if form.has_draft_version
+            html << render(FormStatusTagComponent::View.new(status: :live)) if form.has_live_version
+            html.html_safe
           end
         end
       end
--- a/app/views/live/show_form.html.erb
+++ b/app/views/live/show_form.html.erb
@@ -54,7 +54,9 @@
 
     <% if FeatureService.enabled?(:draft_live_versioning) %>
       <p class="govuk-body">
-        <%= govuk_button_link_to t('show_live_form.draft_create'), form_path(form.id) %>
+        <%# i18n-tasks-use t('show_live_form.draft_create') %>
+        <%# i18n-tasks-use t('show_live_form.draft_edit') %>
+        <%= govuk_button_link_to t("show_live_form.draft_#{ form_metadata.has_draft_version ? 'edit': 'create'}"), form_path(form.id) %>
       </p>
     <% end %>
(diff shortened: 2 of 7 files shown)


hannahkc (approved):
Approved from my perspective - though I have only looked at the screenshots! :)

app/views/forms/index.html.erb
```diff
@@ -31,7 +31,10 @@
             end
           end
           row.cell(numeric: true) do
-            render FormStatusTagComponent::View.new(status: form.status)
+            html = ""
+            html << render(FormStatusTagComponent::View.new(status: :draft)) if form.has_draft_version
+            html << render(FormStatusTagComponent::View.new(status: :live)) if form.has_live_version
+            html.html_safe
```
aliuk2012, line 37: Why do we need to concat these two renders? 
Cant we just render each option if form is draft/live etc. like we already have here

app/views/live/show_form.html.erb
```diff
@@ -53,9 +53,15 @@
...
-      </p>
+      <% if form_metadata.has_draft_version %>
+        <p class="govuk-body">
+          <%= govuk_button_link_to t('show_live_form.draft_edit'), form_path(form.id) %>
+        </p>
+      <% else %>
+        <p class="govuk-body">
+          <%= govuk_button_link_to t('show_live_form.draft_create'), form_path(form.id) %>
+        </p>
+      <% end %>
```
aliuk2012, line 64:
```suggestion
        <p class="govuk-body">
          <%= govuk_button_link_to t("show_live_form.draft_#{ form_metadata.has_draft_version? "edit": "create"}"), form_path(form.id) %>
        </p>
```

app/views/forms/index.html.erb
lfdebrux: We want both tags to show if a form has been made live and has a draft with edits.

aliuk2012: Oh, I see now. It might be worth updating the style of the table body block so there isn't a need to concat and then call `html_safe` on that string

```
 <%= table.body do |body| %>
      <% @forms.each do |form| %>
        <% body.row do |row| %>
          <% row.cell do %>
            <% if form.has_live_version %>
              <%= govuk_link_to(form.name, live_form_path(form.id)) %>
            <% else %>
              <%= govuk_link_to(form.name, form_path(form.id)) %>
            <% end %>
          <% end %>
          <%= row.cell(numeric: true) do  %>
            <%= render(FormStatusTagComponent::View.new(status: :draft)) if form.has_draft_version %>
            <%= render(FormStatusTagComponent::View.new(status: :live)) if form.has_live_version %>
          <% end %>
        <% end %>
      <% end %>
    <% end %>
    ```

aliuk2012: or could abstract it out into a component that given a form, it works out which tags to return 🤷‍♂️

lfdebrux: Hmm, I'm not sure about the ERB suggested, it is a lot less readable to my eyes. Abstracting it out could work, but I'm not sure I see the necessity, and it would be a bit of work.

Are you happy for us to leave this code as-is for now and come back to it later?

thomasiles (approved):
nice 👍

deberny (approved):
I've had a click around the journeys and all is working as expected 😄 

LGTM 🚀

PR #365, opened 2023-03-30
Bump node to latest LTS

aliuk2012:

We should use the latest LTS version for node but it also fixes this failure which is affecting all PR failing https://github.com/alphagov/forms-admin/actions/runs/4562963082/jobs/8050846856?pr=364#step:3:106

Also fixes a security risk identified by sonar cloud in our dockerfile which would mean malicious scripts could run as part of npm install. 
Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


.nvmrc
```diff
@@ -1 +1 @@
-16.19
+18.15.0
```
cadmiumcat, line 1:
```suggestion
18.15
```
I think we probably want to allow patches, so pinning the major and minor only.
I'm not 100% sure how we work with PaaS yet but I _think_ the latest buildpack available for nodejs may be 18.14.  I might be really wrong about that/it might not matter here

danworth (approved):
Looks good, docker image is only used on AWS so safe to roll out and check it. If this causes an issue with the pass/github action deployment we can fix it if necessary (since it changes the .nvmrc which likely has some effect there).

.nvmrc
aliuk2012: 🤦‍♂️ Sorry @cadmiumcat  Only just seen your comment after merging it from Dans 👍. 

I've had problems in the past with `asdf`(my local version manager for ruby & node) where it needs to know what minor version we are using. Oddly it doesn't seem to be an issue with forms-runner where the nvmrc pins minor and patch

danworth: Yeah I was wondering the same, I think Node is necessary for the asset compilation which when deploying to PaaS we seem to do that before cf push https://github.com/alphagov/forms-admin/blob/main/.github/workflows/deploy.yml#L24

PR #366, opened 2023-03-30
Update nvmrc to a version nodejs-buildpack supports

aliuk2012:
GOV.UK PaaS is currently using nodejs-buildpack 1.8.6 https://github.com/cloudfoundry/nodejs-buildpack/releases/tag/v1.8.6

This version doesn't support the lts version 18.15.0 so We have had to bump it down.

We know that this node version is different from the version in dockerfile but thats because alphine builds dont have 18.14

Node is only used for asset compiling so we can be a bit more relaxed about versions.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


DavidBiddle:
README.md mentions Node 16 - may be worth updating that either in this PR or a separate one

DavidBiddle (approved):

cadmiumcat (approved):
Yes please
Wondering if putting the trello link on a comment also works?
https://trello.com/c/0z7S98eN/655-migrate-to-node-18

PR #370, opened 2023-04-03
Add pundit

aliuk2012:

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


thomasiles (approved):
awesome 🏅

app/controllers/pages/address_settings_controller.rb
```diff
@@ -1,5 +1,7 @@
 class Pages::AddressSettingsController < PagesController
   def new
+    authorize @form, :can_view_form?
```
DavidBiddle, line 3: Wondering if this could be moved into a before_action? Don't think it would work everywhere but in this controller and some of the others (the settings controllers, PagesController, WhatHappensNextController, SubmissionEmailController) we seem to be calling it consistently at the beginning of the action and it's a bit repetitive.

DavidBiddle (commented):
Looks really good just got a question about whether we can DRY up the `authorize` calls in the controllers

app/controllers/pages/address_settings_controller.rb
aliuk2012: Yeah, I was also a bit on the fence when doing commit fffda12e3ad6a96c75ae34883d7a41374050895b 
and as you will see here https://github.com/alphagov/forms-admin/pull/370/commits/fffda12e3ad6a96c75ae34883d7a41374050895b#diff-d22397cc3f79c2827d23d685cdf4d04de82c72087b07825ecabb264fbe91f82fL2 we used just call it once for PagesController and it applied to all the methods.

Do you reckon we should go with just that single reference in the PagesController?

DavidBiddle: Yeah I think I'd prefer that - seems like it's less repetition, plus means we don't have to remember to do it every time (I know `verify_authorized` should catch that case, but easier not to have to think about it)

DavidBiddle: But not dead set on it if there are reasons not to do it that I'm missing!

PR #372, opened 2023-04-04
SPIKE: Add super admin user concept

aliuk2012:

https://user-images.githubusercontent.com/3441519/229755349-08f0a31d-a70e-4150-a174-5e309859401f.mov

- Adds a super basic `/users` page which can only be accessed if the user manually updates the url and is they are considered a super admin

image

- Adds a super basic "Edit user" page, where a super admin can mark other users as super admins or remove their super admin status

image

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,4 +2,9 @@ class User < ApplicationRecord
   include GDS::SSO::User
 
   serialize :permissions, Array
+
+  enum :role, {
+    super_admin: "super_admin",
+    editor: "editor",
+  }
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,5 @@
 class User < ApplicationRecord
   include GDS::SSO::User
-  attr_accessor :super_admin_user
+
   serialize :permissions, Array
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,5 @@
 class User < ApplicationRecord
   include GDS::SSO::User
-
+  attr_accessor :super_admin_user
   serialize :permissions, Array
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,0 +1,5 @@
+class UsersController < ApplicationController
+  def index
+    render template: "users/index", locals: { users: User.all}
+  end
+end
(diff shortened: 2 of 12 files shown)


app/models/user.rb
```diff
@@ -2,4 +2,7 @@ class User < ApplicationRecord
   include GDS::SSO::User
 
   serialize :permissions, Array
+  def is_super_admin?
+    email.downcase.include?("@digital.cabinet-office.gov.uk")
```
aliuk2012, line 6:
```suggestion
    email.downcase.include?("@digital.cabinet-office.gov.uk")
```

This is a temporary "phase 1" implementation of how we identify super admins.

We would use this is a pundit policy to restrict who can access the list of users and be able to add/edit them.

Once we had setup  super admins initially in prod then we can update this to use the `super_admin_user` value
or even just the signons permission (which I dont think we need to use at all)

app/models/user.rb
```diff
@@ -1,5 +1,5 @@
 class User < ApplicationRecord
   include GDS::SSO::User
-
+  attr_accessor :super_admin_user
```
aliuk2012: dont think we need this

DavidBiddle:
Instead of doing phase 1, could we:
- skip straight to the phase 2 solution
- write a migration to make either of these super admin:
  - all digital.cabinet-office users
  - only our team's accounts (my preference, as it reduces the chances of us making someone we don't know admin by accident)

aliuk2012:
@DavidBiddle 

- write a migration script (dont want to give all gds users admin access and have to manually them to editor,(risk missing one off and end up with super admin) nor do I think having all our team email addresses exposed in the repo is a safe option) 
- use rails console (this would work for PAAS but not for AWS)
- use postman/curl to update the users (we would need to know user id’s and might be tricky to leave those endpoints exposed until we have some people with super admin access)
- give signon admin users the ability to update users and signon admin to update our accounts. Not sure how this would work on AWS

As you can see all the options have their pros/cons but either way you cut it there will have to be some form of phased approach. Its also worth noting that this is a one time exercise that we will have to do to identify super admin users from normal editors

Just wanted to say thank you for question it, it got me thinking about other things I hadn’t thought about like [UR](https://github.com/alphagov/forms-admin/blob/main/app/controllers/application_controller.rb#L37-L41), and local dev (going to have to change https://github.com/alphagov/forms-admin/blob/main/db/seeds.rb#L10) and think about AWS

PR #374, opened 2023-04-04
Pundit errors in User research environment.

aliuk2012:

fixes https://govuk-forms.sentry.io/issues/4061297374/?project=6576190&referrer=issue-stream&stream_index=0

I tested this locally by enabling basic auth on my local build. I was able to replicate the error that was logged in sentry and could test the fix.

By default pundit uses `current_user` which worked perfectly for signon. But for basic auth `current_user` isn't available and so we created an instance var called `@current_user` which we setup for both authentication methods.

https://github.com/varvet/pundit#customize-pundit-user

Trello card: https://trello.com/c/MxEtFS0k/682-fixing-ur-environment

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -101,4 +101,11 @@ class ApplicationController < ActionController::Base
     first_ip_string = forwarded_for.split(",").first
     Regexp.union([Resolv::IPv4::Regex, Resolv::IPv6::Regex]).match(first_ip_string) && first_ip_string
   end
+
+  # By default pundit uses `current_user` which worked when we are using signon
+  # but if we use basic auth `current_user` method isn't set and so we manually
+  # create @current_user and set that.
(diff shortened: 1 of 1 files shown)


danworth (approved):
I have no prior knowledge of pundit or its configuration but the explanation makes sense. It seems like there is a risk that this change could have unexpected side-effects on the signon flow since theres been long standing confusion between the use of `current_user` and `@current_user`, although some work has been done to align on `@current_user` I wouldn't be surprised if there was some unknown interaction behind the scenes that might be effected (especially since Pundit is a new addition to the app)... however if this has been well tested and released carefully then it looks ok to go out. 

I'm curious that no tests needed updating along with this change but all of CI has passed on this PR so I suspect this isn't covered by our tests, should it be?

PR #376, opened 2023-04-05
Introduce super admin and policy

aliuk2012:

Based off this spike https://github.com/alphagov/forms-admin/pull/372, we should be able to merge and deploy this without any impact to our users while the UI is being designed/built

- Adds a new attribute to User table which will be an enum representing a users specific role
- Adds a policy on who can manage users

Trello card: https://trello.com/c/aFI2ii6b/684-create-a-method-to-identify-editors-from-super-admins

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -2,4 +2,9 @@ class User < ApplicationRecord
   include GDS::SSO::User
 
   serialize :permissions, Array
+
+  enum :role, {
+    super_admin: "super_admin",
+    editor: "editor",
+  }
 end
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -1,0 +1,16 @@
+class UserPolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def can_manage_user?
+    # Phase 1
+    user.email.downcase.include?("@digital.cabinet-office.gov.uk")
+
+    # Phase 2
+    # user.super_admin?
+  end
+end
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -1,6 +1,22 @@
 class UserPolicy
   attr_reader :user, :record
 
+  class Scope
+    attr_reader :user, :record, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      if user.email.downcase.include?("@digital.cabinet-office.gov.uk")
+        scope
+          .all
+      end
+    end
+  end
+
   def initialize(user, record)
     @user = user
     @record = record
--- a/db/migrate/20230404075703_add_role_to_users.rb
+++ b/db/migrate/20230404075703_add_role_to_users.rb
@@ -1,0 +1,5 @@
+class AddRoleToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :role, :string, default: "editor"
+  end
+end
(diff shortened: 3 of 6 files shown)


thomasiles (approved):
👍

app/policies/user_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def can_manage_user?
+    # Phase 1
+    user.email.downcase.include?("@digital.cabinet-office.gov.uk")
```
DavidBiddle, line 27: Raised a comment on the spike PR but I guess I'm still not entirely sure what the rationale for the phased approach is - couldn't we do phase 2 and add a migration or a script to make our accounts / `@digital.cabinet-office.gov.uk` accounts into super admins?

aliuk2012: https://github.com/alphagov/forms-admin/pull/372#issuecomment-1497471076

SamJamCul (approved):

app/models/user.rb
```diff
@@ -2,4 +2,9 @@ class User < ApplicationRecord
   include GDS::SSO::User
 
   serialize :permissions, Array
+
+  enum :role, {
+    super_admin: "super_admin",
```
danworth, line 7: A bit late to the party and this is just a question rather than an objection,  but instead of `super_admin` was `system_admin` e.g. us as system administrators, and `org_admin`, e.g. admins for an org to manage their own users considered? Those names have more distinction to me because super_admin isn't immediately clear what scope it has, could be a super user on the org for example, but since its easily clarified with some knowledge sharing its not a big deal.

This is based on my understanding that `super_admin` will eventually just be us and `admin` will be used for org admins?

db/migrate/20230404075703_add_role_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddRoleToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :role, :string, default: "editor"
```
danworth, line 3: The active record migration docs aren't clear on this. Does the `default: "editor"` mean that all subsequent inserts of users to the `users` table will be given `editor` role unless otherwise provided when they're created? Does it mean that all existing user entries will be set to `editor`? If so, is that safe, do we have something to remember to change this since it might not be the right starting role. Perhaps a default of 'none' might be safer, then we actively update them to grant them a role?

danworth (commented):
Just a couple of questions but don't let it stop progress, thanks for showing this at tech catch up earlier 👍

app/models/user.rb
DavidBiddle: We could even make it _really_ clear who these users are by calling it `forms_service_admin` or `service_admin`

db/migrate/20230404075703_add_role_to_users.rb
aliuk2012: @danworth 

> Does the default: "editor" mean that all subsequent inserts of users to the users table will be given editor role unless > otherwise provided when they're created?
Correct

> Does it mean that all existing user entries will be set to editor?  If so, is that safe, do we have something to 
> remember to change this since it might not be the right starting role
Correct, for the time being it should be safe as we are simply allowing signed in users to do what they currently being doing. We haven't really discussed a users with no role, TBH this is kind of where "trial" accounts are more likely to come into play. When we get to that point we can do another migration to change the default role going forward.

app/models/user.rb
aliuk2012: Team decided to review this at another time but I do agree with you that system_admin is a bit clearer what take of admin is actually being done.

PR #377, opened 2023-04-05
Add users screen

thomasiles:
# Add first iteration of user management routes

Trello card: https://trello.com/c/e0HKO8s2/685-create-new-pages-to-show-list-of-users-and-be-able-to-edit-users-role

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


spec/factories/models/users.rb
```diff
@@ -4,5 +4,15 @@
...
     organisation_slug { "test-org" }
+    role { "editor" }
+
+    trait :super_admin do
+      # TODO: remove this once we have a super_admin role
+      email { Faker::Internet.email(domain: "digital.cabinet-office.gov.uk") }
+      # Phase 2, included already because it makes testing for super_admin? now
+      # easier, rather than adding more methods to the model which will change
+      # in phase 2
+      role { "super_admin" }
```
thomasiles, line 15: include phase 1 and 2 here, to make testing easier

spec/support/authentication_helpers.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def login_as_editor_user
+    login_as editor_user
+  end
+
+  def login_as_super_admin_user
+    login_as super_admin_user
+  end
+end
+
```
thomasiles, line 22: [Inspired by](https://github.com/alphagov/content-tagger/blob/79047af5d0ac877fc9ad31ea78bc96d823a8eca0/spec/support/authentication_helpers.rb).

I think these are probably handy - there are questions on style and when to use them.

app/components/header_component/view.html.erb
```diff
@@ -1,6 +1,7 @@
 <% if @current_user %>
   <%= govuk_header navigation_classes: ["govuk-header__navigation--end"] do |header|
     header.product_name(name: t("header.product_name"))
+    header.navigation_item(text: t('users.menu'), href: users_path, active: false) if @current_user.super_admin?
```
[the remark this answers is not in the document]
aliuk2012: I think it would be better if we passed this into the component itself. It will make unit testing the component a lot easier because we wouldn't need to create a user record and mock out these method. we should call the component and pass in a boolean value. In the view, when we call the component, we would use `policy(@current_user).can_manage_user?` which will use the policy to control whether the user can see the link or not. 

When we come to phase 2, we can update the policy and everything should still carry on working. This is the whole point of having a centralised permission policy pattern in that you can go to one place to edit policies that may affect various parts of the system.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,6 @@
+class UsersController < ApplicationController
+  def index
+    authorize current_user, :can_manage_user?
+    render template: "users/index", locals: { users: User.all }
```
aliuk2012, line 4:
```suggestion
    render template: "users/index", locals: { users: policy_scope(User) }
```

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,6 @@
+class UsersController < ApplicationController
+  def index
```
aliuk2012, line 2:
```suggestion
  after_action :verify_authorized
  after_action :verify_policy_scoped, only: :index

  def index
```

app/components/header_component/view.html.erb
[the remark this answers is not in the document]
aliuk2012: Actually, could we maybe pair on this component? I think we should try remove the instance vars being using and `attr_accessors` with the same var names. the reason I'm saying this is that when I first reviewed this I thought `@current_user` would be the exact same object as `@current_user` which we using all over the app and so it wasn't clear that we were actually passing in the `@current_user` and then setting another instance var with exactly the same name. very confusing to follow without viewing the components ruby class and where its being called from.

spec/support/authentication_helpers.rb
aliuk2012: I agree, these would be extremely helpful

spec/factories/models/users.rb
```diff
@@ -4,5 +4,15 @@
     name { Faker::Name.name }
     uid { Faker::Internet.uuid }
     organisation_slug { "test-org" }
+    role { "editor" }
```
aliuk2012, line 7:
```suggestion
    role { :editor }
```

spec/factories/models/users.rb
aliuk2012, line 15:
```suggestion
      role { :super_admin }
```

app/views/users/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          row.cell( text: user.organisation_slug)
+          row.cell( text: t("users.roles.#{user.role}"))
+        end
+      end
+    end %>
+
+  <% end %>
+
+
+  <% end %>
```
aliuk2012, line 31: indenting and spacing looks weird here.

aliuk2012 (approved):
Looks great, only one minor issue with indenting/spacing on one file.

PR #381, opened 2023-04-06
Refactor header component

aliuk2012:

The header component was doing to much than just concentrating on what to display. It was trying to work out what user profile/signout links to use based on what type of authentication method the system was using rather than working that out before calling the component. It meant that the component have dependencies on specific features in the service to work.

This refactor simplifies the component by abstracting that logic and dependencies to an application view helper method. Now the component will render what ever is passed to it. This makes testing easier and should allow us to enhance this component further in the future.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/components/header_component/view.html.erb
+++ b/app/components/header_component/view.html.erb
@@ -1,8 +1,8 @@
-<% if @current_user %>
+<% if is_signed_in? %>
   <%= govuk_header navigation_classes: ["govuk-header__navigation--end"] do |header|
     header.product_name(name: t("header.product_name"))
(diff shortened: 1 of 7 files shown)


app/components/header_component/view.html.erb
```diff
@@ -1,8 +1,8 @@
-<% if @current_user %>
```
aliuk2012, line 1: This was really confusing because just from looking at this file, you may assume that its the instance variable from the view and its not actually passed into this component.  

https://github.com/alphagov/forms-admin/pull/381/files#diff-ac22ef01c268c39ce344f65c3939793ba98084978c87909988d4322090512d6bL7

PR #382, opened 2023-04-06
Add user edit page

thomasiles:
# Add a way for super admins to edit users role

This adds the edit screen and routes to allow super admins to change the role of other users.

Trello card: https://trello.com/c/e0HKO8s2/685-create-new-pages-to-show-list-of-users-and-be-able-to-edit-users-role

It works but these are still todo: 
- Finish the update route to handle missing records, invalid params
- add view tests for the edit page
- extract the roles from the view 
- tidy up the new locale entries a bit

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -13,16 +13,14 @@ class UsersController < ApplicationController
 
   def edit
     authorize current_user, :can_manage_user?
-    @user = User.find(params[:id])
+    user
   end
 
   def update
     authorize current_user, :can_manage_user?
+    user.role = user_params[:role]
 
-    @user = User.find(params[:id])
-    @user.role = user_params[:role]
-
-    if @user.save
+    if user.save
       redirect_to users_path
     else
       render :edit
@@ -33,5 +31,9 @@ private
 
   def user_params
     params.require(:user).permit(:role)
+  end
+
+  def user
+    @user ||= User.find(params[:id])
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,8 +2,36 @@ class UsersController < ApplicationController
   after_action :verify_authorized
   after_action :verify_policy_scoped, only: :index
 
+  rescue_from ActiveRecord::RecordNotFound do
+    render template: "errors/not_found", status: :not_found
+  end
+
   def index
     authorize current_user, :can_manage_user?
     render template: "users/index", locals: { users: policy_scope(User) }
+  end
+
+  def edit
+    authorize current_user, :can_manage_user?
+    @user = User.find(params[:id])
+  end
+
+  def update
+    authorize current_user, :can_manage_user?
+
+    @user = User.find(params[:id])
+    @user.role = user_params[:role]
+
+    if @user.save
+      redirect_to users_path
+    else
+      render :edit
+    end
+  end
+
+private
+
+  def user_params
+    params.require(:user).permit(:role)
   end
 end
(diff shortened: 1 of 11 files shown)


app/views/users/edit.html.erb
```diff
@@ -0,0 +1,45 @@
...
+    end %>
+
+    <%= form_with(model: @user, url: user_path) do |f| %>
+      <% if @user.errors.any? %>
+        <%= f.govuk_error_summary %>
+      <% end %>
+      <span class="govuk-caption-l"><%= @item_name %></span>
+      <%= f.govuk_collection_radio_buttons :role,       [
+        OpenStruct.new(label: t('users.roles.editor.name'), value: :editor, description: t('users.roles.editor.description')),
+        OpenStruct.new(label:t('users.roles.super_admin.name'), value: :super_admin, description: t('users.roles.super_admin.description')),
```
thomasiles, line 35: We should be able to put these somewhere better

app/views/users/edit.html.erb
```diff
@@ -0,0 +1,45 @@
...
+      ], :value,:label, :description,
+      legend: { text: "Role", size: 'm', tag: 'h2' } %>
+
+    <%= f.govuk_submit t('users.save') do
+      govuk_link_to(t('users.cancel'), users_path, class: 'govuk-button govuk-button--secondary')
```
thomasiles, line 40: Is it ok for the cancel button to just be a link?

config/locales/en.yml
```diff
@@ -463,6 +463,14 @@ en:
         role: Role
       title: Users
       users_caption: All users
+    edit:
+      title: Edit user
```
thomasiles, line 467: We already have a titles section in the locales where these could be put

config/brakeman.ignore
```diff
@@ -0,0 +1,29 @@
...
+      "confidence": "Medium",
+      "cwe_id": [
+        915
+      ],
+      "note": ""
+    }
+  ],
+  "updated": "2023-04-11 12:44:26 +0100",
+  "brakeman_version": "5.4.1"
+}
```
thomasiles, line 29: Is this safe? We want to set user role, we are using strong params, is the check just to catch times when that's not the case?

app/views/users/edit.html.erb
lfdebrux: Can we use `govuk_button_link_to` instead?

https://govuk-components.netlify.app/helpers/button/#link-styled-like-a-button

app/controllers/users_controller.rb
```diff
@@ -2,8 +2,36 @@ class UsersController < ApplicationController
...
+    authorize current_user, :can_manage_user?
+
+    @user = User.find(params[:id])
+    @user.role = user_params[:role]
+
+    if @user.save
+      redirect_to users_path
+    else
+      render :edit
+    end
```
aliuk2012, line 29: Can we memorise `@user` something like https://github.com/alphagov/forms-admin/pull/372/files#diff-cfdccd0a9d5df5a43aaad2a35d36ebbe187c52ad5fdc9846fa189d04537adb6eR8-R15
?

app/views/users/edit.html.erb
thomasiles: yes!

config/brakeman.ignore
aliuk2012: This is a weird one. I wonder if it's flagging it up because it thinks it "auth" related, i.e params on "user" and "roles" so would be unsafe to mass assign which we aren't 

Should be alright to just ignore this as you have done.

aliuk2012 (approved):

PR #386, opened 2023-04-12
Remove dependency on edge packages

thomasiles:
We've upgraded to alpine3.17 which means we no longer need edge repo to install node 18 and openssl 3.

We faced issues building ruby packages when we included the edge repo, probably related to changes upstream.

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


Dockerfile
```diff
@@ -1,12 +1,10 @@
-FROM ruby:3.2.0-alpine3.17@sha256:c690d3b45ef65e0098104c272f1736b0595b824908d5639c3a3d17636581a905  AS build
+FROM ruby:3.2.0-alpine3.17@sha256:c690d3b45ef65e0098104c272f1736b0595b824908d5639c3a3d17636581a905 AS build
 
 WORKDIR /app
 
-# Edge repo is necessary for Node 16 and openssl 3
-RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories
 RUN apk update
 RUN apk upgrade --available
-RUN apk add libc6-compat openssl-dev build-base libpq-dev nodejs=18.15.0-r0 npm
```
cadmiumcat, line 9: Is there a reason for removing the specific version of nodejs?

aliuk2012:
I was able to successfully run `docker build .` on my M1.

Dockerfile
aliuk2012: Thats interesting because we have 18.14.1 in the `.nvmrc` file.

We know that in PaaS the buildpack didn't include that version which is why we ended up with 18.15.0 here.


If you leave in the node version  then it seems to want to use node 19 (i.e latest node version rather than LTS version)

cadmiumcat: I though using 19 that was a side effect of `edge`, no? It was installing 18.15.0 fine before

Dockerfile
```diff
@@ -1,12 +1,10 @@
...
+RUN apk add libc6-compat openssl-dev build-base libpq-dev nodejs npm
```
aliuk2012, line 7:
```suggestion
RUN apk add libc6-compat openssl-dev build-base libpq-dev nodejs=18.14.2-r0 npm
```

Dockerfile
aliuk2012: @cadmiumcat `18-14.2-r0` might be coming straight from the alpine version https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.17&repo=&arch=&maintainer=

Dockerfile
aliuk2012: This was tested here https://github.com/alphagov/forms-admin/actions/runs/4677839417/jobs/8285819444

aliuk2012: I suppose we could leave it off

Dockerfile
cadmiumcat: Are we happy with that? [Last time we had decided to go to 18.15.0](https://github.com/alphagov/forms-admin/commit/2c9292839a171bffbe788e2f669687e3a5fdfc81)

thomasiles: This is a good point. I think the correct solution would be to build a specific node version in the container. That is more work and a bit of a hassle though.

It looks like [alpine 3.17 is tracking node 18](https://gitlab.alpinelinux.org/alpine/aports/-/commit/680c2e279b7e926bfe84e7b90e818c7bb11f33fb) and looks like it would stick to node even numbers for LTS.

Could we change our .nvm to just `v18`? Node is only included for building assets and isn't included in the image which runs so we would be able to pickup issues quickly, I think.

aliuk2012 (approved):

cadmiumcat (approved):

PR #389, opened 2023-04-17
Add conditions to pages

aliuk2012:

- Adds two new views
  - `forms/{form id}/new-condition` - a user sees a list of questions that are select one from a list
  - `forms/forms id}/pages/{page id}/conditions/new` - a user can pick an answer and page to redirect to if the form filler selects that answer
- adds a new secondary button to the pages index page so users can add conditions
- adds a new form policy called `can_add_page_routing_conditions?` which is used to controll access to the pages and ui

Trello card: https://trello.com/c/gxgmZmXA/564-add-conditions-to-pages-in-admin

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/views/pages/conditions/new.html.erb
+++ b/app/views/pages/conditions/new.html.erb
@@ -1,0 +1,31 @@
+<% set_page_title(t("page_titles.routing_page")) %>
+<% content_for :back_link, govuk_back_link_to(routing_page_path(form)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <h1 class="govuk-heading-l">
+      <span class="govuk-caption-l"><%= form.name %></span>
+      <%= t("page_titles.routing_page") %>
+    </h1>
+
+    <!-- If the question "where were you born" Change -->
+    <dl class="govuk-summary-list">
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key"><%= t("new_condition.routing_page_text") %></dt>
+        <dd class="govuk-summary-list__value">“<%= page.question_text %>”</dd>
+        <dd class="govuk-summary-list__actions">
+          <%= govuk_link_to routing_page_path(form) do %>
+            <%= t("new_condition.routing_page_change_html") %>
+          <% end %>
+        </dd>
+      </div>
+    </dl>
+
+    <%= form_with(model: nil, url: "#", method: 'POST') do |f| %>
+      <%= f.govuk_collection_select :routing_answer, page.answer_settings.selection_options, :name, :name, label: { text: "is answered as" } %>
+      <%= f.govuk_collection_select :goto_page_id, form.pages, :id, :question_text, label: { text: "take the person to" } %>
+      <%= f.govuk_submit t("save_and_continue") %>
+    <% end %>
+  </div>
+</div>
--- a/app/views/pages/conditions/new.html.erb
+++ b/app/views/pages/conditions/new.html.erb
@@ -9,18 +9,13 @@
       <%= t("page_titles.routing_page") %>
     </h1>
 
-    <!-- If the question "where were you born" Change -->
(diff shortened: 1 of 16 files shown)


app/views/pages/conditions/new.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key"><%= t("new_condition.routing_page_text") %></dt>
+        <dd class="govuk-summary-list__value">“<%= page.question_text %>”</dd>
+        <dd class="govuk-summary-list__actions">
+          <%= govuk_link_to routing_page_path(form) do %>
+            <%= t("new_condition.routing_page_change_html") %>
+          <% end %>
+        </dd>
+      </div>
+    </dl>
```
aliuk2012, line 23: use govuk summary list component

app/views/pages/conditions/new.html.erb
```diff
@@ -9,18 +9,13 @@
...
-            <%= t("new_condition.routing_page_change_html") %>
-          <% end %>
-        </dd>
-      </div>
-    </dl>
+    <%= govuk_summary_list do |summary_list|
+      summary_list.with_row do |row|
+        row.with_key { t("new_condition.routing_page_text") }
+        row.with_value { page.question_text }
+        row.with_action(text: "Change", href: routing_page_path(form), visually_hidden_text: "routing question")
```
DavidBiddle, line 16: Very minor but wonder if we should have a translation for this 'Change' text?

app/views/pages/conditions/new.html.erb
```diff
@@ -0,0 +1,26 @@
+<% set_page_title(t("page_titles.routing_page")) %>
```
DavidBiddle, line 1: Not a blocker for this PR and needs design input, but we should make sure that this page and the next one have different titles

app/views/pages/conditions/new.html.erb
DavidBiddle: I think we use `Edit` everywhere else so wonder if this is another question for the designers. Can probably ignore in this PR

app/views/pages/index.html.erb
```diff
@@ -23,6 +23,9 @@
     <% end %>
     <div class="govuk-button-group">
       <%= govuk_button_link_to t("pages.index.add_question"), type_of_answer_new_path(@form), class:"govuk-!-margin-bottom-3 govuk-!-margin-top-3" %>
+      <% if policy(@form).can_add_page_routing_conditions? %>
+        <%= govuk_button_link_to "Add a question route", routing_page_path(@form), secondary: true, class:"govuk-!-margin-bottom-3 govuk-!-margin-top-3" %>
```
DavidBiddle, line 27: This button text should be a translation, but that's a) my fault and b) something I'm happy to deal with in a separate PR (probably the one where we add the rules for this button appearing)

spec/requests/pages/conditions_controller_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+      ActiveResource::HttpMock.respond_to do |mock|
+        mock.get "/api/v1/forms/1", req_headers, form.to_json, 200
+        mock.get "/api/v1/forms/1/pages", req_headers, pages.to_json, 200
+      end
+
+      if expected_to_raise_error
+        allow(Pundit).to receive(:authorize).and_raise(Pundit::NotAuthorizedError)
+      else
+        allow(Pundit).to receive(:authorize).and_return(true)
+      end
```
DavidBiddle, line 34: Nice solution to an annoying problem!

DavidBiddle (approved):
Looks good! Raised a couple of things related to translations, but I think they can be handled outside this PR 🎉

app/views/pages/index.html.erb
DavidBiddle: Added this to the ACs on the button card

PR #392, opened 2023-04-18
Add condition form class and create action

DavidBiddle:

- Adds a new `Pages::ConditionsForm` class for validations and custom functionality needed on the new conditions page
- Add new `create` post endpoint to create the actual condition record
- basic form validations on the new conditions pages
- adds default nil options to the answer_value and goto_page_id select list
- adds a "None of the above" option to answer_value if the page is optional

Trello card: https://trello.com/c/gxgmZmXA/564-add-conditions-to-pages-in-admin

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/forms/pages/conditions_form.rb
+++ b/app/forms/pages/conditions_form.rb
@@ -1,0 +1,31 @@
+class Pages::ConditionsForm
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  attr_accessor :form, :page, :check_page_id, :routing_page_id, :answer_value, :goto_page_id
+
+  validates :answer_value, :goto_page_id, presence: true
+
+  def submit
+    return false if invalid?
+
+    Condition.create!(form_id: form.id,
+                      page_id: page.id,
+                      check_page_id: page.id,
+                      routing_page_id: page.id,
+                      answer_value:,
+                      goto_page_id:)
+  end
+
+  def routing_answer_options
+    options = page.answer_settings.selection_options
+    options << OpenStruct.new(name: I18n.t("page_options_service.selection_type.none_of_the_above")) if page.is_optional
+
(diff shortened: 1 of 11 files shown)


app/forms/pages/conditions_form.rb
```diff
@@ -0,0 +1,28 @@
...
+  attr_accessor :form, :page, :check_page_id, :routing_page_id, :answer_value, :goto_page_id
+
+  validates :answer_value, :goto_page_id, presence: true
+
+  def submit
+    return false if invalid?
+
+    Condition.create!(form_id: form.id,
+                      page_id: page.id,
+                      check_page_id: page.id, routing_page_id: page.id, answer_value:, goto_page_id:)
```
SamJamCul, line 14: should this be separated out into one line per argument?

spec/factories/models/pages.rb
```diff
@@ -10,7 +10,7 @@ def as_json(*args)
 
 FactoryBot.define do
   factory :page, class: "Page" do
-    association :form
+    id { Faker::Number.number(digits: 2) }
```
SamJamCul, line 13: does this risk overlapping with other pages generated within the same test?

SamJamCul (approved):
All good, just a little cosmetic thing really!

spec/factories/models/pages.rb
DavidBiddle: Yes, but I'm not sure it matters at the moment - we aren't using it anywhere else, and in this PR it's only used to stop `goto_page_options` returning `id: nil` for every page in the form. But the tests will still pass if we remove this line so I don't mind removing it if we think the potential name clash is a concern.

aliuk2012: I think its fine for the moment, we will be visiting this again when we add more functionality

PR #401, opened 2023-04-23
Delete route conditions

aliuk2012:

https://trello.com/c/gxgmZmXA/564-add-conditions-to-pages-in-admin

- Adds delete route view and controller action
- Add destroy controller action that calls forms-api to destroy the condition
- Adds tests for the delete page and controller action
- Adds validation for the deletion form

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/pages/conditions_controller.rb
+++ b/app/controllers/pages/conditions_controller.rb
@@ -54,6 +54,27 @@ class Pages::ConditionsController < PagesController
     render template: "pages/conditions/delete", locals: { condition_form: }
   end
 
+  def destroy
+    condition = Condition.find(params[:condition_id], params: { form_id: @form.id, page_id: page.id })
+
+    condition_form = Pages::ConditionsForm.new(form: @form, page:, record: condition, answer_value: condition.answer_value, goto_page_id: condition.goto_page_id)
+
+    case condition_form_params[:confirm_deletion]
+    when "false"
+      redirect_to edit_condition_path(@form.id, page.id, params[:condition_id])
+    when "true"
+      if condition.destroy
+        redirect_to form_pages_path(@form.id, page.id)
+      else
+        raise StandardError, "Deletion unsuccessful"
+      end
+    else
+      raise StandardError, "Deletion unsuccessful"
+    end
+  rescue StandardError
+    render template: "pages/conditions/delete", locals: { condition_form: }, status: :unprocessable_entity
+  end
+
 private
 
   def can_add_page_routing
@@ -61,6 +82,6 @@ private
   end
 
   def condition_form_params
-    params.require(:pages_conditions_form).permit(:answer_value, :goto_page_id).merge(form: @form, page:)
+    params.require(:pages_conditions_form).permit(:answer_value, :goto_page_id, :confirm_deletion).merge(form: @form, page:)
   end
 end
--- a/app/controllers/pages/conditions_controller.rb
+++ b/app/controllers/pages/conditions_controller.rb
@@ -57,8 +57,6 @@ class Pages::ConditionsController < PagesController
   def destroy
     condition = Condition.find(params[:condition_id], params: { form_id: @form.id, page_id: page.id })
 
-    condition_form = Pages::ConditionsForm.new(form: @form, page:, record: condition, answer_value: condition.answer_value, goto_page_id: condition.goto_page_id)
-
     case condition_form_params[:confirm_deletion]
     when "false"
       redirect_to edit_condition_path(@form.id, page.id, params[:condition_id])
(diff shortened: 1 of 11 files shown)


app/controllers/pages/conditions_controller.rb
```diff
@@ -54,13 +54,30 @@ def delete
...
+    condition = Condition.find(params[:condition_id], params: { form_id: @form.id, page_id: page.id })
+
+    case condition_form_params[:confirm_deletion]
+    when "false"
+      redirect_to edit_condition_path(@form.id, page.id, params[:condition_id])
+    when "true"
+      if condition.destroy
+        redirect_to form_pages_path(@form.id, page.id)
+      else
+        render template: "pages/conditions/delete", locals: { condition_form: }
```
lfdebrux, line 67: Shouldn't it be an error if the condition can't be destroyed?

SamJamCul: I've added the 422 error, I think that'd be the appropriate one here

SamJamCul: Good point! I've also spotted another test set that I should really build out

app/views/pages/conditions/delete.html.erb
```diff
@@ -0,0 +1,50 @@
...
+        </p>
+
+        <p>Do you want this to happen? If you do, select ‘Yes’.</p>
+      <% end %>
+      <%= f.govuk_collection_radio_buttons :confirm_deletion,  [
+        OpenStruct.new(value: "true", label: "Yes"),
+        OpenStruct.new(value: "false", label: "No"),
+      ], :value, :label,
+                                           legend: { text: "Are you sure you want to delete this question route?"}
+      %>
```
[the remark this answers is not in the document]
SamJamCul: Good idea, I think it's better to do it now rather than make fiddly tech debt

app/controllers/pages/conditions_controller.rb
lfdebrux: It feels like something that shouldn't happen in normal operation, so not sure if it needs a test. I think it would only happen if there was something in a callback going on, or a network error? But we should at least log to sentry...

SamJamCul: I've added an error that's raised when the deletion doesn't take place, as well as a test for the destroy controller action.

app/views/pages/conditions/delete.html.erb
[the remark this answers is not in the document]
SamJamCul: I've done some stuff, but there's a couple of bits that I think need to be matched in other parts of the conditions work. I'll write a trello card!

app/controllers/pages/conditions_controller.rb
lfdebrux: Looks good!

package-lock.json
lfdebrux, line 1: A bunch of dependency bumps seems to have snuck in here somehow, could we have these in a separate commit if they're necessary?

lfdebrux (approved):
Couple of small comments but otherwise looks good to me 👍

app/views/pages/conditions/delete.html.erb
```diff
@@ -0,0 +1,46 @@
+<% set_page_title(title_with_error_prefix(t('page_titles.routing_page'), condition_form.errors&.any?)) %>
+<% content_for :back_link, govuk_back_link_to(edit_condition_path(condition_form.form.id, condition_form.page.id, condition_form.record.id)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with(model: condition_form, url: delete_condition_path(condition_form.form.id, condition_form.page.id, condition_form.record.id), method: :delete) do |f| %>
+      <% if condition_form&.errors.any? %>
```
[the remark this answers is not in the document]
DavidBiddle: On searching forms-admin it looks like we use both in various places. I'm not sure we actually need either - I think we always pass a `Pages::ConditionsForm` to this view, and since that model includes `ActiveModel::Validations` it should always have an `errors` method.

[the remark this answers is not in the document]
SamJamCul: I've removed it, can't seem to find a reason we've put it there!

package-lock.json
SamJamCul: It's gone now, synced up with main

app/forms/pages/delete_condition_form.rb
```diff
@@ -0,0 +1,24 @@
...
+    if confirm_deletion == "true"
+      result = record.destroy
+    end
+
+    result
+  end
+
+  def goto_page_options
+    [OpenStruct.new(id: nil, question_text: I18n.t("helpers.label.pages_conditions_form.default_goto_page_id")), form.pages.map { |p| OpenStruct.new(id: p.id, question_text: p.question_text) }].flatten
+  end
```
lfdebrux, line 23: This code looks to be the same as what is in the `Pages::ConditionsForm` class... can we somehow include that method in this class, to avoid the repetition?

SamJamCul: I've changed it so that it's a better suited method for the delete form - please could you review @lfdebrux

aliuk2012: I've reviewed this again and it looks good to merge

PR #404, opened 2023-04-26
Add errors to conditions in page list component

DavidBiddle:
Adds error messages to the page list component, so that users can see which routes have errors at a glance.

Doesn't include the error summary or the errors on individual route pages.

Trello card: https://trello.com/c/aiXXPe86/674-error-display-in-admin-for-basic-routing

TODO:

- [x] link errors to specific fields
- [x] switch between 'For the question...' and 'If the question...' if answer value is in error
- [x] display errors in a UL if answer value is in error
- [x] remove logic for adding 'and' text

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/components/page_list_component/_index.scss
+++ b/app/components/page_list_component/_index.scss
@@ -9,5 +9,21 @@
 }
 
 .app-page-list__key {
-  width: 15%;
+  @include govuk-media-query($from: desktop) {
+    width: 20%;
+  }
+}
+
+.app-page-list__row--error {
+  border-left: 5px solid $govuk-error-colour;
+  padding-left: 10px;
+  @include govuk-media-query($from: tablet) {
+    padding-left: 0;
+  }
+}
+
+.app-page-list__row--error .app-page-list__key {
+  @include govuk-media-query($from: tablet) {
+    padding-left: 10px;
+  }
 }
--- a/app/components/page_list_component/_index.scss
+++ b/app/components/page_list_component/_index.scss
@@ -14,16 +14,13 @@
   }
 }
 
-.app-page-list__row--error {
-  border-left: 5px solid $govuk-error-colour;
-  padding-left: 10px;
+.govuk-form-group--error .app-page-list__key {
   @include govuk-media-query($from: tablet) {
-    padding-left: 0;
+    padding-left: govuk-spacing(3);
   }
 }
 
-.app-page-list__row--error .app-page-list__key {
-  @include govuk-media-query($from: tablet) {
-    padding-left: 10px;
-  }
+.app-page_list__route-text--error:link,
+.app-page_list__route-text--error:visited {
+  color: $govuk-error-colour;
 }
--- a/app/components/page_list_component/view.rb
+++ b/app/components/page_list_component/view.rb
@@ -19,5 +19,35 @@ module PageListComponent
     def question_text_for_page(id)
       @pages.find { |page| page.id == id }.question_text
     end
+
+    def error_id(number)
+      "condition_#{number}"
+    end
+
+    def error_link(error_name, edit_link)
+      content_tag(
+        :a,
+        I18n.t("page_conditions.errors.#{error_name}"),
+        class: "govuk-link app-page_list__route-text--error",
+        href: edit_link,
+      )
+    end
+
+    def answer_value_text_for_condition(condition, edit_link)
+      if condition.errors_include("answer_value_doesnt_exist")
+        error_link("answer_value_doesnt_exist", edit_link)
+      else
+        t("page_conditions.condition_answer_value_text", answer_value: condition.answer_value)
+      end
+    end
+
+    def goto_page_text_for_condition(condition, edit_link)
+      if condition.errors_include("goto_page_doesnt_exist")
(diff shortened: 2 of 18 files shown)


app/components/page_list_component/_index.scss
```diff
@@ -9,5 +9,26 @@
 }
 
 .app-page-list__key {
-  width: 15%;
+  @include govuk-media-query($from: desktop) {
+    width: 20%;
+  }
+}
+
+.app-page-list__row--error {
```
aliuk2012, line 17: Can we not use the same class that is used for input fields that have errors or maybe use govuk spacing scale too https://design-system.service.gov.uk/styles/spacing/?

DavidBiddle: The padding on the input error class doesn't get on with the summary list's `display: table-row;` declaration on wider screen sizes, but good point about the spacing scale!

DavidBiddle: Actually, we can use the `govuk-form-group--error` class as you suggest - and then add a more targeted fix for the bit where it breaks. Great spot 🙂

app/components/page_list_component/_index.scss
```diff
@@ -9,5 +9,18 @@
...
+
+.govuk-form-group--error .app-page-list__key {
+  @include govuk-media-query($from: tablet) {
+    padding-left: govuk-spacing(3);
+  }
+}
+
+.app-page_list__route-text--error:link,
+.app-page_list__route-text--error:visited {
+  color: $govuk-error-colour;
```
aliuk2012, line 25: Doesn't this cause an A11y issue? Being able to identify if a user has visited a page before?

app/components/page_list_component/view.rb
```diff
@@ -19,5 +19,35 @@ def show_down_button(index)
...
+      if condition.errors_include?("answer_value_doesnt_exist")
+        error_link("answer_value_doesnt_exist", edit_link)
+      else
+        t("page_conditions.condition_answer_value_text", answer_value: condition.answer_value)
+      end
+    end
+
+    def goto_page_text_for_condition(condition, edit_link)
+      if condition.errors_include?("goto_page_doesnt_exist")
+        error_key = condition.errors_include?("answer_value_doesnt_exist") ? "goto_page_doesnt_exist_and_nor_does_answer_value" : "goto_page_doesnt_exist"
```
aliuk2012, line 46: do we need all this logic? Just thinking that the api should be responsible for api validation errors and would return a locale key and we could use that and not have to make up new ones in forms-admin

config/locales/en.yml
```diff
@@ -316,12 +316,15 @@ en:
...
-    condition_html: |
-      If the question “%{check_page_text}”<br>
-      is answered as “%{answer_value}”<br>
-      take the person to “%{goto_page_text}”
+    condition_goto_page_text: take the person to “%{goto_page_text}”
     condition_name: Question %{page_index}'s route
+    errors:
+      answer_value_doesnt_exist: select the answer you want to base your route on
+      goto_page_doesnt_exist: select the question or page you want to take the person to, or delete this route
+      goto_page_doesnt_exist_and_nor_does_answer_value: and select the question or page you want to take the person to, or delete this route
```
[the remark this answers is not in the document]
DavidBiddle: Yeah, I see what you mean. I'm not 100% sure how to handle this case, other than the way I've done it here or having a separate 'and' translation that appears and disappears depending whether or not both errors are present.

app/components/page_list_component/view.rb
DavidBiddle: I think even if we were passing the full message from the API a lot of the logic would still be needed, since we'd still need to decide whether to render a link or a normal string for each part of the condition unlkess we start sending the HTML from the API. But I do think it's worth keeping the translations in `forms-admin` since it'll a) make sure all of our translations are in one place and easier to update by content designers, and b) make things easier if we do need to add support for Welsh or other languages in future.

Would be interested in knowing your thoughts on moving some of this logic into the view - as I understand it it's cheaper to test in the view.rb file, but I think it might possibly be marginally more readable in view.html.erb, so I'm open to suggestions.

config/locales/en.yml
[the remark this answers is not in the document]
aliuk2012: We are pushing back on this so this shouldn't be needed for the time being

app/components/page_list_component/view.rb
aliuk2012: We are pushing back on this so this shouldn't be needed for the time being

app/models/condition.rb
```diff
@@ -5,4 +5,21 @@ class Condition < ActiveResource::Base
...
+  def has_errors_for_field?(field)
+    errors_with_fields.filter { |error| error[:field] == field }.any?
+  end
+
+  def has_errors?
+    validation_errors.any?
+  end
+
+  def errors_include?(error_name)
+    has_errors? && validation_errors.map(&:name)&.include?(error_name)
```
aliuk2012, line 23: Why do we have a safe nav operator here? are we expecting the map to be nil?

aliuk2012 (approved):

PR #405, opened 2023-04-26
Make the forbidden error message generic

thomasiles:
The forbidden error message is shown whenever a pundit policy fails. For example, it's shown when a user tries to access a from from another organisation and when en Editor tries to access the /users page.

The current error message was written for the organisation case.

This commit updates the text so that the same page works for both.

image

Trello card: https://trello.com/c/zPSGHu5z/693-update-error-message-editor-user-would-see-if-they-tried-to-access-pages-that-relate-to-managing-users

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


lfdebrux (approved):

hannahkc:
Looks good to me!

PR #408, opened 2023-04-27
Add organisation table

lfdebrux:

To be able to update a user's organisation within our service we need a separate organisation table, because `gds-sso` gem will revert any changes to organisation whenever a user logs in [[1]].

This PR creates the `Organisation` model, which has many users, and populates it with organisations of current users.

We assume that we are going to be using the same list of organisations (managed by GOV.UK [[2]]) that Signon uses, so the organisation table has the `content_id` as well as `slug`, as the content ID is a stable UUID while the slug can occasionally change. We also keep the organisation name in the table, to make it nicer for superadmins when they are changing the organisation of a user (feature to come in a subsequent PR).

Note that in general this PR handles the minimum needed to handle our current users and organisations; we don't have a way of adding new organisations to the table. We also don't do anything about organisations in form objects, as I think some changes to the API will be needed for that first. Tickets to fill in some of the gaps are already in the backlog however, so this state of affairs shouldn't persist for too long.

Also note that after merging this PR, new users will not be able to use the service as they will get the "Your organisation is not set" error. Until #414 is merged we will not have a way to assign users to an organisation.

[1]: https://github.com/alphagov/gds-sso/blob/a9c2a018dbebd07c35682b1523a2ad27a41d1185/lib/gds-sso/user.rb#L22
[2]: https://www.api.gov.uk/gds/gov-uk-organisations/#gov-uk-organisations

### DIFF
--- a/app/policies/form_policy.rb
+++ b/app/policies/form_policy.rb
@@ -7,19 +7,21 @@ class FormPolicy
     attr_reader :user, :form, :scope
 
     def initialize(user, scope)
+      raise UserMissingOrganisationError, "Missing required attribute organisation_slug" if user.organisation_slug.blank?
+
       @user = user
       @scope = scope
     end
 
     def resolve
-      raise UserMissingOrganisationError, "Missing required attribute organisation_slug" if user.organisation_slug.blank?
-
       scope
         .where(org: user.organisation_slug)
     end
   end
 
   def initialize(user, form)
+    raise UserMissingOrganisationError, "Missing required attribute organisation_slug" if user.organisation_slug.blank?
+
     @user = user
     @form = form
   end
--- a/app/policies/form_policy.rb
+++ b/app/policies/form_policy.rb
@@ -35,6 +35,7 @@ class FormPolicy
 private
 
   def users_organisation_owns_form
-    user.organisation_slug == form.org
+    organisation_slug = user.organisation&.slug || user.organisation_slug
+    organisation_slug == form.org
   end
 end
--- a/db/migrate/20230426150134_create_organisations.rb
+++ b/db/migrate/20230426150134_create_organisations.rb
@@ -1,0 +1,14 @@
+class CreateOrganisations < ActiveRecord::Migration[7.0]
+  def change
+    create_table :organisations do |t|
+      t.string :content_id, null: false
+      t.string :slug, null: false
+      t.string :name, null: false
+
+      t.timestamps
+    end
+
+    add_index :organisations, :content_id, unique: true
+    add_index :organisations, :slug, unique: true
+  end
+end
--- a/db/migrate/20230426150950_add_organisation_ref_to_users.rb
+++ b/db/migrate/20230426150950_add_organisation_ref_to_users.rb
@@ -1,0 +1,8 @@
+require "json"
+require "net/http"
+
+class AddOrganisationRefToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_reference :users, :organisation, null: true, foreign_key: true
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -9,12 +9,13 @@
 require "factory_bot"
 
 if HostingEnvironment.local_development? && User.none?
-  Organisation.create!(content_id: "af07d5a5-df63-4ddc-9383-6a666845ebe9", slug: "government-digital-service", name: "Government Digital Service")
+  gds = Organisation.create!(content_id: "af07d5a5-df63-4ddc-9383-6a666845ebe9", slug: "government-digital-service", name: "Government Digital Service")
 
   # Create default super-admin
   User.create!({ email: "example@example.com",
                  organisation_slug: "government-digital-service",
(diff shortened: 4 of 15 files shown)


app/policies/form_policy.rb
```diff
@@ -35,6 +35,7 @@ def can_add_page_routing_conditions?
 private
 
   def users_organisation_owns_form
-    user.organisation_slug == form.org
+    organisation_slug = user.organisation&.slug || user.organisation_slug
```
aliuk2012, line 38: Why are we using safe navigation operator here? I would rather add a guard to the method so only execute the rest of this method if the user has an organisation or organisation_slug

db/seeds.rb
```diff
@@ -9,9 +9,13 @@
 require "factory_bot"
 
 if HostingEnvironment.local_development? && User.none?
+  Organisation.create!(id: 0, content_id: "af07d5a5-df63-4ddc-9383-6a666845ebe9", slug: "government-digital-service", name: "Government Digital Service")
```
aliuk2012, line 12: Where has this content_id come from? Is it worth maybe placing it into the `settings.yml`?
Could we also use id 1 rather than 0?

```suggestion
  Organisation.create!(id: 1, content_id: "af07d5a5-df63-4ddc-9383-6a666845ebe9", slug: "government-digital-service", name: "Government Digital Service")
```

lfdebrux: `content_id` comes from https://www.gov.uk/api/organisations/government-digital-service. I don't think we'd ever need to change it.

Sure we can change `id`, what's the reason for using 1 instead of 0?

db/migrate/20230426150950_add_organisation_ref_to_users.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      dir.down do
+        User.where.not(organisation_content_id: nil).find_each do |user|
+          organisation = Organisation.find_by(content_id: user.organisation_content_id)
+          if organisation
+            user.update!(organisation_slug: organisation.slug)
+          end
+        end
+      end
+    end
```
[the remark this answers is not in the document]
lfdebrux: Okay will do

db/seeds.rb
aliuk2012: That makes sense, might be worth still sticking that value into settings.yml with a comment above it explaining where that figure comes from.

Rails tends to start id off using 1 rather than 0.

db/migrate/20230426150950_add_organisation_ref_to_users.rb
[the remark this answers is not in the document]
lfdebrux: Done

db/migrate/20230426150134_create_organisations.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    reversible do |dir|
+      dir.up do
+        Organisation.create(content_id: "af07d5a5-df63-4ddc-9383-6a666845ebe9", slug: "government-digital-service", name: "Government Digital Service")
+      end
+
+      dir.down do
+        Organisation.delete_all
+      end
+    end
```
aliuk2012, line 22: Thanks for splitting up the other migration. Could you do the same here?

lfdebrux: Done.

app/policies/form_policy.rb
lfdebrux: The reason we're using the safe navigation operator is because users might not have `organisation_id` set in the table, if they sign in for the first time after this PR has been merged their user record won't be linked to the organisation record.

aliuk2012: I was thinking something along the lines of adding ` raise Pundit::NotAuthorizedError` at the top of method if the user doesn't have any sort of org assigned to their profile. Users should only really be allowed to view forms if they have an org assigned to them and it matches the forms org.

I would rather handle errors or issues if `user.organisation` is  nil rather than just suppressing them.

lfdebrux: Doesn't the `resolve` method in the policy already raises an error if the user doesn't have an organisation set?

https://github.com/alphagov/forms-admin/blob/39a6005da4eab03b36e1cecff8649a2e061efc32/app/policies/form_policy.rb#L15

aliuk2012: `resolve` is part of the policy subclass called `Scope` which is used when you want to return scope the data you are returning back to the user. See https://github.com/alphagov/forms-admin/blob/main/app/controllers/forms_controller.rb#L10 where we scope the forms to just the ones in their org.

🤔  we could maybe move that further up the chain and place it into the different `initialize` methods instead.

aliuk2012: Nice, That's much better than using the safe navigation operator and shows a user has to have at least an organisation to be able to do anything with the forms.

aliuk2012: Could we remove the save navigation operator now?

aliuk2012 (approved):

PR #414, opened 2023-05-03
Add organisation select to edit user form

lfdebrux:

We want superadmins to be able to assign users to organisations. Change the `/users` page in forms-admin to add a [select](https://govuk-form-builder.netlify.app/form-elements/select/) dropdown to choose organisation for user.

#### Screenshots

Screenshot of /users/:id/edit with the changes from this PR

### DIFF
--- a/spec/features/users/set_organisation_spec.rb
+++ b/spec/features/users/set_organisation_spec.rb
@@ -1,0 +1,100 @@
+require "rails_helper"
+
+feature "Set or change a user's organisation", type: :feature do
+  let(:gds_forms) do
+    [build(:form, id: 1, org: "government-digital-service", name: "Test GDS Form")]
+  end
+
+  let(:test_org_forms) do
+    [build(:form, id: 2, org: "test-org", name: "Test Org Form")]
+  end
+
+  let(:req_headers) do
+    {
+      "X-API-Token" => Settings.forms_api.auth_key,
+      "Accept" => "application/json",
+    }
+  end
+
+  before do
+    Rails.application.load_seed
+
+    ActiveResource::HttpMock.respond_to do |mock|
+      mock.get "/api/v1/forms?org=government-digital-service", req_headers, gds_forms.to_json, 200
+      mock.get "/api/v1/forms?org=test-org", req_headers, test_org_forms.to_json, 200
+    end
+
+    login_as_super_admin_user
+  end
+
+  scenario "Super admin can change a user's organisation" do
+    given_i_am_viewing_the_users_page
+    and_i_choose_a_user_to_edit
+    when_i_change_the_users_organisation
+    then_the_users_organisation_name_is_updated
+  end
+
+  scenario "Super admin can change their own organisation" do
+    given_i_am_a_super_admin_in_the_government_digital_service_organisation
+    when_i_change_my_organisation_to_test_org
+    then_i_cannot_see_the_government_digital_service_forms
+    but_i_can_see_the_test_org_forms
+  end
+
+private
+
+  def given_i_am_viewing_the_users_page
+    visit users_path
+  end
+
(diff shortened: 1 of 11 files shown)


app/forms/users/edit_user_form.rb
```diff
@@ -23,4 +26,23 @@ def user_role_options(roles = User.roles.keys)
...
+      OpenStruct.new(value: slug, label: name)
+    end
+  end
+
+private
+
+  def available_user_organisations
+    organisations = Organisation.all.pluck(:slug, :name)
+    organisations.sort_by! { |slug, _| slug }
+    organisations
```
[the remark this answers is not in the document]
lfdebrux: > Couple of minor syntax questions here:
> 
> 1. Do you need the `_` argument for `sort_by!` or can you just do `{ |slug| slug }`

So what I thought I was doing with `|slug, _|` was using Array decomposition to get the first element of each Array in the Array of organisations (because organisations would be something like `[['test-org', 'Test Org'], ['department-for-tests', 'Department For Tests']]`). That is what is happening, but it turns out that a) the documentation for Ruby suggests you need to put parentheses around the arguments for it to work (so `|(slug, _)|`) [[1](https://docs.ruby-lang.org/en/3.0/syntax/methods_rdoc.html#label-Array+Decomposition)] b) the definition of the `<=>` operator for Array means that a normal `sort!` would probably work just as well, as it tests each element in each Array in turn [[2](https://docs.ruby-lang.org/en/3.0/Array.html#method-i-3C-3D-3E)] (and we're not going to have two orgs with the same slug).

I'm kind of torn, I prefer the explicitness of saying "we're going to sort by slug", but if it isn't obvious from what I've written already then maybe it doesn't help much.

> 2. Is it possible to return `organisations.sort_by! { |slug, _| slug }`, or do you need the `organisations` line?

Yes, you're right, `organisations` line is not necessary, `sort_by!` returns `self`.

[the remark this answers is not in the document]
DavidBiddle: I think what you've said about being explicit about sorting by slug makes sense. I think I'd have understood it / been more likely to look up the decomposition syntax with the parentheses there.  But as you say, it's probably easier (and I think possibly faster?) to do the ordering on the DB side.

deberny:
The Edit user page looks all good 🎉

deberny (approved):

spec/features/users/set_organisation_spec.rb
aliuk2012, line 1: Could you switch the instance vars to `let`? I'm surprised that rubocop didn't actually complain about it but we shouldn't be using instance vars in specs.

spec/policies/form_policy_spec.rb
```diff
@@ -64,6 +66,17 @@
 
     let(:gds_forms) { build_list :form, 2, org: "gds" }
 
+    it "uses organisation slug to scope what forms a user can see" do
```
aliuk2012, line 69: This feels like we are testing implementation rather than behaviour that we are expecting from the scope policy.  

We shouldn't care how we get a list of forms owned by the user but more of a case of "only return a list of forms the user should be able to see.

lfdebrux: The reason for this test was to catch a very specific issue; I forgot to change `user.organisation_slug` in the policy scope to `user.organisation.slug`, and spent ages trying to figure it out because it wasn't super obvious. I added feature tests in https://github.com/alphagov/forms-admin/pull/414/files#diff-7765f51619f0b6c7cb0ecacea452722004fcbc8bca758bdc4b35ef64614c476aR37 to test the behaviour (and they helped me track down the issue), but I wanted to make sure the test was very clear on what the implementation should be, so it would be clear if someone accidentally changed the code what was going on.

aliuk2012 (approved):

PR #415, opened 2023-05-03
Add role to summary list on edit user page

lfdebrux:
Part of https://trello.com/c/6qASsHQl

To make the summary table consistent we're adding the role to the summary table, and keeping the organisation in the table when there is a form input for changing the user organisation.

See designs on Mural: https://app.mural.co/t/gaap0347/m/gaap0347/1680614423424/1147a31af301a59c17158cb623d11391a95cff62?wid=0-1683116546338

### Screenshot
Screenshot of `/users/<id>/edit` with changes in this PR

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -23,6 +23,11 @@
         row.with_key { t('users.index.table_headings.organisation') }
         row.with_value { @user.organisation_slug }
       end
+
+      summary_list.with_row do |row|
+        row.with_key { t('users.index.table_headings.role') }
+        row.with_value { t("users.roles.#{@user.role}.name") }
+      end
     end %>
 
     <%= form_with(model: @user) do |f| %>
--- a/spec/views/users/edit.html.erb_spec.rb
+++ b/spec/views/users/edit.html.erb_spec.rb
@@ -30,6 +30,10 @@ describe "users/edit.html.erb" do
     it "contains org" do
       expect(summary_list).to have_text(user.organisation_slug)
     end
+
+    it "contains role" do
+      expect(summary_list).to have_text("Editor")
+    end
   end
 
   it "has form fields" do
--- a/spec/views/users/edit.html.erb_spec.rb
+++ b/spec/views/users/edit.html.erb_spec.rb
@@ -14,16 +14,22 @@ describe "users/edit.html.erb" do
     expect(rendered).to have_css("h1.govuk-heading-l", text: /Edit user/)
   end
 
-  it "contains name" do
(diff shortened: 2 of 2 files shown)


app/views/users/edit.html.erb
```diff
@@ -23,6 +23,11 @@
         row.with_key { t('users.index.table_headings.organisation') }
         row.with_value { @user.organisation_slug }
       end
+
+      summary_list.with_row do |row|
+        row.with_key { t('users.index.table_headings.role') }
+        row.with_value { t("users.roles.#{@user.role}.name") }
+      end
```
aliuk2012, line 30: Hi @lfdebrux @deberny, 

Could I ask why are we displaying org & roles as readonly state at the top of the page and then repeating the information again but as a form field? I thought we would have just had the form fields to show the user what is it currently as with any edit form page.

deberny: Screenshot 2023-05-04 at 09 31 27

We are showing the User details in a summary at the top, playing back what is the current state.

And under organisation we allow the superadmin to change or assign the organisation.
Under role we allow the superadmin to change the user role.

This is not a typical edit page, it's more of a mix of 

- here are the details
- here you can change them

aliuk2012: Sorry for wanting to dig in a bit deeper to the needs. I'm sure this was already covered in other meetings so happy with what the team have agreed on.

spec/views/users/edit.html.erb_spec.rb
```diff
@@ -14,16 +14,26 @@
     expect(rendered).to have_css("h1.govuk-heading-l", text: /Edit user/)
   end
 
-  it "contains name" do
-    expect(rendered).to have_text(user.name)
-  end
+  describe "summary list" do
+    let(:summary_list) do
+      Capybara.string(rendered).find(".govuk-summary-list")
+    end
```
aliuk2012, line 20: I love how you have scoped the component here to test the various attributes being played back on the page

aliuk2012 (approved):

app/views/users/edit.html.erb
deberny: It's important to display that **'Organisation is not set'** in the summary and users list column, so that the admin can tell the difference and go in to set it.

Playing back the role as a part of summary list in key-value pattern will make it easier to scan with assistive tech.  I wouldn't want the user to have to rely on navigating to editing fields with 'Radio option checked' to learn what the current value is.

Otherwise we's need to change the page to only playback the read only details, add 'change' action link to the summary list, and build two more pages to 'Edit role' and 'Edit organisation'

PR #417, opened 2023-05-04
Align Github WorkFlows Between Apps

danworth:
1. Rename files for consistent `snake_case`.
2. Add a note that the .ruby-version is used to install the correct version of ruby.

Improves consistency between our apps which makes maintaining simpler. Documentation for the ruby/setup-ruby action: https://github.com/ruby/setup-ruby#single-job

Also see: https://github.com/alphagov/forms-runner/pull/311


lfdebrux (approved):
🚢

.github/workflows/test.yml
```diff
@@ -46,19 +50,7 @@ jobs:
...
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
-      - name: Install Ruby and gems
-        uses: ruby/setup-ruby@d2b39ad0b52eca07d23f3aa14fdf2a3fcc1f411c # v1.117.0
-        with:
-          bundler-cache: true
-      - name: Install frontend dependencies
-        run: npm ci
```
aliuk2012, line 61: Can we also delete https://github.com/alphagov/forms-admin/blob/main/package.json#L18  if we arent calling `npm ci`?

danworth: No because we're still calling `npm ci` on line 43 above, its just that we're no longer doing it twice.

PR #421, opened 2023-05-05
Refactor edit user page

lfdebrux:
Part of https://trello.com/c/6qASsHQl

Refactors the controller and view of the page `/user/:id/edit` to use a form object.

This makes it easier to add validation, and also makes PR #414 easier as we can put some extra logic between the form response and the model.


aliuk2012 (changes requested):
I dont think we need to create a user form object when we could just use the User model.

lfdebrux:
> I dont think we need to create a user form object when we could just use the User model.

Okay, I've pulled that change.

lfdebrux:
Had to close this due to GitHub being down and getting out of sync, recreating at #424

PR #425, opened 2023-05-09
Show user's organisation name in table caption on homepage

lfdebrux:
Make it easier for users to see when their organisation has been changed by showing the name of the organisation in the table caption.

Trello card: https://trello.com/c/KaIq9kJN

### Screenshot

Screenshot of forms-admin homepage

### DIFF
--- a/spec/views/forms/index.html.erb_spec.rb
+++ b/spec/views/forms/index.html.erb_spec.rb
@@ -1,9 +1,11 @@
 require "rails_helper"
 
 describe "forms/index.html.erb" do
+  let(:user) { build :user }
   let(:forms) { [] }
 
   before do
+    assign(:current_user, user)
     assign(:forms, forms)
     render template: "forms/index"
   end
@@ -14,7 +16,7 @@ describe "forms/index.html.erb" do
     end
 
     it "does not contain a a list of forms" do
-      expect(rendered).not_to have_text "Your forms"
+      expect(rendered).not_to have_table
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/views/forms/index.html.erb_spec.rb
```diff
@@ -14,7 +16,7 @@
     end
 
     it "does not contain a a list of forms" do
-      expect(rendered).not_to have_text "Your forms"
+      expect(rendered).not_to have_table
       expect(rendered).not_to have_css ".govuk-summary-list"
```
aliuk2012, line 20: Is this check still needed? I thought we only using a table on the page now.

lfdebrux: I don't know, if that was the case that change was made before I joined the team...

spec/views/forms/index.html.erb_spec.rb
```diff
@@ -14,7 +16,7 @@
     end
 
     it "does not contain a a list of forms" do
-      expect(rendered).not_to have_text "Your forms"
+      expect(rendered).not_to have_table
```
aliuk2012, line 19: Is there any other content on the page that we could also check to ensure that its not returning a list of forms?

The problem with this test is that its will still pass if we change the content from using a table to was an `<ul>` or even summary card components and for some reason it outputs content to page.

spec/views/forms/index.html.erb_spec.rb
aliuk2012: I suspect this is overhang when the list of forms were using summary list component. It might be worth removing this line as we aren't using summary list component any more. 

This is one reason why we need to be careful using `not_to` in the specs they can result in false positives.

aliuk2012 (approved):
Only a couple of comments relating to specs but otherwise looks great.

spec/views/forms/index.html.erb_spec.rb
lfdebrux: I've removed it.

PR #426, opened 2023-05-09
Fix table caption in case where user does not have an organisation

lfdebrux:
Fixes PR #425, [alert #11127](https://governmentdigitalservice.pagerduty.com/incidents/Q2PZ1FGZZXTVFB), trello card https://trello.com/c/KaIq9kJN

We forgot to handle the case where the user does not have an organisation, this PR restores the original 'Your forms' caption in that situation.

### DIFF
--- a/app/views/forms/index.html.erb
+++ b/app/views/forms/index.html.erb
@@ -11,7 +11,7 @@
 
 <% if @forms.any? %>
   <%= govuk_table do |table| %>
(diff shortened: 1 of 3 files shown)


danworth (approved):
Seems to fix the case in question. Perhaps separately if you need to get this out to fix things up, we should log when a user doesn't have an organisation because I think that's an exception which we would need to take action on (since our authorisation methods would rely on it?)

PR #429, opened 2023-05-11
Add generic custom error summary component

DavidBiddle:

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


app/components/error_summary_component/view.html.erb
```diff
@@ -0,0 +1,18 @@
+<% if @errors.any? %>
```
aliuk2012, line 1: We shouldn't need this check if we add a method called `render?` and add the condition there

```
def render?
  @errors.any?
end
```

app/components/error_summary_component/view.rb
```diff
@@ -0,0 +1,8 @@
+module ErrorSummaryComponent
+  class View < ViewComponent::Base
```
aliuk2012, line 2:
```suggestion
  class View < ViewComponent::Base
  
  attr_accessor :errors
```

then in the view we can use `errors` rather than `@errors` which a dev may assume has come from a controller action etc.

app/components/page_list_component/error_summary.html.erb
aliuk2012, line 1: Could we place this file and `error_summary.rb` into a subfolder called `error_summary` and then they too could be called `view.html.rb` and `view.rb` to follow along with the pattern of the other components

DavidBiddle: Good suggestion!

aliuk2012 (approved):
# :shipit:  🚀

PR #435, opened 2023-05-12
Add edit conditions policy

SamJamCul:
This adds an edit conditions policy, which is used instead of the feature flag check and add conditions policy.

### DIFF
--- a/app/components/page_list_component/view.html.erb
+++ b/app/components/page_list_component/view.html.erb
@@ -32,7 +32,7 @@
             </dd>
           </div>
 
-          <% if FeatureService.enabled?(:basic_routing) %>
+          <% if render_routing? %>
             <% page.conditions.each do |condition| %>
               <div id="<%= PageListComponent::ErrorSummary::View.error_id(condition.id) %>" class="govuk-summary-list__row app-page-list__row <%= class_names( "govuk-form-group--error": condition.has_errors?) %>">
                 <dt class="govuk-summary-list__key govuk-summary-list__key app-page-list__key">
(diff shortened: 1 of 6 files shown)


app/components/page_list_component/view.html.erb
```diff
@@ -32,7 +32,7 @@
             </dd>
           </div>
 
-          <% if FeatureService.enabled?(:basic_routing) %>
+          <% if @can_edit_page_routing %>
```
DavidBiddle, line 35: Getting in here before @aliuk2012 does - suggest using attr_accessor in the view.rb to allow you to remove the `@`:
```suggestion
          <% if can_edit_page_routing %>
```

app/components/page_list_component/view.rb
```diff
@@ -2,10 +2,11 @@ module PageListComponent
   class View < ViewComponent::Base
     delegate :question_text_with_optional_suffix, to: :helpers
```
DavidBiddle, line 3:
```suggestion
    delegate :question_text_with_optional_suffix, to: :helpers
    attr_accessor :can_edit_page_routing
```

app/components/page_list_component/view.html.erb
DavidBiddle: You could potentially do the same with `@form` and `@pages` but if you want to keep this PR focused on the policy change I'm happy to do it in the `can_add_page_routing_conditions?` update PR

app/components/page_list_component/view.rb
DavidBiddle: Your `render_routing?` method makes this unnecessary 👍

DavidBiddle (approved):
💯

PR #437, opened 2023-05-12
Update can_add_page_routing_conditions policy

DavidBiddle:
Updates the qualifying_route_pages form method so that a form will only allow new routes to be added to pages which meet all of these conditions:
- are 'select from a list' questions with only one answer allowed (i.e. radio questions)
- aren't the last page in the form
- don't already have routes attached to them.

We also update the `can_add_page_routing_conditions?` form policy so that it:
- no longer includes the super admin check (so routing will only be visible on environments with the feature flag enabled)
- checks that there are qualifying route pages available.

Note: should probably wait for #435 to be merged and rebase before merging this one. 

Trello card: https://trello.com/c/Ga9bXJ8S/765-modify-the-logic-for-when-the-add-a-route-button-is-displayed

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -16,7 +16,7 @@ class Form < ActiveResource::Base
   end
 
   def qualifying_route_pages
-    pages.filter { |p| p.answer_type == "selection" && p.answer_settings.only_one_option == "true" }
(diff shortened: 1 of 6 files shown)


SamJamCul (approved):
Very intricate tests, pretty cool stuff!

PR #442, opened 2023-05-16
Preview link no longer need to open up in a new tab

aliuk2012:
User research has found that users get confused when opening up previews in new tab and how they get back to form admin

There is another ticket for forms-runner to add a link in forms-runner banner to bring the user back to forms-admin. The idea is that this will be merged at the same time as the changes to forms-runner so it will be easier for users to "toggle" between forms-runner and forms-admin.

Trello card: https://trello.com/c/Scetz4bk/772-change-preview-link-to-open-in-current-tab-window

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


DavidBiddle (approved):
Looks good!

PR #445, opened 2023-05-17
Fix event triggers for Trello Poster action

lfdebrux:
We noticed that the Trello Poster was only working some of the time, it
turns out there was a typo in the event trigger configuration, the
action `created` is not valid, it should be `opened` [[1]].

[1]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request


aliuk2012 (approved):
Good spot!! 👏

PR #449, opened 2023-05-19
Make service unavailable page visible without signing in

lfdebrux:
Users shouldn't need to sign in to see that the service is unavailable.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,8 +4,8 @@ class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods unless Settings.basic_auth.enabled
   include Pundit::Authorization
   before_action :set_request_id
-  before_action :authenticate
   before_action :check_service_unavailable
+  before_action :authenticate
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -4,8 +4,8 @@ class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods unless Settings.basic_auth.enabled
   include Pundit::Authorization
   before_action :set_request_id
-  before_action :authenticate
   before_action :check_service_unavailable
+  before_action :authenticate
```
aliuk2012, line 8: I'm really not a huge fan of having loads of before_actions, especially if the order is important. Could we maybe create a private method in this controller to merge some of these and then have a test for the order that they called?

lfdebrux: I put the `authenticate` and `check_access` before actions together in one method. I don't think we can put any other before actions together, as we need to be able to skip different actions in different controllers.

jamie-o-wilkinson (approved):

PR #452, opened 2023-05-22
Use select list if there are more than 10 routable pages

DavidBiddle:
Switches the routing page UI to use a select list instead of a radio button group if there are more than 10 options to choose from.

Trello card: https://trello.com/c/F4Hpw6tU/675-update-new-route-page-to-use-a-select-list-when-there-are-more-than-10-options

#### Checklist

  - [n/a] README.md

### DIFF
--- a/spec/views/pages/conditions/routing_page.html.erb_spec.rb
+++ b/spec/views/pages/conditions/routing_page.html.erb_spec.rb
@@ -57,6 +57,10 @@ describe "pages/conditions/routing_page.html.erb" do
     it "has a select option for each routing pages" do
       expect(rendered).to have_css("select > option", count: pages.length)
     end
+
+    it "includes the page number and question text" do
+      expect(rendered).to have_text(pages.first.question_with_number)
+    end
   end
 
   it "has a submit button" do
--- a/spec/views/pages/conditions/routing_page.html.erb_spec.rb
+++ b/spec/views/pages/conditions/routing_page.html.erb_spec.rb
@@ -22,13 +22,41 @@ describe "pages/conditions/routing_page.html.erb" do
     expect(rendered).to have_css("p.govuk-body", text: t("routing_page.body_text"))
   end
 
-  it "contains a fieldset legend asking a user to select a question page" do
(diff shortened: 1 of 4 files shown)


spec/views/pages/conditions/routing_page.html.erb_spec.rb
```diff
@@ -22,13 +22,41 @@
...
+
+  context "with more than 10 options" do
+    let(:pages) { build_list :page, 11, :with_selections_settings, form_id: 1 }
+
+    it "contains a fieldset legend asking a user to select a question page" do
+      expect(rendered).to have_css(".govuk-label", text: t("routing_page.legend_text"))
+      expect(rendered).to have_css("div.govuk-hint", text: t("routing_page.legend_hint_text"))
+    end
+
+    it "has a radio option for each routing pages" do
```
jamie-o-wilkinson, line 57: Is this a typo in the test name? Should it be something like "has a select option for each routing page"?

DavidBiddle: Found my lazy copy-pasting 😬

DavidBiddle: Have updated

spec/views/pages/conditions/routing_page.html.erb_spec.rb
```diff
@@ -22,13 +22,41 @@
     expect(rendered).to have_css("p.govuk-body", text: t("routing_page.body_text"))
   end
 
-  it "contains a fieldset legend asking a user to select a question page" do
-    expect(rendered).to have_css(".govuk-fieldset__legend", text: t("routing_page.legend_text"))
-    expect(rendered).to have_css("div.govuk-hint", text: t("routing_page.legend_hint_text"))
+  context "with fewer than 10 options" do
```
aliuk2012, line 25: Might be worth adding a view spec when there are no options. Do you expect to see a "blank" radio or select list. Or should we be explaining to the user why theres no valid option to select from?

aliuk2012: Ah yes the policy will stop that.

PR #457, opened 2023-05-23
Use govuk-frontend global styles

DavidBiddle:

- Enables `$govuk-global-styles` in our CSS - this means that we won't have to add "govuk-body" and "govuk-link" classes any more
- Removes existing uses of govuk-body and govuk-link
- Also enables `$govuk-new-link-styles` - this gives us [improved, more accessible link styles](https://designnotes.blog.gov.uk/2021/07/07/making-links-easier-to-see-and-read-on-gov-uk/)

We already had both of these things enabled in forms-runner and forms-product-pages, but we missed forms-admin.

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/components/form_url_component/view.html.erb
+++ b/app/components/form_url_component/view.html.erb
@@ -3,7 +3,7 @@
     <%= t("form_url_component.form_url") %>
   </h2>
 
-  <p class="govuk-body" data-copy-target>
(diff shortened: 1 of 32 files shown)


aliuk2012 (approved):
Nice refactor 🎖️

PR #460, opened 2023-05-24
Simplify routing error UI

DavidBiddle:
Updates the UI for routing errors:
- the errors in the page list are no longer links
- the errors in the page list now have the same text as those in the error summary
- the errors now appear above the route details, instead of replacing bits of it
- the answer value and goto page text will now show a placeholder if the relevant information doesn't exist

Trello card: https://trello.com/c/1KqRjuJT/794-change-routing-error-message-presentation

#### Checklist

  - [n/a] README.md
  - [n/a ] Elsewhere (please link)

### DIFF
--- a/app/components/page_list_component/view.html.erb
+++ b/app/components/page_list_component/view.html.erb
@@ -40,28 +40,25 @@
                 </dt>
 
                 <dd class="govuk-summary-list__value">
-                  <p class="govuk-!-margin-2">
-                    <%= t("page_conditions.#{condition.validation_errors.length > 1 || condition.has_errors_for_field?(:answer_value) ? "condition_check_page_text_with_errors" : "condition_check_page_text"}", check_page_text: question_text_for_page(condition.check_page_id)) %>
+                  <ul class="govuk-list govuk-!-margin-0">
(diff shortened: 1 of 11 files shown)


app/components/page_list_component/view.html.erb
```diff
@@ -40,28 +40,25 @@
                 </dt>
 
                 <dd class="govuk-summary-list__value">
-                  <p class="govuk-!-margin-2">
-                    <%= t("page_conditions.#{condition.validation_errors.length > 1 || condition.has_errors_for_field?(:answer_value) ? "condition_check_page_text_with_errors" : "condition_check_page_text"}", check_page_text: question_text_for_page(condition.check_page_id)) %>
+                  <ul class="govuk-list govuk-!-margin-0">
+                    <% condition.validation_errors.each do |error| %>
```
aliuk2012, line 44: Do we want use `<ul>` if there is only 1 validation error?

DavidBiddle: I think this is OK and it's not worth complicating it - a list with one item is semantically valid, and this is what the Design System error summary component does which is probably the most similar existing component

aliuk2012 (approved):

PR #461, opened 2023-05-24
Add configuration of authentication provider

lfdebrux:
We want to be able to use other authentication strategies in forms-admin. This PR refactors the application controller to allow configuration of the strategy using the settings files or environment variable.

We also deprecate the old way of selecting the basic authentication strategy, and move the basic authentication strategy out the application controller and into a Warden strategy.

Note that for the setting, we call it `auth_provider`, not `auth_strategy`. The term strategy is used consistently in Warden, but in OmniAuth provider and strategy are used somewhat interchangeably. I think arguably you could say the provider is the external identity provider and the strategy is the code you use to communicate with it, but in the context of a configuration setting there isn't too much difference in it. To be consistent with what we save in our database (see PR https://github.com/alphagov/forms-admin/pull/469) and to better express the intention of the configuration variable, I think we should stick with provider.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ gem "puma", "~> 6.3.0"
 # Used for handling authentication
 gem "gds-sso"
 gem "omniauth"
+gem "warden"
 
 # Used for handling authorisation policies
 gem "pundit"
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,7 @@
 require "resolv"
 
 class ApplicationController < ActionController::Base
-  include GDS::SSO::ControllerMethods unless Settings.basic_auth.enabled
+  include GDS::SSO::ControllerMethods
   include Pundit::Authorization
   before_action :set_request_id
   before_action :check_service_unavailable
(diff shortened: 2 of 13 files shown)


jamie-o-wilkinson (approved):

app/controllers/application_controller.rb
```diff
@@ -20,51 +20,12 @@ class ApplicationController < ActionController::Base
...
-    @current_user = current_user
-  end
-
   def check_service_unavailable
     if Settings.service_unavailable
       render "errors/service_unavailable", status: :service_unavailable, formats: :html
     end
   end
 
-  def check_access
```
jamie-o-wilkinson, line 62: Didn't realise this was now only called from one place. Nice 👍

app/helpers/application_helper.rb
```diff
@@ -65,11 +65,18 @@ def govuk_assets_path
   end
 
   def header_component_options(user:, can_manage_users:)
+    links = Hash.new({}).update( # rubocop:disable Rails/SaveBang
```
jamie-o-wilkinson, line 68: It took me a moment to figure out how this worked (though I may just be slow after not enough sleep last night 😪 )...might be worth a chat to see if there is a clearer way?

Gemfile
```diff
@@ -14,6 +14,10 @@ gem "pg", "~> 1.5"
 # Use the Puma web server [https://github.com/puma/puma]
 gem "puma", "~> 6.2.2"
 
+# Used for handling authentication
+gem "gds-sso"
+gem "warden"
```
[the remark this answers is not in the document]
lfdebrux: Yeah, gds-sso depends on Warden.

app/helpers/application_helper.rb
lfdebrux: I would like a clearer way haha...

lfdebrux: I've refactored to make more explicit, thanks for the nudge

lfdebrux:
I've renamed things a bit, I decided that it would be better to be consistent in our configuration interface with what we save in our db, so I've changed the name of the configuration variable to `auth_provider`.

DavidBiddle (approved):
- Looks good overall 🎉 
- Makes application controller less complex 👏
- Adds integration tests 😻

jamie-o-wilkinson (approved):

PR #463, opened 2023-05-25
Allow user to select check your answers page as goto page

DavidBiddle:
Makes changes to allow users to route to the check your answers page.

- Adds 'Check your answers' as an option in the dropdown list
- Adds some logic to the condition form object to convert the id from the dropdown into the format the API expects
- Adds a translation for 'Check your answers' and displays this in the various places that currently show the goto_page question text

Trello card: https://trello.com/c/27Y35zlE/676-add-the-option-for-a-route-to-go-to-the-check-your-answers-page

#### Checklist

  - [n/a] README.md

### DIFF
--- a/spec/forms/pages/conditions_form_spec.rb
+++ b/spec/forms/pages/conditions_form_spec.rb
@@ -126,7 +126,7 @@ RSpec.describe Pages::ConditionsForm, type: :model do
   describe "#goto_page_options" do
     it "returns a list of answers for the given page" do
       result = described_class.new(form:, page: pages.first).goto_page_options
-      expect(result).to eq([OpenStruct.new(id: nil, question_text: I18n.t("helpers.label.pages_conditions_form.default_goto_page_id")), form.pages.map { |p| OpenStruct.new(id: p.id, question_text: p.question_text) }, OpenStruct.new(id: "check_your_answers", question_text: "Check your answers")].flatten)
+      expect(result).to eq([OpenStruct.new(id: nil, question_text: I18n.t("helpers.label.pages_conditions_form.default_goto_page_id")), form.pages.map { |p| OpenStruct.new(id: p.id, question_text: p.question_text) }, OpenStruct.new(id: "check_your_answers", question_text: I18n.t("page_conditions.check_your_answers"))].flatten)
     end
   end
 
--- a/spec/forms/pages/conditions_form_spec.rb
+++ b/spec/forms/pages/conditions_form_spec.rb
@@ -32,7 +32,7 @@ RSpec.describe Pages::ConditionsForm, type: :model do
 
   describe "#submit" do
     context "when validation pass" do
-      it "creates a condition" do
+      before do
         ActiveResource::HttpMock.respond_to do |mock|
           mock.post "/api/v1/forms/1/pages/2/conditions", post_headers, { id: 2 }.to_json, 200
         end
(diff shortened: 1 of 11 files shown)


spec/forms/pages/conditions_form_spec.rb
```diff
@@ -32,15 +32,24 @@
...
         conditions_form.answer_value = "Rabbit"
         conditions_form.goto_page_id = 4
+      end
+
+      it "calls assign_skip_to_end" do
+        allow(conditions_form).to receive(:assign_skip_to_end)
+        conditions_form.submit
+
+        expect(conditions_form).to have_received(:assign_skip_to_end).exactly(1).times
+      end
```
aliuk2012, line 50: I think this is testing implementation details rather than behaviour. What is the expected behaviour we expecting to happen when submit is called with specific params passed in.

spec/forms/pages/conditions_form_spec.rb
```diff
@@ -53,27 +62,37 @@
...
+
+      it "calls assign_skip_to_end" do
+        allow(conditions_form).to receive(:assign_skip_to_end)
 
-        expect(conditions_form.update).to be_truthy
+        conditions_form.update_condition
+
+        expect(conditions_form).to have_received(:assign_skip_to_end).exactly(1).times
+      end
```
aliuk2012, line 85: I think this is testing implementation details rather than behaviour. What is the expected behaviour we expecting to happen when submit is called with specific params passed in.

spec/forms/pages/conditions_form_spec.rb
DavidBiddle: Do I need to test that? I'm testing the method itself in the `describe "#assign_skip_to_end"` block later in the file, so if I test it again in this test I feel like I'll just be repeating those cases

spec/forms/pages/delete_condition_form_spec.rb
```diff
@@ -26,7 +26,7 @@
 
   describe "#delete" do
     context "when validation pass" do
-      let(:condition) { Condition.new id: 2, form_id: 1, page_id: 2, routing_page_id: 1, check_page_id: 1, answer_value: "Wales", goto_page_id: 3 }
+      let(:condition) { Condition.new id: 2, form_id: 1, page_id: 2, routing_page_id: 1, check_page_id: 1, answer_value: "Wales", goto_page_id: 3, skip_to_end: false }
```
aliuk2012, line 29: should we maybe use conditions factory?

spec/forms/pages/delete_condition_form_spec.rb
```diff
@@ -49,7 +49,7 @@
 
   describe "#goto_page_question_text" do
     context "when there is a goto_page_id" do
-      let(:condition) { Condition.new id: 2, form_id: 1, page_id: 1, routing_page_id: page.id, check_page_id: page.id, answer_value: "Wales", goto_page_id: goto_page.id }
+      let(:condition) { Condition.new id: 2, form_id: 1, page_id: 1, routing_page_id: page.id, check_page_id: page.id, answer_value: "Wales", goto_page_id: goto_page.id, skip_to_end: false }
```
aliuk2012, line 52: maybe use conditions factory?

spec/forms/pages/conditions_form_spec.rb
aliuk2012: ah sorry, I didn't spot that when I was reviewing this code. 

Ok, I see you have united tested the `assign_skip_to_end`.

spec/forms/pages/conditions_form_spec.rb
aliuk2012: Ok, I see you have united tested the `assign_skip_to_end`.

spec/forms/pages/delete_condition_form_spec.rb
DavidBiddle: Hmm, when I use the conditions factory I get:
```
Failure/Error: result = record.destroy
ActiveResource::MissingPrefixParam:
form_id prefix_option is missing
```

aliuk2012 (approved):

aliuk2012 (approved):

PR #467, opened 2023-05-31
Prevent marking complete if routing has errors

DavidBiddle:
Adds a custom validation to the Mark Complete form to prevent a user marking the pages section complete if the form has any routing errors.

**Note**: needs to be merged and deployed _after_ the corresponding API PR (https://github.com/alphagov/forms-api/pull/282) - this relies on some methods that are introduced in that PR.

Trello card: https://trello.com/c/W5mLjhd3/804-prevent-users-marking-the-pages-section-as-complete-while-there-are-routing-errors

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/components/page_list_component/view.html.erb
+++ b/app/components/page_list_component/view.html.erb
@@ -34,7 +34,7 @@
 
           <% if render_routing? %>
             <% page.conditions.each do |condition| %>
-              <div id="<%= PageListComponent::ErrorSummary::View.error_id(condition.id) %>" class="govuk-summary-list__row app-page-list__row <%= class_names( "govuk-form-group--error": condition.has_errors?) %>">
+              <div id="<%= PageListComponent::ErrorSummary::View.error_id(condition.id) %>" class="govuk-summary-list__row app-page-list__row <%= class_names( "govuk-form-group--error": condition.has_routing_errors?) %>">
                 <dt class="govuk-summary-list__key govuk-summary-list__key app-page-list__key">
                   <%= t("page_conditions.condition_name", page_index: page.position) %>
                 </dt>
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -31,7 +31,6 @@ class FormsController < ApplicationController
     if @mark_complete_form.mark_section
       redirect_to form_path(@form)
     else
-      flash[:message] = "Save unsuccessful"
       render "pages/index", status: :unprocessable_entity
     end
   end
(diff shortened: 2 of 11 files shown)


app/components/page_list_component/view.html.erb
```diff
@@ -34,7 +34,7 @@
 
           <% if render_routing? %>
             <% page.conditions.each do |condition| %>
-              <div id="<%= PageListComponent::ErrorSummary::View.error_id(condition.id) %>" class="govuk-summary-list__row app-page-list__row <%= class_names( "govuk-form-group--error": condition.has_errors?) %>">
+              <div id="<%= PageListComponent::ErrorSummary::View.error_id(condition.id) %>" class="govuk-summary-list__row app-page-list__row <%= class_names( "govuk-form-group--error": condition.has_routing_errors) %>">
```
aliuk2012, line 37:
```suggestion
              <div id="<%= PageListComponent::ErrorSummary::View.error_id(condition.id) %>" class="govuk-summary-list__row app-page-list__row <%= class_names( "govuk-form-group--error": condition.has_routing_errors?) %>">
```

Rails is clever enough when you add `?` to the method name to expect the value to be true/false if that is what is expected from this attribute.

app/controllers/forms_controller.rb
```diff
@@ -31,7 +31,6 @@ def mark_pages_section_completed
     if @mark_complete_form.mark_section
       redirect_to form_path(@form)
     else
-      flash[:message] = "Save unsuccessful"
       render "pages/index", status: :unprocessable_entity
```
aliuk2012, line 34: I wonder if its worth "resetting" the radio button. Currently if there are validation messages it will still show the "Yes" selected for completed when it hasn't actually been saved.

app/forms/forms/mark_complete_form.rb
```diff
@@ -21,4 +22,12 @@ def assign_form_values
     self.mark_complete = form.try(:question_section_completed)
     self
   end
+
+  def marked_complete?
+    mark_complete == "true"
+  end
+
+  def has_validation_errors
```
aliuk2012, line 30: should this maybe be called `has_routing_errors`? I'm only asking because it feels like we are introduction a whole new concept here like whats the difference between validation errors and routing errors? Dont have strong opinions on this only a suggestion.

app/controllers/forms_controller.rb
DavidBiddle: Think this is a good idea, have implemented

app/components/page_list_component/view.html.erb
DavidBiddle: Nice!

aliuk2012 (approved):

PR #469, opened 2023-06-01
Add auth provider to User model

lfdebrux:

Part of https://trello.com/c/KoHbF6Xc.

To make switching from one OAuth based identity provider to another a bit less risky, we add the name of provider to the users table so the `uid` can be disambiguated.

I'd suggest doing it this way (rather than just allowing the uids to be incosistent, or getting rid of uid altogether) because a) it's compatible with what Devise does [[1]] and b) because then we'll to be able to identify where users in our database have come from, which seems like it might be useful.

We can then, for audit purposes, log when a user signs in with a different provider, and record the changes that are made to the user record.

[1]: https://github.com/heartcombo/devise/wiki/OmniAuth:-Overview

### DIFF
--- a/spec/integration/gds_sso_spec.rb
+++ b/spec/integration/gds_sso_spec.rb
@@ -50,4 +50,32 @@ RSpec.describe "usage of gds-sso gem" do
       expect(request.env["warden"].authenticated?).to be true
     end
   end
+
+  describe User do
+    describe ".find_for_gds_oauth" do
+      it "is called by the gds_sso Warden strategy" do
+        allow(described_class).to receive(:find_for_gds_oauth).and_call_original
+        allow(described_class).to receive(:find_for_auth).and_call_original
+
+        gds_sso = Warden::Strategies[:gds_sso].new({
+          "omniauth.auth" => omniauth_hash,
+        })
+        gds_sso.authenticate!
+
+        expect(described_class).to have_received(:find_for_gds_oauth)
+
+        expect(described_class).to have_received(:find_for_auth).with(
+          uid: "123456",
+          email: "test@example.com",
+          name: "Test User",
+          permissions: ["---"],
+          organisation_slug: "test-org",
+          organisation_content_id: "00000000-0000-0000-0000-000000000000",
+          disabled: false,
+        )
+
+        expect(gds_sso.successful?).to be true
+      end
+    end
+  end
 end
--- a/spec/integration/gds_sso_spec.rb
+++ b/spec/integration/gds_sso_spec.rb
@@ -1,0 +1,53 @@
+require "rails_helper"
+
+RSpec.describe "usage of gds-sso gem" do
+  let(:omniauth_hash) do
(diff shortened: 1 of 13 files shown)


spec/integration/gds_sso_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      expect(response).to redirect_to("/auth/gds")
+    end
+
+    it "authenticates with OmniAuth and Warden" do
+      OmniAuth.config.mock_auth[:gds] = omniauth_hash
+
+      get "/auth/gds"
+
+      expect(response).to redirect_to("/auth/gds/callback")
```
aliuk2012, line 46: using rails path helpers `expect(response).to redirect_to(gds_sign_in_path)

spec/integration/gds_sso_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      get "/"
+
+      expect(response).to redirect_to("/auth/gds")
+    end
+
+    it "authenticates with OmniAuth and Warden" do
+      OmniAuth.config.mock_auth[:gds] = omniauth_hash
+
+      get "/auth/gds"
```
aliuk2012, line 44: Is this path correct? I dont see it mentioned when I run `rails routes`

lfdebrux: Yes, it's part of OmniAuth [[1](https://github.com/omniauth/omniauth#integrating-omniauth-into-your-application)]:

> To use OmniAuth, you need only to redirect users to `/auth/:provider`, where `:provider` is the name of the strategy (for example, `developer` or `twitter`). From there, OmniAuth will take over and take the user through the necessary steps to authenticate them with the chosen strategy.

Because it's handled by the OmniAuth middleware and not by the Rails router these routes don't show up in `rails routes` however.

spec/integration/gds_sso_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      expect(response).to redirect_to("/auth/gds/callback")
+
+      get "/auth/gds/callback"
```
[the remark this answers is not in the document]
lfdebrux: Done

spec/integration/gds_sso_spec.rb
lfdebrux: Done

spec/integration/gds_sso_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    after do
+      OmniAuth.config.test_mode = false
+    end
+
+    it "redirects to OmniAuth when no user is logged in" do
+      logout
+
+      get "/"
```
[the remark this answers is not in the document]
lfdebrux: Done

DavidBiddle (approved):
Looks good 🎉

jamie-o-wilkinson (approved):
Looks good, nice work! 👏

aliuk2012 (approved):

PR #474, opened 2023-06-06
Implement trial role and initial restrictions

jamie-o-wilkinson:
We want to be able to give new users a ‘trial’ role so they have permissions to do a limited number of things while deciding whether GOV.UK Forms is a good fit for their team.

This PR:
- adds the trial role to the User model
- allows super-admins to assign the trial role to users
- sets the creator_id on new form creation
- makes organisation optional for trial users
- allows trial users to create, view, and edit only forms they have created

Trello card: https://trello.com/c/Bfl4ckBp

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,6 +10,7 @@ class User < ApplicationRecord
   enum :role, {
     super_admin: "super_admin",
     editor: "editor",
+    trial: "trial",
   }
 
   validates :role, presence: true
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -50,4 +50,8 @@ class User < ApplicationRecord
       create!(attributes)
     end
   end
+
+  def organisation_valid?
+    trial? || organisation.present?
+  end
 end
--- a/app/policies/form_policy.rb
+++ b/app/policies/form_policy.rb
@@ -14,8 +14,12 @@ class FormPolicy
     end
 
     def resolve
-      scope
-        .where(org: user.organisation.slug)
+      if user.trial?
+        scope.where(creator_id: user.id)
+      else
+        scope
+          .where(org: user.organisation.slug)
+      end
     end
   end
 
@@ -44,6 +48,10 @@ class FormPolicy
   alias_method :can_delete_page_routing_conditions?, :can_edit_page_routing_conditions?
 
 private
+
+  def user_is_form_creator
+    form.respond_to?(:creator_id) ? user.id == form.creator_id : false
+  end
 
   def users_organisation_owns_form
     user.organisation.slug == form.org
--- a/app/policies/form_policy.rb
+++ b/app/policies/form_policy.rb
@@ -7,7 +7,7 @@ class FormPolicy
     attr_reader :user, :form, :scope
 
     def initialize(user, scope)
-      raise UserMissingOrganisationError, "Missing required attribute organisation_id" if user.organisation.blank?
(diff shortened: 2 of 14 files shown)


app/models/user.rb
```diff
@@ -50,4 +51,8 @@ def self.find_for_auth(attributes)
       create!(attributes)
     end
   end
+
+  def organisation_valid?
```
jamie-o-wilkinson, line 55: I don't love the name of this...suggestions welcome

lfdebrux: Maybe if we move this method into the policy object and call it `user_has_valid_organisation?` then it might feel less awkward? I'm assuming we're not going to need it for anything except the `UserMissingOrganisationError` on the policy.

app/policies/form_policy.rb
```diff
@@ -45,7 +49,11 @@ def can_edit_page_routing_conditions?
 
 private
 
+  def user_is_form_creator
+    form.respond_to?(:creator_id) ? user.id == form.creator_id : false
+  end
+
   def users_organisation_owns_form
-    user.organisation.slug == form.org
+    user.organisation&.slug == form.org
```
lfdebrux, line 57: This will always be true if the form does not have an org and and the user does not have an org; which I think means in theory that `can_view_form?` will always be true for any combination of trial user and form created by trial user; which I think means any trial user will be able to view and edit any trial form...

spec/policies/form_policy_spec.rb
```diff
@@ -32,10 +32,41 @@
...
+
+          it { is_expected.to forbid_actions(%i[can_view_form]) }
+        end
+      end
+    end
   end
 
   describe "#can_add_page_routing_conditions?" do
-    let(:form) { build :form, pages:, org: "gds" }
+    let(:form) { build :form, pages:, org: "gds", creator_id: 123 }
```
lfdebrux, line 79: Is adding a `creator_id` to this test form needed for some reason? I'm surprised that the `#can_add_page_routing_conditions?` check is affected by the creator ID.

app/models/user.rb
jamie-o-wilkinson: Hmm...We need it in both the policy and scope. I think I moved it into the User model as it only needs data about the user, so I pushed the function down as far as possible.

spec/policies/form_policy_spec.rb
```diff
@@ -32,10 +42,41 @@
         end
       end
     end
+
+    context "with a trial role" do
```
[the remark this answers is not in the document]
lfdebrux: Sorry, to be clear, I meant we should have

```
let(:form) { build :form, creator_id: 1234, org: nil }
```

spec/policies/form_policy_spec.rb
```diff
@@ -105,11 +138,51 @@
       expect(scope).to have_received(:where).with(org: organisation_slug)
     end
 
+    context "with a trial user role" do
+      let(:user) { build :user, :with_no_org, :with_trial, id: 123 }
+      let(:form) { build(:form, creator_id: 123) }
```
lfdebrux, line 143:
```suggestion
      let(:form) { build(:form, creator_id: 123, org: nil) }
```

lfdebrux (approved):
Nice stuff! 🙌

PR #475, opened 2023-06-07
Let task list component render section with body instead of rows

lfdebrux:

For the trial user we want some of the form creation/editing task list sections to have text instead of tasks. See for example tickets https://trello.com/c/I44RySs9 and https://trello.com/c/zz3V7GLv.

This PR adds the optional `body_text` attribute to the task list section struct for this purpose.

If you have this branch checked out locally you can preview this feature at http://localhost:3000/preview/task_list_component/task_list_component/default. You may need to run `bin/vite build` first to rebuild the app stylesheets with the new CSS classes.

### Screenshot

Screenshot of http://localhost:3000/preview/task_list_component/task_list_component/default

### DIFF
--- a/spec/components/task_list_component/task_list_component_preview.rb
+++ b/spec/components/task_list_component/task_list_component_preview.rb
@@ -8,7 +8,16 @@ class TaskListComponent::TaskListComponentPreview < ViewComponent::Preview
           { task_name: "Edit the email address", path: "#", status: :cannot_start },
           { task_name: "Confirm the submission email address", path: "#", status: :cannot_start, active: false },
         ] },
-      { title: "do something else with a form",
+      { title: "Do something else with a form",
+        rows: [
+          { task_name: "Edit the name of your form", path: "#", status: :completed, active: true },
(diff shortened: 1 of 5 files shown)


spec/components/task_list_component/task_list_component_preview.rb
```diff
@@ -8,7 +8,16 @@ def default
           { task_name: "Edit the email address", path: "#", status: :cannot_start },
           { task_name: "Confirm the submission email address", path: "#", status: :cannot_start, active: false },
         ] },
-      { title: "do something else with a form",
+      { title: "Do something else with a form",
```
jamie-o-wilkinson, line 11: "Do something else with a form" duplicates "Do yet another thing with a form" - was that deliberate?

lfdebrux: I wanted to see that the spacing between each section was consistent whether using body_text or rows, and just copied-pasted the existing example a bit more

jamie-o-wilkinson (approved):
Looks nice! 🎉

PR #476, opened 2023-06-08
Fix spacing under status tag form task list page

lfdebrux:

When there are no questions there is unnecessary extra space under the draft tag. This happens because we're including a paragraph with lots of margin for the preview link, but there's no preview link (because there are no pages to preview).

This commit changes the template to only render the container for the preview link if preview link will be rendered.

#### Before

Screenshot of task list page heading area before change

#### After

Screenshot of task list page heading area after change

### DIFF
--- a/app/views/forms/show.html.erb
+++ b/app/views/forms/show.html.erb
@@ -15,9 +15,12 @@
       <p><%= flash[:message] %></p>
     <% end %>
 
-    <p class="govuk-!-margin-bottom-9">
-      <%= render PreviewLinkComponent::View.new(@form.pages, link_to_runner(Settings.forms_runner.url, @form.id, @form.form_slug)) %>
-    </p>
+    <% preview_link = PreviewLinkComponent::View.new(@form.pages, link_to_runner(Settings.forms_runner.url, @form.id, @form.form_slug)) %>
+    <% if preview_link.render? %>
+      <p class="govuk-!-margin-bottom-9">
+        <%= render preview_link %>
+      </p>
+    <% end %>
 
     <% if @form.has_live_version %>
       <div class="govuk-inset-text">


app/views/forms/show.html.erb
```diff
@@ -15,9 +15,12 @@
...
 
-    <p class="govuk-!-margin-bottom-9">
-      <%= render PreviewLinkComponent::View.new(@form.pages, link_to_runner(Settings.forms_runner.url, @form.id, @form.form_slug)) %>
-    </p>
+    <% preview_link = PreviewLinkComponent::View.new(@form.pages, link_to_runner(Settings.forms_runner.url, @form.id, @form.form_slug)) %>
+    <% if preview_link.render? %>
+      <p class="govuk-!-margin-bottom-9">
+        <%= render preview_link %>
+      </p>
+    <% end %>
```
aliuk2012, line 23: Could we not wrap this up into the component itself? 

Ideally we want to call `<%= render PreviewLinkComponent::View.new...`  and let the component decide how its rendered or if its rendered

lfdebrux: That would be cool... is there a way to have it be an optional part of the component? We use it differently in different places:

```
$ git grep -C 1 'render PreviewLinkComponent'
app/views/forms/show.html.erb-    <p class="govuk-!-margin-bottom-9">
app/views/forms/show.html.erb:      <%= render PreviewLinkComponent::View.new(@form.pages, link_to_runner(Settings.forms_runner.url, @form.id, @form.form_slug)) %>
app/views/forms/show.html.erb-    </p>
--
app/views/live/show_form.html.erb-    <p>
app/views/live/show_form.html.erb:      <%= render PreviewLinkComponent::View.new(form.pages, link_to_runner(Settings.forms_runner.url, form.id, form.form_slug, mode: :preview_live)) %>
app/views/live/show_form.html.erb-    </p>
--
app/views/pages/index.html.erb-      <% end %>
app/views/pages/index.html.erb:      <%= render PreviewLinkComponent::View.new(@pages, link_to_runner(Settings.forms_runner.url, @form.id, @form.form_slug)) %>
app/views/pages/index.html.erb-    </div>

```

aliuk2012: 🤔  just off the top of my head, could pass in an optional argument (default to include `<p>`) and then either include it or not in `view.rb`, Interesting to see there's a difference between `show.html.erb` and `show_live.html.erb`. Any chance we could standardise it?

lfdebrux: I think given the three places where we use this component we have it in different containers, it's more pain to move it in than not... are you happy with leaving it as is @aliuk2012?

DavidBiddle (approved):
Looks good to me 🎉

app/views/forms/show.html.erb
aliuk2012: Sorry haven't had much time to spike something out to show you but happy to leave it as is and we can iterate at another time. t

aliuk2012 (approved):

PR #480, opened 2023-06-08
Tweak content for routing dropdowns

DavidBiddle:
May be easier to read by going through each commit - this includes:
- Adding default value to routing page dropdown (previously it was just defaulting to the first item in the list)
- Changing 'Check your answers' to 'Check your answers before submitting' in the goto_page select box (to match the h1 on the CYA page)
- Changing 'routes' to 'route' on the edit route page (see https://gds.slack.com/archives/C048R8RJJBH/p1686129714380289?thread_ts=1683711412.643269&cid=C048R8RJJBH)
- Moving the default text for the goto_page and answer_value dropdowns into the prompt option (I only realised you could do this when adding the default value to the routing page dropdown)

Trello card: https://trello.com/c/WyFxtJJg/792-implement-drop-down-menu-content-for-simple-routing (plus some other small bits) which don't have cards

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/forms/pages/conditions_form.rb
+++ b/app/forms/pages/conditions_form.rb
@@ -36,12 +36,14 @@ class Pages::ConditionsForm
     options = page.answer_settings.selection_options.map { |option| OpenStruct.new(value: option.name, label: option.name) }
     options << OpenStruct.new(value: I18n.t("page_options_service.selection_type.none_of_the_above"), label: I18n.t("page_options_service.selection_type.none_of_the_above")) if page.is_optional
 
-    [OpenStruct.new(value: nil, label: I18n.t("helpers.label.pages_conditions_form.default_answer_value")), options].flatten
+    options
   end
 
   def goto_page_options
     page_options = pages_after_current_page(form.pages, page).map { |p| OpenStruct.new(id: p.id, question_text: p.question_with_number) }
-    [OpenStruct.new(id: nil, question_text: I18n.t("helpers.label.pages_conditions_form.default_goto_page_id")), page_options, OpenStruct.new(id: "check_your_answers", question_text: I18n.t("page_conditions.check_your_answers"))].flatten
+    page_options << OpenStruct.new(id: "check_your_answers", question_text: I18n.t("page_conditions.check_your_answers"))
+
+    page_options
   end
 
   def check_errors_from_api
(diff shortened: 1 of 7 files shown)


app/forms/pages/conditions_form.rb
```diff
@@ -36,12 +36,12 @@ def routing_answer_options
...
     options << OpenStruct.new(value: I18n.t("page_options_service.selection_type.none_of_the_above"), label: I18n.t("page_options_service.selection_type.none_of_the_above")) if page.is_optional
 
-    [OpenStruct.new(value: nil, label: I18n.t("helpers.label.pages_conditions_form.default_answer_value")), options].flatten
+    options
   end
 
   def goto_page_options
     page_options = pages_after_current_page(form.pages, page).map { |p| OpenStruct.new(id: p.id, question_text: p.question_with_number) }
-    [OpenStruct.new(id: nil, question_text: I18n.t("helpers.label.pages_conditions_form.default_goto_page_id")), page_options, OpenStruct.new(id: "check_your_answers", question_text: I18n.t("page_conditions.check_your_answers"))].flatten
+    [page_options, OpenStruct.new(id: "check_your_answers", question_text: I18n.t("page_conditions.check_your_answers"))].flatten
```
aliuk2012, line 44: 🤔  is `flatten` still needed?

aliuk2012 (dismissed):
only a minor question but otherwise some good refactors

app/forms/pages/conditions_form.rb
DavidBiddle: Good question! We still need it for adding the check your answers item, but I think we could accomplish the same with either the splat operator:
```suggestion
    [*page_options, OpenStruct.new(id: "check_your_answers", question_text: I18n.t("page_conditions.check_your_answers"))]
```
or by manually adding the item to the array before returning it:
```suggestion
    page_options << OpenStruct.new(id: "check_your_answers", question_text: I18n.t("page_conditions.check_your_answers"))
    
    page_options
```

On reflection I think I prefer both of those to the .flatten approach - have you got any strong opinions on which of those is better?

DavidBiddle: Going to go with the second one for consistency with how we do it in `routing_answer_options` above

aliuk2012 (approved):

PR #482, opened 2023-06-09
Change uses of `@current_user` to `current_user` in controllers

lfdebrux:
`current_user` should be available in all cases in controllers, whereas the instance variable `@current_user` is set only when `authenticate_and_check_access` has run. Prefer to use `current_user`.


aliuk2012 (approved):
❤️ nice, so good to go back to using standard current_user method

PR #483, opened 2023-06-09
Stop trial users from being able to change a form submission email address

lfdebrux:

Trello card: https://trello.com/c/I44RySs9

For users with trial accounts, we want to restrict them so that they can only use their own email address for draft form submissions. This PR sets the email address to the user's email address if the user has a trial account, updates the task list following the designs in the ticket, and adds a new permission check to the form policy, restricting who can access the submission email pages.

### Screenshots

Screenshot of form task list page for trial user


jamie-o-wilkinson (approved):
Looks good to me 🚀

PR #484, opened 2023-06-09
Require an organisation to be set when changing a user to editor role

lfdebrux:

Trello card: https://trello.com/c/myv5aZuf

We want to make sure all editor users belong to an organisation. To make it easier for admins to remember, when changing a user's role from trial to editor, if an organisation has not be chosen, show an error.


jamie-o-wilkinson (approved):
This was quick, nice one! 🥇

PR #485, opened 2023-06-09
Set correct none_of_the_above value in routes

DavidBiddle:
Sets the routing answer_value for None of the above to "none_of_the_above", which is what the runner uses, rather than "None of the above". Without this, routes based on None of the Above won't skip correctly.

Trello card: https://trello.com/c/uqnosctc/826-choosing-none-of-the-above-answered-as-should-not-display-a-validation-error

#### Checklist

  - [n/a] README.md


lfdebrux (approved):
Makes sense to me

PR #486, opened 2023-06-09
Fix typo in submisison email controller

lfdebrux:

Correct name of permission check from `:change_form_submission_email?` to `:can_change_form_submission_email?`. This mistake was introduced in PR #483, partly because there were no tests for the controller, so this commit also adds tests.

Trello card: https://trello.com/c/I44RySs9


DavidBiddle (approved):
Nice tests 🎉

PR #489, opened 2023-06-13
Fix ping service unavailable

aliuk2012:

Trello card:

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


spec/requests/heartbeat_controller_spec.rb
```diff
@@ -7,5 +7,16 @@
 
       expect(response.body).to eq "PONG"
     end
+
+    context "when service is in maintenance mode" do
+      before do
+        allow(Settings).to receive(:service_unavailable).and_return(true)
```
SamJamCul, line 13:
```suggestion
        allow(Settings).to receive(:maintenance_mode).and_return(true)
```

app/controllers/application_controller.rb
```diff
@@ -19,9 +19,9 @@ class ApplicationController < ActionController::Base
     render "errors/forbidden", status: :forbidden, formats: :html
   end
 
-  def check_service_unavailable
-    if Settings.service_unavailable
-      render "errors/service_unavailable", status: :service_unavailable, formats: :html
+  def check_maintenance_mode_is_enabled
+    if Settings.maintenance_mode_enabled
+      redirect_to maintenance_page_path
```
DavidBiddle, line 24: Do we still need to get rid of the `:service unavailable` status if we're exempting the ping action from running `check_maintenance_mode_is_enabled`?

spec/requests/heartbeat_controller_spec.rb
```diff
@@ -7,5 +7,16 @@
...
     end
+
+    context "when service is in maintenance mode" do
+      before do
+        allow(Settings).to receive(:maintenance_mode_enabled).and_return(true)
+      end
+
+      it "returns PONG" do
+        get "/ping"
+        expect(response.body).to eq "PONG"
```
DavidBiddle, line 18: Might be worth checking the status here too, since I think that's one of the things the AWS check is looking for

app/controllers/application_controller.rb
aliuk2012: We happy to keep it as :ok/200 status code because we are aware of why users are getting the page. its also redirected to `/maintenance` so we can track number of times users are hitting this page.

spec/requests/heartbeat_controller_spec.rb
aliuk2012: I've added a check on the status too.

DavidBiddle (approved):
Looks good! 🎉

PR #494, opened 2023-06-14
Add question text form to selection settings journey

DavidBiddle:
Adds a new question text form to the selection question journey so the user adds the question before they add their selection options.

The new form only appears when the user is adding a new page - existing pages will already have `question_text`, and the user can edit this on the edit question page, so the form is unnecessary. 

Trello card: https://trello.com/c/RBGKvaBI/838-add-question-first-for-select-from-options-answer-type

#### Checklist

  - [n/a] README.md

### DIFF
--- a/app/forms/forms/question_text_form.rb
+++ b/app/forms/forms/question_text_form.rb
@@ -1,0 +1,15 @@
+class Forms::QuestionTextForm
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  attr_accessor :question_text, :form, :page
+
+  validates :question_text, presence: true
+
+  def submit(session)
+    return false if invalid?
+
+    session[:page] = {} if session[:page].blank?
+    session[:page][:question_text] = question_text
+  end
+end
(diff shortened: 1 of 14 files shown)


jamie-o-wilkinson (approved):
One optional suggestion...otherwise looks good to me 🙌

config/locales/forms/question_text.yml
```diff
@@ -0,0 +1,12 @@
+en:
+  helpers:
+    label:
+      forms_question_text_form:
+        question_text: What's your question?
```
lfdebrux, line 5: I think this should be a curly apostrophe?

```suggestion
        question_text: What’s your question?
```

app/forms/forms/question_text_form.rb
```diff
@@ -0,0 +1,15 @@
+class Forms::QuestionTextForm
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  attr_accessor :question_text, :form, :page
```
lfdebrux, line 5: Possibly stupid question... where do `form` and `page` get used?

lfdebrux (approved):
I have a couple of questions about things I don't understand 😅 

But having tested it locally, it all seems to work!

lfdebrux (approved):
Looks good to me!

PR #499, opened 2023-06-15
Let specific IP addresses bypass the maintenance page

aliuk2012:

We need some way of bypassing the maintenance page if the users IP is in a specific list of IPs. 
These changes allow us to let the user bypass the maintenance page if their IP matches any single IP
address or is included within a specific CIDR notation range of IP

Trello card: https://trello.com/c/P9CEz21c/824-test-and-update-holding-page

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,7 +20,7 @@ class ApplicationController < ActionController::Base
   end
 
   def check_maintenance_mode_is_enabled
-    if Settings.maintenance_mode.enabled
+    if Settings.maintenance_mode.enabled && non_maintenance_bypass_ip_address?
       redirect_to maintenance_page_path
     end
   end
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -104,4 +104,11 @@ def authenticate_and_check_access
       render "errors/access_denied", status: :forbidden, formats: :html
     end
   end
+
+  def non_maintenance_bypass_ip_address?
+    return true if Settings.maintenance_mode.bypass_ips.blank?
+
+    bypass_ip_list = Settings.maintenance_mode.bypass_ips.split(",").map { |ip| IPAddr.new ip.strip }
```
DavidBiddle, line 111: You could maybe move `Settings.maintenance_mode.bypass_ips` into a variable
```suggestion
    bypass_ips = Settings.maintenance_mode.bypass_ips
    return true if bypass_ips.blank?

    bypass_ip_list = bypass_ips.split(",").map { |ip| IPAddr.new ip.strip }
```

DavidBiddle (approved):
Looks good overall, one small optional suggestion.

I don't think you'll be able to merge this while the Docker build is failing, so you'll have to rebase onto main after #506 is merged.

app/controllers/application_controller.rb
SamJamCul: I've made this change

DavidBiddle (approved):

PR #507, opened 2023-06-21
Pin node to minor version 18.16 in Dockerfile

DavidBiddle:
Fixes an issue with x86 and ARM architectures supporting different versions of the [Alpine nodejs package](https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.17&repo=&arch=&maintainer=).

x86 architectures support 18.16.1 (so our Github action passed) but our build uses ARM, which currently only supports 18.16.0. Pinning to the minor version supports both, and should make build-breaking updates slightly less frequent.

#### Checklist

  - [n/a] README.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ WORKDIR /app
 
 RUN apk update
 RUN apk upgrade --available
-RUN apk add libc6-compat openssl-dev build-base libpq-dev nodejs=18.16.1-r0 npm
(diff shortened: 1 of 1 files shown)


danworth (approved):
Passes on CI and I can see its been successfully deployed to AWS Dev. This only modifies the AWS deployment so I presume the previous Node version bump worked ok on PaaS. 🏅

PR #513, opened 2023-06-26
Add support for CDDO SSO

lfdebrux:

Trello card: https://trello.com/c/75nMIWFC

We want to be able to test https://sso.service.security.gov.uk as a authentication provider for users.

This PR adds support using the OmniAuth strategy [omniauth_openid_connect](https://rubygems.org/gems/omniauth_openid_connect), which integrates OmniAuth with the [openid_connect](https://rubygems.org/gems/openid_connect) gem. We then integrate this strategy with our own Warden/OmniAuth framework.

### Gems added

This PR adds one new gem, [omniauth_openid_connect](https://rubygems.org/gems/omniauth_openid_connect). To review the gem I looked at its stats on [ruby-toolbox.com](https://ruby-toolbox.com), and also checked security issues using GitHub Security Advisories and the [snyk.io vulnerability database](https://security.snyk.io/).

I also reviewed [openid_connect](https://rubygems.org/gems/openid_connect) which `omniauth_openid_connect` heavily relies on.

#### omniauth_openid_connect

- [x] Does it solve a specific problem we have?
        Yes, it lets us use sso.service.security.gov.uk, and it integrates well with our auth framework in the admin app
- [x] Is it updated regularly?
        Yes, it's had a couple of releases this year to date
- [x] Is it widely used?
        Yes, it has a large number of downloads, including a large number of downloads in the last month
- [x] Is it well documented?
        The documentation is reasonably good
- [x] Does it have any major security issues?
        None that I could see
- [x] Would writing the code ourselves have a better outcome?
        I don't think so, while we could try using the omniauth-oauth2 library we're already using (as OIDC is based on OAuth2), I think that would be tricky and could make our codebase less secure

#### openid_connect

- [x] Does it solve a specific problem we have?
        Yes, it lets us use sso.service.security.gov.uk, which communicates with consumers using the OpenID Connect protocol
- [x] Is it updated regularly?
        Reasonably so, no releases this year to date, but there aren't any super important changes sitting waiting to be released, and there have been one or two releases each year for over ten years now
- [x] Is it widely used?
        Yes, it has a large number of downloads, and a good number of gems
- [x] Is it well documented?
        There isn't much in the way of written documentation, but there is a good number of working examples from the library maintainer, and because we're relying on `omniauth_openid_connect` as a wrapper we don't need to read this gems documentation much
- [x] Does it have any major security issues?
        None that I could see
- [x] Would writing the code ourselves have a better outcome?
        I don't think so; see answer for `omniauth_openid_connect` above

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ gem "puma", "~> 6.3.0"
 gem "gds-sso"
 gem "omniauth"
 gem "omniauth-auth0"
+gem "omniauth_openid_connect"
 gem "warden"
 
 # Used for handling authorisation policies
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -76,7 +76,9 @@ GEM
       tzinfo (~> 2.0)
     addressable (2.8.4)
       public_suffix (>= 2.0.2, < 6.0)
+    aes_key_wrap (1.1.0)
     ast (2.4.2)
+    attr_required (1.0.1)
     better_html (2.0.1)
       actionview (>= 6.0)
       activesupport (>= 6.0)
@@ -84,6 +86,7 @@ GEM
       erubi (~> 1.4)
       parser (>= 2.4)
       smart_properties
+    bindata (2.4.15)
     bindex (0.8.1)
     bootsnap (1.16.0)
       msgpack (~> 1.2)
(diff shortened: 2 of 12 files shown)


lfdebrux:
As part of this PR I need to write up why I think it's okay to add these two new gems; I'm working on that at the moment.

thomasiles (approved):
This looks like some good problem solving, nice one 🏅 

The auth implementation would be a good topic to playback to the whole development team if you could be tempted :)

jamie-o-wilkinson (approved):
Looks good to me. I like the new gem evaluation checklist 👏

PR #516, opened 2023-06-28
Refactor add route UI logic

aliuk2012:

Users were confused as to why sometimes they see an option to add a new route or if they hadn't seen it before they wouldn't know that the feature existed

- Display the "Add new route" button all the time with no restrictions
- 2 new sets of copy to help explain to the user why they cannot progress.
   1. [First time user] They have tried to create a new route when they dont have enough questions or questions that we can route from 
   2. [Returning user] They do have the minimum number of questions and one or more select-one-from-a-list answer type but they already have route setup for those questions.

Trello card: https://trello.com/c/x2rDK7ek/867-routing-consider-making-the-add-a-question-route-button-visible-at-all-times-development

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -487,7 +487,7 @@ en:
     no_remaining_routes_heading: You have no more questions to start a route from
     routing_requirements_not_met_heading: What you need to create a route
     routing_requirements_not_met_html: |
-      <p>Before you can add a route you'll need:</p>
+      <p>Before you can add a route you’ll need:</p>
       <ul class="govuk-list govuk-list--bullet">
         <li>2 or more questions</li>
         <li>at least one question where people must select a single item from a list of options</li>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -483,6 +483,16 @@ en:
     dropdown_default_text: Select a question to start your route from
     legend_hint_text: A route can only start from a question where people select one item from a list
     legend_text: Which question do you want your route to start from?
+    routing_requirements_not_met_heading: What you need to create a route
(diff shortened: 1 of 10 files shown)


config/locales/en.yml
```diff
@@ -483,6 +483,18 @@ en:
     dropdown_default_text: Select a question to start your route from
     legend_hint_text: A route can only start from a question where people select one item from a list
     legend_text: Which question do you want your route to start from?
+    no_remaining_routes: A route can only start from a question where people select one item from a list. You can only add one route per question.
+    no_remaining_routes_heading: You have no more questions to start a route from
+    routing_requirements_not_met_heading: What you need to create a route
+    routing_requirements_not_met_html: |
+      <p>Before you can add a route you'll need:</p>
```
DavidBiddle, line 490: Can use a curly quote here:
```suggestion
      <p>Before you can add a route you’ll need:</p>
```

spec/models/form_spec.rb
```diff
@@ -130,6 +130,33 @@
...
+      end
+    end
+
+    let(:form) { build :form, pages: selection_pages_with_routes }
+
+    it "returns true if no available routes" do
+      expect(form.has_no_remaining_routes_available?).to eq true
+    end
+
+    context "when there is at least on selection page with no route" do
```
DavidBiddle, line 151:
```suggestion
    context "when there is at least one selection page with no route" do
```

DavidBiddle (dismissed):
Looks good 🎉  just a couple of tiny comments

DavidBiddle (approved):

PR #518, opened 2023-06-29
Don't use mock_gds_sso for tests

lfdebrux:

Our tests currently rely on the gds-sso gem for user management during tests. However, the way the `mock_gds_sso` provider works is a little difficult to follow, and has some surprising side effects. This has been causing me an issue when trying to work on tests, both for the CDDO SSO provider, and for changing the API to use organisation IDs instead of slugs.

This PR changes our tests to use Warden's own test mode instead of relying on the `mock_gds_sso` provider. To achieve this we had to clean up a lot of not obviously related stuff, so this PR is quite long.

I've been working on this as part of https://trello.com/c/75nMIWFC.

### DIFF
--- a/spec/requests/pages_controller_spec.rb
+++ b/spec/requests/pages_controller_spec.rb
@@ -18,6 +18,10 @@ RSpec.describe PagesController, type: :request do
     (build :form, id: 2)
   end
 
+  before do
+    login_as_editor_user
+  end
+
   describe "#index" do
     let(:pages) do
       [build(:page, id: 99),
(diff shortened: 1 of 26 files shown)


jamie-o-wilkinson (approved):
Tested locally and looks good to me. Glad this will be merged before my changes since there's quite a bit of overlap 😅

spec/requests/pages_controller_spec.rb
```diff
@@ -18,6 +18,10 @@
     (build :form, id: 2)
   end
 
+  before do
```
chao-xian, line 21: (Sorry for butting in, could these not go into the application_controller?)

PR #520, opened 2023-06-29
Move bulk form org update to user controller

jamie-o-wilkinson:
Previously a trial user's forms would have their org set via a patch request to a bulk update endpoint in forms-api, triggered by a `before_save` filter in the User model. This overly coupled the User model and Form active resource, such that tests would require mocking of the forms-api requests etc. This how now been avoided by only updating a trial user's forms in the user controller update action.

This is a precursor to making the trial role the default (since doing so means lots of tests would need mocking of the `update_org_for_creator` in forms-api.

Trello card: https://trello.com/c/KkZfWSar

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -20,6 +20,7 @@ class UsersController < ApplicationController
     authorize current_user, :can_manage_user?
 
     if user.update(user_params)
+      user.update_org
       redirect_to users_path
     else
       render :edit, status: :unprocessable_entity
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -20,7 +20,7 @@ class UsersController < ApplicationController
     authorize current_user, :can_manage_user?
 
     if user.update(user_params)
-      user.update_org
+      user.update_user_forms
       redirect_to users_path
     else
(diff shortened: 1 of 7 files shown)


app/controllers/users_controller.rb
lfdebrux, line 1: We should probably have a test that `UsersController#update` does call `user.update_org`, no?

spec/models/user_spec.rb
```diff
@@ -119,26 +119,26 @@
...
+        user = create(:user, role: :trial)
 
-          expect(Form).to receive(:update_org_for_creator).with(user.id, user.organisation.slug)
+        expect(Form).to receive(:update_org_for_creator).with(user.id, user.organisation.slug)
 
-          user.role = role_value
-          user.save!
-        end
+        user.role = role_value
+        user.update_org
```
lfdebrux, line 132: It's kind of weird that we have to call `update_org` after changing the role... I get that we're updating the org for the forms, but when we have these two lines next to each other it's not obvious why they go together. I think it would be quite easy to assume that they're not related, when in fact they are.

I wonder if we calling the new method `update_user_forms` instead, would make things clearer...

lfdebrux (approved):

PR #521, opened 2023-06-29
Explicitly set role in tests

jamie-o-wilkinson:
Prior to making the trial role the default for new users (which is blocked by the migration), we need to explicitly set the user role for tests that currently use the default of `editor` as this default will be changing to `trial`.

This PR also makes the role transitions that will trigger an update of a user's forms' org more explicit, to avoid unexpected role transitions causing requests to be made to the `update-org-for-creator` endpoint incorrectly.

Trello card: https://trello.com/c/KkZfWSar

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -56,7 +56,7 @@ class User < ApplicationRecord
   end
 
   def update_user_forms
-    if role_previously_changed?(from: :trial)
+    if role_previously_changed?(from: :trial, to: :editor) || role_previously_changed?(from: :trial, to: :super_admin)
       Form.update_org_for_creator(id, organisation.slug)
     end
   end
--- a/spec/factories/models/users.rb
+++ b/spec/factories/models/users.rb
@@ -7,12 +7,9 @@ FactoryBot.define do
     role { :editor }
     has_access { true }
 
-    trait :with_super_admin do
-      role { :super_admin }
-    end
-
-    trait :with_trial do
+    trait :with_trial_role do
       role { :trial }
+      with_no_org
     end
 
     organisation_slug { "test-org" }
--- a/spec/requests/forms_controller_spec.rb
+++ b/spec/requests/forms_controller_spec.rb
@@ -95,7 +95,7 @@ RSpec.describe FormsController, type: :request do
           mock.get "/api/v1/forms/2/pages", headers, form.pages.to_json, 200
         end
 
-        login_as build(:user, :with_no_org)
+        login_as build(:user, :with_no_org, role: :editor)
 
         get form_path(2)
       end
@@ -114,7 +114,7 @@ RSpec.describe FormsController, type: :request do
     end
 
     context "with a user with a trial account" do
-      let(:user) { build(:user, role: :trial) }
+      let(:user) { build(:user, :with_trial_role, id: 123) }
       let(:form) { build(:form, :ready_for_live, id: 2, creator_id: user.id) }
 
       before do
@@ -245,7 +245,7 @@ RSpec.describe FormsController, type: :request do
           mock.get "/api/v1/forms?org=", headers, forms_response.to_json, 200
         end
 
-        login_as build(:user, :with_no_org)
+        login_as build(:user, :with_no_org, role: :editor)
 
         get root_path
       end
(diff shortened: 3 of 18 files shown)


spec/requests/forms_controller_spec.rb
```diff
@@ -114,7 +114,7 @@
     end
 
     context "with a user with a trial account" do
-      let(:user) { build(:user, role: :trial) }
+      let(:user) { build(:user, :with_trial_role, id: 123) }
```
lfdebrux, line 117: What happens if a user ID is not set here?

spec/requests/users_controller_spec.rb
```diff
@@ -1,6 +1,10 @@
 require "rails_helper"
 
 RSpec.describe UsersController, type: :request do
+  before do
+    ActiveResource::HttpMock.reset!
```
lfdebrux, line 5: I don't understand why this is necessary

spec/factories/models/users.rb
```diff
@@ -7,12 +7,9 @@
...
     has_access { true }
 
-    trait :with_super_admin do
-      role { :super_admin }
-    end
-
-    trait :with_trial do
+    trait :with_trial_role do
       role { :trial }
+      with_no_org
```
lfdebrux, line 12: Nice idea, I like this!

app/models/user.rb
```diff
@@ -56,7 +56,7 @@ def organisation_valid?
   end
 
   def update_user_forms
-    if role_previously_changed?(from: :trial)
+    if role_previously_changed?(from: :trial, to: :editor) || role_previously_changed?(from: :trial, to: :super_admin)
```
lfdebrux, line 59: I'm a bit nervous about this, because when we add other roles (such as `:admin`) we'll need to update this test. Can we instead make it `&& !` the transitions we don't care about?

lfdebrux (commented):
I've had a read through, I have a couple of small questions, but this looks good 👍

spec/requests/forms_controller_spec.rb
jamie-o-wilkinson: The factory for the user never sets `id`, so the form won't have a `creator_id` set, which means the "you don't have access to this page" page is shown.

lfdebrux: Ah, interesting! That makes sense, thanks for explaining. But then, why were the tests were passing before? 🤔

jamie-o-wilkinson: So that's interesting. It turns out that the default organisation for the form and user factories is the same. Previously we only specified the role, but now with `with_trial_role` we also specify no organisation. It previously worked because if the user's org and form's org match it shows the form (app/policies/form_policy.rb:33), even though the user and creator IDs are nil 😬

spec/requests/users_controller_spec.rb
jamie-o-wilkinson: This was because I had these tests passing in isolation but failing when run alongside all other tests. The failure was related to request mocks existing in this context despite being set up in other tests. I'm hoping this is a red herring and something peculiar to my working copy at the time, as I've not been able to reproduce the same situation since. I've removed this reset now.

spec/requests/forms_controller_spec.rb
lfdebrux: Ah interesting. But that's good though! It means your changes have tightened up the tests, and we're now getting a more accurate check from this test. Thanks for checking.

lfdebrux: It's interesting that a trial user with an org set can see forms from that org... I wonder if we should change that.

app/models/user.rb
jamie-o-wilkinson: I don't _think_ there would be any valid transitions that would be excluded at the moment if we did that. By explicitly including rather than excluding transitions we implicitly exclude invalid transitions like the role being set to nil, or any other garbage that isn't in the roles enum. This should never happen because validation, but I don't think there's any harm in being defensive here as well.

spec/requests/forms_controller_spec.rb
jamie-o-wilkinson: Yes it has opened up that question. My instinct is that is incorrect...

lfdebrux (approved):

spec/requests/forms_controller_spec.rb
jamie-o-wilkinson: I've created this card to track fixing this: https://trello.com/c/hUoW5S1e

I'll fix in a subsequent PR.

PR #529, opened 2023-07-06
Use organisation ID instead of slug to associate organisations with forms

lfdebrux:

Trello card: https://trello.com/c/pt9lphfe

We want to associate forms in the forms-api database with an organisation using the organisation primary key, as this is more stable than the organisation slug. This PR makes the changes needed in the forms-admin app. The changes needed in forms-api have already been merged and deployed, see alphagov/forms-api#295.

This is quite a big changeset, so I've tried to split it into more commits than strictly necessary so if you review commit by commit it's less fatiguing :sweat_smile:.

~Note that this PR should be merged only after the script `copy_organisation_id_from_admin_to_api` in PR https://github.com/alphagov/forms-deploy/pull/244 has been run on each deployment environment.~ This is now ready to merge, as the script has been run.


thomasiles:
This looks good to me, nice one.

Is there any benefit to splitting of the changes to creating a form to add both org_slug and org_id before running the migartion? That might stop the need to run the migration twice to catch the stragglers.

lfdebrux:
> Is there any benefit to splitting of the changes to creating a form to add both org_slug and org_id before running the migration? That might stop the need to run the migration twice to catch the stragglers.

Hmm, that's a good idea, I hadn't thought of that. I'd need to think about it more but that could work.

lfdebrux:
> > Is there any benefit to splitting of the changes to creating a form to add both org_slug and org_id before running the migration? That might stop the need to run the migration twice to catch the stragglers.
> 
> Hmm, that's a good idea, I hadn't thought of that. I'd need to think about it more but that could work.

Okay, I've now split this PR into two, #537 has now been merged and deployed, so this PR is changing the admin app to use the organisation ID instead of the slug, and then removing the slug to clean up.

jamie-o-wilkinson (approved):
I think this looks good to me. Nice work 👏

PR #536, opened 2023-07-09
Add success messages to various user journeys

aliuk2012:

Added all the success messages in https://docs.google.com/document/d/1MpxzrabkqY-st-RW78mdetXekpLgmprOfsvxpsB2kRk/edit except for 

>After a user saves a question, show the notification on the next page
>Question x has been saved
It will mean making more requests to find the pages position for newly created pages 

Trello card:https://trello.com/c/MXWkGpUw/900-add-banner-notifications-into-the-product-after-a-successful-action

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/forms/delete_confirmation_controller.rb
+++ b/app/controllers/forms/delete_confirmation_controller.rb
@@ -68,8 +68,7 @@ module Forms
 
       authorize form, :can_view_form?
       if form.destroy
-        flash[:message] = "Successfully deleted #{form.name}"
-        redirect_to success_url, status: :see_other
+        redirect_to success_url, status: :see_other, success: "Successfully deleted ‘#{form.name}’"
       else
         raise StandardError, "Deletion unsuccessful"
       end
@@ -80,8 +79,7 @@ module Forms
 
       authorize form, :can_view_form?
       if page.destroy
-        flash[:message] = "Successfully deleted #{page.question_text}"
-        redirect_to success_url, status: :see_other
+        redirect_to success_url, status: :see_other, success: "Successfully deleted ‘#{page.question_text}’"
       else
         raise StandardError, "Deletion unsuccessful"
       end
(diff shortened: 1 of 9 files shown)


jamie-o-wilkinson (commented):
Do we have existing test coverage for success banners? Just wondering if (especially for the cases with some logic involved, e.g. app/controllers/pages_controller.rb) we should be adding corresponding tests or that doesn't provide much value here

aliuk2012:
@jamie-o-wilkinson I'm not sure it adds much value to the tests. The most important part is the redirecting and making sure users end up on the correct page.

aliuk2012:
If you remove the `success:...` part after the redirect i18n-tasks will fail because it has locales that arent being used and the same if you were to delete the translations it would fail because of missing translations

lfdebrux (approved):
I tested these locally, they all work for me 🙌

app/controllers/forms/delete_confirmation_controller.rb
```diff
@@ -68,8 +68,7 @@ def delete_form(form)
 
       authorize form, :can_view_form?
       if form.destroy
-        flash[:message] = "Successfully deleted #{form.name}"
-        redirect_to success_url, status: :see_other
+        redirect_to success_url, status: :see_other, success: "Successfully deleted ‘#{form.name}‘"
```
lfdebrux, line 71: I think this is two left curly quotes, it should be:

```suggestion
        redirect_to success_url, status: :see_other, success: "Successfully deleted ‘#{form.name}’"
```

app/controllers/forms/delete_confirmation_controller.rb
```diff
@@ -80,8 +79,7 @@ def delete_page(form, page)
 
       authorize form, :can_view_form?
       if page.destroy
-        flash[:message] = "Successfully deleted #{page.question_text}"
-        redirect_to success_url, status: :see_other
+        redirect_to success_url, status: :see_other, success: "Successfully deleted ‘#{page.question_text}‘"
```
lfdebrux, line 82: I think this is two left curly quotes, it should be:

```suggestion
        redirect_to success_url, status: :see_other, success: "Successfully deleted ‘#{page.question_text}’"
```

lfdebrux (approved):

PR #538, opened 2023-07-10
Prevent setting of form org data for trial users

jamie-o-wilkinson:
Currently, if a trial user has organisation data, this will get set on any forms that user creates. We want to prevent that so that other non-trial users in the same organisation don't see the trial user's forms.

Trello card: https://trello.com/c/aZo4snWS

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/app/controllers/forms/change_name_controller.rb
+++ b/app/controllers/forms/change_name_controller.rb
@@ -7,16 +7,17 @@ module Forms
     end
 
     def create
-      form = Form.new({
-        name: params[:name],
-        org: current_user.organisation&.slug,
-        organisation_id: current_user.organisation_id,
-        creator_id: current_user.id,
-      })
+      form_args = { name: params[:name], creator_id: @current_user.id }
 
+      # don't set organisation data for forms created by trial users
       if @current_user.trial?
-        form.submission_email = @current_user.email
+        form_args[:submission_email] = @current_user.email
+      else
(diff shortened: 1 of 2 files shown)


app/controllers/forms/change_name_controller.rb
```diff
@@ -7,16 +7,21 @@ def new
...
+                 submission_email: @current_user.email,
+               })
+             else
+               Form.new({
+                 name: params[:name],
+                 org: @current_user.organisation&.slug,
+                 organisation_id: @current_user.organisation_id,
+                 creator_id: @current_user.id,
+               })
+             end
```
lfdebrux, line 24: I feel like there should be a way to write this with less repetition...

lfdebrux (dismissed):

spec/requests/forms/change_name_controller_spec.rb
```diff
@@ -58,7 +58,7 @@
         mock.put "/api/v1/forms/2", post_headers
         mock.post "/api/v1/forms", post_headers, { id: 2 }.to_json, 200
       end
-      post new_form_path, params: { forms_change_name_form: { name: form_data[:name] } }
+      post new_form_path, params: { forms_change_name_form: form_data }
```
lfdebrux, line 61: I don't think this line should change

lfdebrux (approved):

PR #539, opened 2023-07-10
Fix sign up action when not using auth0

lfdebrux:

Closes https://governmentdigitalservice.pagerduty.com/incidents/Q0TMFSI4QHG0GD

Just calling `redirect_to_omniauth` is not the right way to send the user to the sign in page; that action expects to be called after Warden has done its thing. Instead we do the the redirect ourselves.


jamie-o-wilkinson (approved):
Nice one 👍

PR #543, opened 2023-07-14
Add automated accessibility tests

DavidBiddle:
- Adds axe-core-rspec for running automated accessibilty checks in our tests (see gem info below)
- Adds accessibility checks to our existing form feature tests 
- Adds feature tests to our existing components, which run against the component previews and check for accessibility issues. 
- Adds a couple of convenience methods:
  - Inside a `context "..." do` block within a feature test, you can use the shared examples `it_behaves_like "a page with no axe errors"` and `it_behaves_like "a component with no axe errors"` to add a test which checks the page for accessibility issues
  - Inside an `it "..." do` block within a feature test, you can use the methods `expect_page_to_have_no_axe_errors(page)` and `expect_component_to_have_no_axe_errors(page)` to assert that the page or component the browser is currently on doesn't have any errors.
- Adds a `component-preview` layout which we can use if a component has issues in combination with the normal page layout
  - in this case I'm using it for the header component, which looks extremely odd when previewed on a layout which already has a header, and also causes accessibility issues because there are ID and `aria-` attribute collisions between the two headers.

Trello card: https://trello.com/c/BARdutlj/916-explore-adding-automated-accessibility-tests

### New gem assessment
This PR adds a new gem, [axe-core-rspec](https://rubygems.org/gems/axe-core-rspec). To review the gem I looked at its stats on [ruby-toolbox.com](https://ruby-toolbox.com/), and also checked security issues using GitHub Security Advisories and the [snyk.io vulnerability database](https://security.snyk.io/).

The gem is maintained by Deque Systems, a major accessibility organisation which also develops the underlying axe-core library that this gem uses.

#### Does it solve a specific problem we have?
Yes, it lets us run automated accessibility tests within our Capybara feature tests. 
#### Is it updated regularly?
Yes, 16 releases so far this year
#### Is it widely used?
Somewhat - 2,973,244 downloads
#### Is it well documented?
The documentation is good, given the limited API of the gem.
#### Does it have any major security issues?
None that I could see.
#### Would writing the code ourselves have a better outcome?
Unlikely - we'd almost certainly still end up using Deque's underlying axe-core library, so we'd still have a dependency on Deque's code and wouldn't gain much. 
#### How much hassle would it be if this gem stopped being maintained?
Some, but not a lot. We would have to rewrite our accessibility tests in another framework (e.g. using jest-axe or a similar library) and we'd likely lose the convenience of having our accessibility tests integrated with our other rspec tests, but that doesn't feel like a major disadavantage.

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)


spec/components/page_list_component/page_list_component_preview.rb
```diff
@@ -6,25 +6,27 @@ def default
   end
 
   def with_pages_and_no_conditions
-    pages = [OpenStruct.new(id: 1, question_text: "Enter your name", routing_conditions: []),
-             OpenStruct.new(id: 2, question_text: "What is your pet's phone number?", routing_conditions: []),
-             OpenStruct.new(id: 3, question_text: "How many pets do you own?", routing_conditions: [])]
+    pages = [Page.new(id: 1, position: 1, is_optional: false, question_text: "Enter your name", routing_conditions: []),
+             Page.new(id: 2, position: 2, is_optional: false, question_text: "What is your pet's phone number?", routing_conditions: []),
+             Page.new(id: 3, position: 3, is_optional: false, question_text: "How many pets do you own?", routing_conditions: [])]
```
aliuk2012, line 11: Why do these need to be specific model classes?

DavidBiddle: Because we're using some of the class methods like `is_optional?` and `conditions` in the component. I guess we could mock these in an OpenStruct, but is there a reason we wouldn't want to use the model class?

aliuk2012: While it makes it hard to reuse these components then without the model classes. I think we should mock those out in the OpenStruct. TBH, it just means that the component requires a collection of objects and those objects should have a method called "is_optional" it shouldn't care that its a Page.

thomasiles (approved):
This looks really cool to me 🏅

app/views/layouts/component_preview.html.erb
```diff
@@ -0,0 +1,51 @@
...
+        <% end %>
+
+        <%= yield %>
+      </main>
+    </div>
+
+    <%= vite_client_tag %>
+    <%= vite_javascript_tag 'application' %>
+  </body>
+</html>
```
aliuk2012, line 51: 🤔  how is this different from our existing application layout?

DavidBiddle: Doesn't include the header or footer (so the component can be viewed in isolation if needed). I'm only using it for the header (which has always looked quite odd in preview because it's duplicated) but we could use it for other components later if we feel the need.

aliuk2012: I'm wondering if we could use this one, as a base then put in the placeholders for header/footer and then for our normal application template we use this based template but set headers and footer

DavidBiddle: Good idea. I'm inclined to look at that in a separate PR though since if I'm messing with application.html.erb I want it to be easy to revert 😁

PR #544, opened 2023-07-14
Explicitly assign users to organisations in database seed

lfdebrux:
The database seed is currently broken, because the user factory if not given a user will try to create an organisation with slug of `test-org` and id of 1, which isn't possible when `government-digital-service` was created first. Assigning the organisation explicitly fixes this and is the better way to go anyway.


jamie-o-wilkinson (approved):
Beat me to it, sorry and thanks! 👏

PR #545, opened 2023-07-17
Convert component_preview layout to a base layout...

DavidBiddle:
... which the application layout inherits from

@aliuk2012 pointed out that the `component_preview` layout added in #543 is essewntially a 'base' layout, and that we could make application.html.erb inherit from it. This PR does that.

Maybe worth noting - this creates a few new `yield` regions in the layout:
- `skip_link`
- `header`
- `before_content` (includes the phase banner and the back link)
- `footer`
The names and placements of these new regions correspond to the area names in the [Design System page template](https://design-system.service.gov.uk/styles/page-template/)

Trello card:

#### Checklist

  - [n/a] README.md


spec/components/header_component/header_component_preview.rb
```diff
@@ -1,5 +1,5 @@
 class HeaderComponent::HeaderComponentPreview < ViewComponent::Preview
-  layout "component_preview"
+  layout "base"
```
aliuk2012, line 2: I think it would be worth to use this layout as a default and not just for this component. 

https://github.com/alphagov/forms-admin/blob/main/app/controllers/component_preview_controller.rb#L5

DavidBiddle: Done!

PR #549, opened 2023-07-19
Add Jest tests to CI workflow

DavidBiddle:
This PR adds the jest tests to the CI workflow.

#### Checklist

  - [n/a] README.md

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,8 +49,10 @@ jobs:
       - name: Set up database schema
         run: bin/rails db:schema:load
       # Add or replace test runners here
-      - name: Run tests
+      - name: Run Ruby tests
         run: bundle exec rspec
+      - name: Run JS tests
+        run: npm test
       # Add or replace any other lints here
       - name: Security audit dependencies
         run: bundler exec bundle audit check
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -31,27 +31,30 @@ jobs:
...
+        uses: ./.github/workflows/build
       # Add or replace test runners here
-      - name: Run tests
+      - name: Run Ruby tests
         run: bundle exec rspec
-      # Add or replace any other lints here
+  lint:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
```
danworth, line 42: Its a bit confusing that we need postgres to run the lint but this only seems necessary because the common build task runs db:migrate (along with installing chrome which also doesn't seem necessary to lint). Do we need to call the common build task for linting besides the need to install ruby/gems which might be simpler to just call directly within the linting job?

.github/workflows/build/action.yml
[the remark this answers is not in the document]
DavidBiddle: Good spot!

.github/workflows/test.yml
DavidBiddle: I've gone with keeping the shared action and moving the postgres bit into the `build_and_test` action. The shared setup is now just installing Ruby and the BE and FE dependencies, but I still think that makes sense as a reusable bit of shared setup

jamie-o-wilkinson (approved):

PR #558, opened 2023-07-26
Pin to ChromeDriver v114

danworth:
ChromeDriver v115 has moved to a different location which breaks the install action at present. We will need to address this properly once the install chrome driver action supports installing v115 from the new location.

Allows the browser tests to pass once again.

Note it is necessary to tell `webdriver` to use the pinned version.


lfdebrux (approved):
🥇

PR #561, opened 2023-07-26
Only Bundle Install Production Scope Only

danworth:
We do not need to install the development and test scoped dependencies into our docker image.

We shouldn't install non production scoped dependencies into the docker image we use to run the app in production.

I noticed this because of a change to configure the test dependeny `webdriver`: https://github.com/alphagov/forms-admin/pull/560

Before this change `webdriver` was installed in our docker image:
```
GDS11321:forms-admin dan.worth$ docker run --rm -it my_admin gem query --local | grep webdriver
selenium-webdriver (4.9.0)
webdrivers (5.2.0)
```
After this change it is not:
```
GDS11321:forms-admin dan.worth$ docker run --rm -it my_admin gem query --local | grep webdriver
GDS11321:forms-admin dan.worth$

```

I have deployed this to dev and the end-to-end tests pass

I'll make similar changes to the other apps shortly.
#### Checklist

- [ x] I've used the pull request template
  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,10 @@ USER ruby
 
 COPY --chown=ruby:ruby Gemfile* ./
 RUN gem install bundler -v 2.4.10
-RUN bundle install --jobs "$(nproc)"
+RUN bundle config set --local without development:test \
+  && bundle config set --local jobs "$(nproc)"
+
+RUN bundle install
 
 COPY --chown=ruby:ruby package.json ./
 RUN npm install --ignore-scripts


Dockerfile
```diff
@@ -12,7 +12,7 @@ USER ruby
 
 COPY --chown=ruby:ruby Gemfile* ./
 RUN gem install bundler -v 2.4.10
-RUN bundle install --jobs "$(nproc)"
+RUN bundle install --jobs "$(nproc)" --without development test
```
lfdebrux, line 15: I think we probably want to use `bundle config` instead of adding the `--without` to the `install` command, as that is deprecated and going away in bundler 3 [[1]]. At the same time we could also set deployment mode [[2]]:

```suggestion
RUN bundle config set --local without development:test && bundle config set --local development true
RUN bundle install --jobs "$(nproc)"
```

[1]: https://bundler.io/v2.4/man/bundle-config.1.html#REMEMBERING-OPTIONS
[2]: https://bundler.io/guides/deploying.html

danworth: Great suggestion Laurence. I've created a ticket for using the deployment mode because it'll need a bit more consideration to implement since it has consequences on the dependency locations and hence what we copy between build and app stages as well as the commands to run rake/rails etc. I've put it here as a ticket for us to pick up as BAU: https://trello.com/c/8di3mMvr/147-use-bundle-deployment-mode-when-building-docker-images

I'll continue with not installing test and dev dependencies using the `bundle config` approach because that's a quick fix.

danworth: @lfdebrux I've also deployed this new commit using `bundle config` to dev and it works as expected. If you'd like to review again please 👍

lfdebrux (approved):

PR #563, opened 2023-07-27
Remove devDependencies after asset build

DavidBiddle:
We have two sets of node dependencies in our package.json - `dependencies` which are packages we need to ship with our app, and `devDependencies` which we need to build our app (or develop locally) but don't need to ship with our app at run time. We have many more `devDependencies` than we do `dependencies`.

In our Docker builds we install our devDependencies so that we can run the frontend asset build. We then copy the entire `app` folder into our runtime environment. Shipping these extra dependencies at run time increases our app's attack surface, so we should remove them once we no longer need them. 

In this PR, I'm running `npm ci --only=production` at the end of the build stage, which will clear out our node_modules file and install only our production dependencies. This means none of our devDependencies will get copied over during the run stage.

I've also changed the build stage to use `npm ci` instead of `npm install` during the initial build stage. This will do a clean install based on the `package-lock.json` file, so our build will be more deterministic.

#### Checklist

  - [n/a] README.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,8 @@ RUN bundle config set --local without development:test \
 
 RUN bundle install
 
-COPY --chown=ruby:ruby package.json ./
-RUN npm install --ignore-scripts
+COPY --chown=ruby:ruby package.json package-lock.json ./
+RUN npm ci --ignore-scripts
 
 ENV RAILS_ENV="${RAILS_ENV:-production}" \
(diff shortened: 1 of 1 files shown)


danworth (dismissed):
Nice change 🏅

danworth (approved):
🏅 Removing non prod dependencies in the final stage makes sense 👍

PR #564, opened 2023-07-27
Remove Webdrivers

danworth:
webdrivers isn't necessary with selenium-webdriver versions greater than 4.10. Also remove setup-chromedriver which shouldn't be necessary and align with our test setup in forms-runner.

Necessary to bump the minor version of Capybara for this to work (now same version as in forms-runner).

Aligns forms-runner and forms-admin setup. Get CI working again without pinning to a particular chromedriver version.

webdrivers isn't necessary with selenium-webdrivers >4.10 https://github.com/titusfortner/webdrivers/issues/247#issuecomment-1641139415

Github runners already include chrome and chromedriver v115 https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers

v 4.11 of selenium-webdriver will include downloading browsers and drivers as necessary: https://github.com/titusfortner/webdrivers/issues/247#issuecomment-1650315937

#### Checklist

  - [ ] README.md
  - [ ] Elsewhere (please link)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,10 +32,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
       # Add or replace dependency steps here
-      - name: Install Chrome for integration/e2e tests
-        uses: nanasess/setup-chromedriver@v2.0.0
-        with:
-          chromedriver-version: "114.0.5735.90"
       # The ruby version is taken from the .ruby-version file, no need to specify here.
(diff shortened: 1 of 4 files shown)


lfdebrux (approved):
Tested this locally, and it seems to work. I worry about what will happen as Chrome and Selenium versions advance, but this is definitely simpler than what we had before, so hopefully if there are issues in future we can understand the environment better.

PR #577, opened 2023-08-03
Add paper trail

thomasiles:

Trello card: [add paper trail to admin](https://trello.com/c/xB8vHrAU/938-add-paper-trail-gem-to-forms-admin-to-track-changes-to-users-orgs-objects)

Add paper_trail gem to to allow us to track changes to the User and Organisation models. This will allow us to keep a history of changes to role.

The 'whodunnit' hook is added to the controller to add the current user id into the versions table to allow tracing. I think this should be fine across all our environments and login methods, now that current_user is defined. It's possible to set this for rake tasks and console use - there is a link in the commit message to the paper_trail docs.

I've set paper_trail to only track the fields super_admins can change for now - role, organisation and has_access. I've done this because I was concerned that a lot of irrelevant information would get scooped up into the versions table every time a user logged in. I could be wrong - it might be useful to track changes to name/email/provider and tracking other info won't be an issue. I'd like opinions on this :) It also captures create events on user, we might find that using the paper_trail :skip attribute is useful to keep login specific attributes out of the version table.

I've checked the behaviour of the `rake organisations:fetch` task. It works, so we don't need to change it. It only adds versions when organisations are created/updated/deleted, not when there are no changes. When the task is run, no whodunnit user is set.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,8 @@ class ApplicationController < ActionController::Base
   before_action :set_request_id
   before_action :check_maintenance_mode_is_enabled
   before_action :authenticate_and_check_access
+  before_action :set_paper_trail_whodunnit
+
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
   before_action :clear_questions_session_data
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,5 +1,6 @@
 class User < ApplicationRecord
   include GDS::SSO::User
+  has_paper_trail only: %i[role organisation_id has_access]
 
   class UserAuthenticationException < StandardError; end
 
--- a/db/migrate/20230802102230_add_object_changes_to_versions.rb
+++ b/db/migrate/20230802102230_add_object_changes_to_versions.rb
@@ -1,0 +1,8 @@
+# This migration adds the optional `object_changes` column, in which PaperTrail
+# will store the `changes` diff for each update event. See the readme for
+# details.
+class AddObjectChangesToVersions < ActiveRecord::Migration[7.0]
+  def change
+    add_column :versions, :object_changes, :jsonb
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -67,6 +67,9 @@ gem "govuk_notify_rails"
 # Use validate_url so we don't have to write custom URL validation
 gem "validate_url"
 
+# For auditing tables
+gem "paper_trail"
+
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem "debug", platforms: %i[mri mingw x64_mingw]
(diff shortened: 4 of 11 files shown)


db/migrate/20230426150135_add_organisations_for_existing_users.rb
```diff
@@ -1,23 +1,27 @@
 class AddOrganisationsForExistingUsers < ActiveRecord::Migration[7.0]
   def up
-    Organisation.create!(content_id: "af07d5a5-df63-4ddc-9383-6a666845ebe9", slug: "government-digital-service", name: "Government Digital Service")
+    PaperTrail.request(enabled: false) do
```
thomasiles, line 3: Changing an existing migration feels like one of those things you should never do! Are there any alternatives here though?

The issue is using the model within the migration. If the model relies on tables which have only been added in later migrations, here versions added by paper_trail, the migration will fail.

app/models/user.rb
```diff
@@ -1,5 +1,6 @@
 class User < ApplicationRecord
   include GDS::SSO::User
+  has_paper_trail only: %i[role organisation_id has_access]
```
thomasiles, line 3: Is this too conservative? This is either a sensible precaution not to log too much information or might be my fear of the unknown.

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,8 @@ class ApplicationController < ActionController::Base
   before_action :set_request_id
   before_action :check_maintenance_mode_is_enabled
   before_action :authenticate_and_check_access
+  before_action :set_paper_trail_whodunnit
```
thomasiles, line 8: This would have caused issues in the past using basic_auth. I think now we are on more stable ground and current_user is always available?

app/models/user.rb
jamie-o-wilkinson: I think this seems reasonable to me...Would we care about the history of the other attributes?

db/migrate/20230802102230_add_object_changes_to_versions.rb
```diff
@@ -0,0 +1,12 @@
+# This migration adds the optional `object_changes` column, in which PaperTrail
+# will store the `changes` diff for each update event. See the readme for
+# details.
+class AddObjectChangesToVersions < ActiveRecord::Migration[7.0]
+  # The largest text column available in all supported RDBMS.
+  # See `create_versions.rb` for details.
+  TEXT_BYTES = 1_073_741_823
```
aliuk2012, line 7: is this needed?

app/controllers/application_controller.rb
lfdebrux: Yes we're using Warden for all auth providers so we always have `current_user`

db/migrate/20230802102230_add_object_changes_to_versions.rb
thomasiles: no - good spot!

db/migrate/20230426150135_add_organisations_for_existing_users.rb
thomasiles: This has been solved in another commit, by deleting the migration.

DavidBiddle (approved):

jamie-o-wilkinson (approved):

PR #580, opened 2023-08-03
Add task to update organisations table

lfdebrux:

Trello card: https://trello.com/c/WEhgdg25 

This PR adds a Rake task that we can run to update the organisations table using the data from the [GOV.UK organisations API](https://docs.publishing.service.gov.uk/repos/collections/api.html). It is very heavily cribbed off of [what GOV.UK Signon does](https://github.com/alphagov/signon/blob/b7a53e282c55d8ef3ab6369a7cb358b6ae100d27/lib/organisations_fetcher.rb).

Note, we don't really have a way of running this in production yet, as we don't haven't got a way to run Rake commands, but I figured we might as well merge this now for when we do.

### DIFF
--- a/app/views/users/edit.html.erb
+++ b/app/views/users/edit.html.erb
@@ -45,7 +45,7 @@
         end
       end %>
 
-      <%= f.govuk_collection_select :organisation_id, Organisation.order(:slug), :id, :name,
+      <%= f.govuk_collection_select :organisation_id, Organisation.order(:name), :id, :name,
         class: ['govuk-!-width-three-quarters'],
         options: { prompt: t('users.edit.organisation_prompt') },
         label: { text: t('users.edit.organisation'), size: 'm', tag: 'h2' },
--- a/lib/advisory_lock.rb
+++ b/lib/advisory_lock.rb
@@ -1,0 +1,17 @@
+class AdvisoryLock
+  module DSL
+    def with_lock(key)
+      lock_id = Zlib.crc32(key)
+      connection = ActiveRecord::Base.connection
+
+      got_lock = connection.get_advisory_lock(lock_id)
+      if got_lock
+        yield
+      else
+        puts "Skipping task, couldn't obtain lock: #{key}"
+      end
+    ensure
+      connection.release_advisory_lock(lock_id) if got_lock
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


thomasiles (approved):
This looks nice! It worked for me locally.

I like the inclusion of the lock (although I don't fully understand it :)) and pulling out the logic of the task into a separate task

A couple of questions/thoughts
- should we have some logging in here, to indicate success/failure
- there are a few exceptions thrown - I don't know what happens to these - are they captured by rake and printed to stdout? I guess they won't go through the JSON logger, if they do? I guess they are not captured by sentry which I think we want to avoid.
- as the code is in a class, can we add testing and is it worth it?
- I guess failure halfway through the process doesn't matter - the solution is to any issues is always re-run it?

lfdebrux:
> This looks nice! It worked for me locally.

🙌 

> I like the inclusion of the lock (although I don't fully understand it :)) and pulling out the logic of the task into a separate task

That was all shamelessly ripped from alphagov/signon, I can't claim any credit for it 😅 
 
> A couple of questions/thoughts
> 
> * should we have some logging in here, to indicate success/failure

It might be nice, I'm not sure what sort of information we want to log though? Other than 'this is happening'?

> * there are a few exceptions thrown - I don't know what happens to these - are they captured by rake and printed to stdout? I guess they won't go through the JSON logger, if they do? I guess they are not captured by sentry which I think we want to avoid.

Good question. I actually had an exception occur in testing in dev and it did end up in Sentry, see https://govuk-forms.sentry.io/issues/4371137842/. Weirdly though the  exception messages on stdout were not JSON formatted, see https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/forms-admin-dev/log-events/forms-admin-dev$252Fforms-admin$252Fa52ee83c7b0f41e19b94c0ba6e862dd6.

I don't know why things turned out this way, my understanding is that the `:environment` task which our custom rake task requires sets up Rails somehow, maybe some of the initialisers in `config` are not run and some are?

> * as the code is in a class, can we add testing and is it worth it?

I think we could add some testing, but I'm not sure to what extent; it might be difficult. I didn't tackle it because I wasn't sure it would be worth the effort; it feels like something that depends too much on an external service for testing to be able to have good coverage. It's more important that we're running it regularly and catching exceptions.

> * I guess failure halfway through the process doesn't matter - the solution is to any issues is always re-run it?

Yes, I didn't think implementing retries of any kind was worth it; we just retry the whole task.

thomasiles:
awesome, thanks

PR #581, opened 2023-08-04
Improve HTML structure for additional guidance

DavidBiddle:

Trello card: None but part of the detailed guidance work

Includes two tweaks to fix some invalid HTML:
- Fixes an issue where `<p>` tags were being rendered inside `<p>` tags
- customises the govuk-details component HTML so that we're not rendering a `h2` inside a `span`

Also includes a change to make our markdown examples more semantic:
- Wraps our markdown examples in `pre` and `code` tags as this is more semantically appropriate for code samples
  -  includes some styling to make these look the same as they do now

### DIFF
--- a/app/views/pages/additional_guidance.html.erb
+++ b/app/views/pages/additional_guidance.html.erb
@@ -29,8 +29,9 @@
 
           <p><%= simple_format(t("guidance.formatting_help.#{format_name}.instructions")) %></p>
 
-          <%= govuk_inset_text do %>
-            <%= simple_format(t("guidance.formatting_help.#{format_name}.example")) %>
+            <%= govuk_inset_text do %>
+              <pre class="app-markdown-example-block"><code class="app-markdown-example-block__code"><%= t("guidance.formatting_help.#{format_name}.example") %></code ></pre>
+            <% end %>
           <% end %>
         <% end %>
       <% end %>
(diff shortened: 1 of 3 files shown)


app/views/pages/additional_guidance.html.erb
```diff
@@ -22,18 +22,24 @@
...
-          <p><%= simple_format(t("guidance.formatting_help.#{format_name}.instructions")) %></p>
-
-          <%= govuk_inset_text do %>
-            <%= simple_format(t("guidance.formatting_help.#{format_name}.example")) %>
+      <details class="govuk-details" data-module="govuk-details">
+        <summary class="govuk-details__summary">
+          <h2 class='govuk-details__summary-text govuk-!-font-size-19 govuk-!-margin-0 govuk-!-font-weight-regular'>
+            <%= t("guidance.formatting_help.heading") %>
+          </h2>
+        </summary>
```
aliuk2012, line 30: We should really be using the details component where we can. Maybe we could offer a PR upstream to toggle which tag to use inside the summary.

DavidBiddle: In this case the govuk-components gem is behaving the same way as the Design System nunjucks macro, so we might need to get both to change their behaviour.

I've added a note about what we're doing to the Design System discussion of the component here: https://github.com/alphagov/govuk-design-system-backlog/issues/44#issuecomment-1664275484

aliuk2012 (approved):

PR #582, opened 2023-08-04
Add autocomplete orgs

thomasiles:
# Add autocomplete for choosing organisation when editing users

Trello card: [trello](https://trello.com/c/ZLGz6aBR/951-use-autocomplete-component-for-choosing-a-users-organisation)

We currently don't have an easy way for admin users to select from large lists of choices.

This commit adds the GOV.UK accessible autocomplete component:

https://github.com/alphagov/accessible-autocomplete#api-documentation

This component is supported by the design system team but it isn't a
priority.

Given that we are using it only within the admin we are probably safe to
use it.

image

https://github.com/DFE-Digital/dfe-autocomplete

We use the dfe-autocomplete package. The other option is to wrap the
accessible autocomplete component (which is JS and CSS) ourselves. There
doesn't seem to be much benefit to doing that, but it's an option if, in
the future the dependency on the DFE component becomes awkward.

Neither the Ruby Gem or node package have been published, both are
fetched from GitHub. This means we now rely on Git to bundle and node
install. Luckily we have just added Git to our docker images so there is
no new dependency.

- [x] Does it solve a specific problem we have?
        Yes, it packages up the accessible autocomplete component and
        allows us to easily integrate it into rails forms.
- [x] Is it updated regularly?
        Yes, it's had a couple of changes this year
- [x] Is it widely used?
        Doesn't seem to be used outside of DFE who have a handful of
        large projects which use it
- [x] Is it well documented?
        It's not bad - there isn't much to document but the readme is
        good
- [x] Does it have any major security issues?
        I don't think so
- [x] Would writing the code ourselves have a better outcome?
        We would be doing something very similar and probably wouldn't
        do it any better. We can always do it ourselves if it ever seems
          like we need to.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ WORKDIR /app
 
 RUN apk update
 RUN apk upgrade --available
-RUN apk add libc6-compat openssl-dev build-base libpq-dev nodejs=~18 npm git
+RUN apk add libc6-compat openssl-dev build-base libpq-dev nodejs=~18 npm git python3
 RUN adduser -D ruby
 RUN mkdir /node_modules && chown ruby:ruby -R /node_modules /app
 
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,8 @@ gem "vite_rails"
 # For GOV.UK branding
 gem "govuk-components", "~> 4.1.0"
 gem "govuk_design_system_formbuilder", "~> 4.1.1"
+# The autocomplete component is not currentky published as a gem
+gem "dfe-autocomplete", require: "dfe/autocomplete", github: "DFE-Digital/dfe-autocomplete"
 
 # Our own custom markdown renderer
 gem "govuk-forms-markdown", github: "alphagov/govuk-forms-markdown", tag: "0.1.0"
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -30,10 +30,23 @@ class UsersController < ApplicationController
 private
 
   def user_params
-    params.require(:user).permit(:has_access, :role, :organisation_id)
+    params.require(:user).permit(:has_access, :role, :organisation_id).tap do |p|
+      # We have to take steps to detect when the autocomplete compoenent is
+      # empty. We use the value of rawAttribute, which is the text input when JS
+      # is enabled. When it's empty, the user has cleared it. This isn't needed
+      # when the no JS select is used but we have to allow organisation_id to be
+      # nil still.
+      if p.key?(:organisation_id) && organisation_id_raw && organisation_id_raw.empty?
+        p[:organisation_id] = nil
+      end
+    end
   end
 
   def user
     @user ||= User.find(params[:id])
+  end
+
+  def organisation_id_raw
+    params.dig(:user, :organisation_id_raw)
   end
 end
--- a/app/frontend/entrypoints/application.js
+++ b/app/frontend/entrypoints/application.js
@@ -14,4 +14,15 @@ document
 
 initAll()
 
-dfeAutocomplete({})
+// We set rawAttribute to true to pass the value of the text input to the
+// controller when the autocomplete is being used, so we can detect when the
+// user has cleared the field. We set source to false to overide the DfE
+// component's handling of sort order and synonyms, as described in the code
+// found in
+// https://github.com/DFE-Digital/dfe-autocomplete/blob/36d80e6b5bba67c92cd9ec6982a4e536d1889aed/src/dfe-autocomplete.js#L54C1-L54C1
+// this means the autocomplete will get options straight from the select
+// element, removing the null option, which isn't needed for autocomplete.
+dfeAutocomplete({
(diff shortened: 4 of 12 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@ WORKDIR /app
 
 RUN apk update
 RUN apk upgrade --available
-RUN apk add libc6-compat openssl-dev build-base libpq-dev nodejs=~18 npm git
+RUN apk add libc6-compat openssl-dev build-base libpq-dev nodejs=~18 npm git python3
```
thomasiles, line 7: Had to add this because accessible autocomplete uses node-sass no dart-sass, which is sad.

Gemfile
```diff
@@ -52,6 +52,8 @@ gem "vite_rails"
 # For GOV.UK branding
 gem "govuk-components", "~> 4.1.0"
 gem "govuk_design_system_formbuilder", "~> 4.1.1"
+# The autocomplete component is not currentky published as a gem
+gem "dfe-autocomplete", require: "dfe/autocomplete", github: "DFE-Digital/dfe-autocomplete"
```
thomasiles, line 56: The dfe/autocomplete component hasn't been published - like our markdown Gem. This makes it feel a little less stable. I guess we can always fork it/ inline it/ come up with something else if it turns out to be an issue.

spec/views/users/edit.html.erb_spec.rb
```diff
@@ -77,10 +77,6 @@
     it "shows no organisation set" do
       expect(rendered).to have_text("No organisation set")
     end
-
```
thomasiles, line 80: We lose the select an organisation text - we could add something similar back, either by calling the blank option "select an organisation" or by using placeholder text. I'm not sure either is recommended by the autocomplete component though?

app/frontend/entrypoints/application.js
```diff
@@ -12,3 +13,16 @@ document
...
+// controller when the autocomplete is being used, so we can detect when the
+// user has cleared the field. We set source to false to overide the DfE
+// component's handling of sort order and synonyms, as described in the code
+// found in
+// https://github.com/DFE-Digital/dfe-autocomplete/blob/36d80e6b5bba67c92cd9ec6982a4e536d1889aed/src/dfe-autocomplete.js#L54C1-L54C1
+// this means the autocomplete will get options straight from the select
+// element, removing the null option, which isn't needed for autocomplete.
+dfeAutocomplete({
+  rawAttribute: true,
+  source: false
```
thomasiles, line 27: This is needed to stop the autocomplete including "Select an organisation" from the list of options. To set no organisation, the user needs to leave the autocomplete blank.

lfdebrux: Awesome, thanks for figuring this out 🙌

lfdebrux: I wonder if we should raise an issue upstream suggesting that the default `source` for dfe-autocomplete filters our null valued options?

app/controllers/users_controller.rb
```diff
@@ -30,10 +30,23 @@ def update
 private
 
   def user_params
-    params.require(:user).permit(:has_access, :role, :organisation_id)
+    params.require(:user).permit(:has_access, :role, :organisation_id).tap do |p|
```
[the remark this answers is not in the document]
lfdebrux: In which situations would `organisation_id_raw` be empty but `organisation_id` have a value?

app/frontend/entrypoints/application.js
thomasiles: yeah - I don't think it would take away anything from their current behaviour. It might just be that they never need empty ones and so it hasn't bothered them. I don't think the code is used that widely.

app/helpers/application_helper.rb
```diff
@@ -2,6 +2,8 @@
 
 # Rubocop ate my lunch
 module ApplicationHelper
+  include DfE::Autocomplete::ApplicationHelper
```
lfdebrux, line 5: It doesn't look like we're actually using any of the helpers in this module currently?

https://github.com/DFE-Digital/dfe-autocomplete/blob/main/app/helpers/dfe/autocomplete/application_helper.rb#L3

lfdebrux (dismissed):
Tested this locally and it works for me! 🙌 This will make life a lot easier when we have all the organisations 😁 

I have left a couple of questions, nothing super important though.

app/controllers/users_controller.rb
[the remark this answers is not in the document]
thomasiles: This happens when you clear the autocomplete box and submit. In that case organisation_id is returns the initial value - the users current organisation. This is discussed a bit in this [autocomplete issue](https://github.com/alphagov/accessible-autocomplete/issues/260). 

In summary, clearing the autocomplete text box doesn't change the value of the hidden select element, which is the value of organisation_id which is submitted.

lfdebrux (dismissed):

lfdebrux (approved):

PR #583, opened 2023-08-07
Add Exceptions To LogRage

danworth:
Add exception information to the LogRage logging output so its simpler to connect the exception with the original request and details (as per our other applications).

BAU, adding this as per out other apps.
Form-runner: https://github.com/alphagov/forms-runner/blob/main/config/application.rb#L76C9-L76C79
Forms-api: https://github.com/alphagov/forms-api/blob/main/config/application.rb#L65


alice-carr (approved):
👍 looks good

thomasiles (approved):
👍

aliuk2012 (approved):

PR #588, opened 2023-08-09
Remove references to PaaS

cadmiumcat:

Trello card: https://trello.com/c/DIr6jt2R/177-remove-paas-references-in-code-base
Removing references to PaaS from our code.

This is my naive attempt at doing it. I don't know if it breaks anything and also not 100% how to check if it does

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -53,8 +53,6 @@ class ApplicationController < ActionController::Base
   end
 
   def set_request_id
-    request.request_id = request_id
-
     # Pass the request id to the API to enable tracing
     if Rails.env.production?
       [Form, Page].each do |active_resource_model|
@@ -63,23 +61,9 @@ class ApplicationController < ActionController::Base
     end
   end
 
-  # PaaS uses a different header to pass on the request_id
-  # https://github.com/cloudfoundry/gorouter/issues/148 If PaaS header exists,
-  # use it, otherwise use standard header or generate new value
-  def request_id
-    vcap_request_id = request.env.fetch("HTTP_X_VCAP_REQUEST_ID", false)
-    if vcap_request_id
-      # This is a user input so take basic precautions by limiting chars to
-      # range and length
-      vcap_request_id.gsub(/[^\w\-@]/, "").first(255)
-    else
-      request.request_id
-    end
-  end
-
   # Because determining the clients real IP is hard, simply return the first
   # value of the x-forwarded_for, checking it's an IP. This will probably be
-  # enough for our basic monitoring in PaaS
+  # enough for our basic monitoring
   def user_ip(forwarded_for = "")
     first_ip_string = forwarded_for.split(",").first
     Regexp.union([Resolv::IPv4::Regex, Resolv::IPv6::Regex]).match(first_ip_string) && first_ip_string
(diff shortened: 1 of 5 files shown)


app/controllers/application_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class ApplicationController < ActionController::Base
   include Pundit::Authorization
+  delegate :request_id, to: :request
```
danworth, line 5: hmm rebased and aligned with main but I'm not sure what this change is, is it unintentional?

cadmiumcat: This is to replace [`def request_id`](https://github.com/alphagov/forms-admin/pull/588/files#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07L67-L76)
```
def request_id
    request.request_id
end
```
Rubocop told me to do it 👀 

But I am not 100% that it is the right way forward

lfdebrux: What was the cop ID?

lfdebrux: Further question, why do we need to define `request_id` at all?

DavidBiddle: Looks like [Rails/Delegate](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsdelegate)

cadmiumcat: @lfdebrux I was wondering too whether it needed to be defined at all - I wasn't sure if it was a Ruby thing.

cadmiumcat: I will push a temporary change without this line and see if it breaks anything 😁

cadmiumcat: ohh @lfdebrux, did you mean we could remove all references to `request_id`?

lfdebrux: I'm just wondering why we need to define `request_id` on the `ApplicationController` class. As far as I can see in our code, the only place where we use the `request_id` method is in `set_request_id`:

https://github.com/alphagov/forms-admin/blob/10fcf54a1bf3878ca5f5208629eacd21d81f9f6c/app/controllers/application_controller.rb#L56C33-L56C33

If we're setting `request.request_id` to itself that seems kind of redundant. So we should be able to get rid of that line also, right? Then what's the point of the `request_id` method?

danworth: Yeah I was thinking the same Laurence it seems unnecessary in the context of ApplicationController, but then, does any controller inheriting from ApplicationController (or somewhere else) use it, I didn't have time to go digging/grepping. Just reminded of the whole `@current_user` vs `current_user` etc...

danworth: @cadmiumcat 
>I will push a temporary change without this line and see if it breaks anything 😁

If you want to try removing it you'll also need to remove the reference to the `request_id` method it was providing in  [set_request_id](https://github.com/alphagov/forms-admin/pull/588/files#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07R54) which is why the tests are currently failing: e.g https://github.com/alphagov/forms-admin/actions/runs/5819019255/job/15776594974?pr=588#step:9:3213

cadmiumcat: Got it, brain has caught up now. I've removed both things, thanks all 🙌

alice-carr (approved):

PR #589, opened 2023-08-10
Move linting gems to development, test groups

lfdebrux:
We don't need these gems installed when running apps in production.

### Notes for reviewers


thomasiles (approved):
👍

jamie-o-wilkinson (approved):

PR #593, opened 2023-08-15
Refactor how we find an existing page

aliuk2012:
PagesController already has a memorised method called `page` but we don` seem to be using it across all the various pages controllers which inherit from PagesController.


DavidBiddle (approved):
Nice!

PR #596, opened 2023-08-16
Add markdown editor toolbar

DavidBiddle:

Trello card: [](https://trello.com/c/aJLuXFEQ/959-make-a-js-enhanced-markdown-formatting-toolbar)

This adds an editing toolbar to the additional guidance textarea as a preogressive enhacement.

The toolbar contains buttons to:
- add a level 2 or 3 heading
- add a link
- add a bulleted or numbered list

It also contains a small improvement to the CopyToClipboard test suite.

Things worth noting:
- I'm not using a component for the textarea here. This is because I think it makes sense to make a 'markdown editor' component which will include both this and the preview, so I think it makes sense to do that as part of [the preview enhancement Trello card](https://trello.com/c/hxlbS85W/1021-enhance-markdown-editing-with-tabs-component-and-javascript-preview) instead
- the labels for the buttons aren't translated - i think I've got a neat solution for passing the translations into the component, but I'll either do that as part of the preview work or as a separate cleanup card after that.


aliuk2012 (approved):
This is awesome, I think you should demo this final component to FE community and see if others could use it.

PR #598, opened 2023-08-18
Add markdown conversion on paste

DavidBiddle:

Trello card: https://trello.com/c/8qmNU2xa/960-allow-user-to-paste-html-into-a-markdown-field-and-have-it-converted-to-markdown

Copies the core code from https://github.com/alphagov/paste-html-to-govspeak and its tests, with some adaptations to make it more suited to the Forms codebase:
- removes the legacy paste function (we don't need it as we don't serve JS to IE, and there's a feature query so it should fail gracefully in case we do serve JS to a browser that doesn't support ClipboardData)
- changes the bullet style to use `* ` instead of `- ` since that's what we're using in our formatting help
- removes dependency on the [insert-text-at-cursor npm package](https://www.npmjs.com/package/insert-text-at-cursor)
- some light refactoring and reformatting


aliuk2012 (approved):
Looks good to me and working on my machine. It might be worth getting @thomasiles  to review this as well.

PR #602, opened 2023-08-24
Rename field to guidance_markdown

SamJamCul:
Renaming the page field additional_guidance_markdown to guidance_markdown for brevity. This change is also reflected in the API as well.

https://github.com/alphagov/forms-api/pull/316


package-lock.json
aliuk2012, line 1: Dont think this should have been committed

SamJamCul: ah so frustrating, I'll add it to my global gitignore

aliuk2012: i wouldn't ignore that file, there might be legit reasons for you to update and commit.
what version of node/npm are you using? sometimes these versions tweak the lock files

aliuk2012 (approved):

PR #604, opened 2023-08-25
Add live preview to guidance markdown editor

DavidBiddle:

Trello card: https://trello.com/c/cUoS52eU/986-enhance-markdown-editing-with-live-javascript-preview

Adds:
- a new endpoint on the guidance controller that takes markdown and returns the rendered HTML in its JSON response
- a new JS module which sends the contents of a textarea to this endpoint and renders the content in a preview area
  - the JS will submit to the endpoint when:
    - the page first loads
    - when the user types (we debounce it to 1 second - i.e. the request won't fire until the user has stopped typing for a second, so we don't accidentally DDoS ourselves by sending a request for every single keystroke)
    - when the user presses one of the toolbar buttons


app/controllers/pages/additional_guidance_controller.rb
```diff
@@ -47,6 +47,13 @@ def update
     end
   end
 
+  def render_preview
+    markdown = params[:guidance_markdown]
+    preview_html = markdown.present? ? GovukFormsMarkdown.render(markdown) : "<p>You have no content to preview</p>"
```
aliuk2012, line 52: Could you not maybe create a new instance the guidance_form which just sets `guidance_markdown` and then reuse `preview_html` method?

spec/requests/pages/additional_guidance_controller_spec.rb
```diff
@@ -195,4 +195,20 @@
...
   end
+
+  describe "#render_preview" do
+    let(:guidance_markdown) { "### Markdown" }
+
+    before do
+      post additional_guidance_render_preview_path(form_id: form.id), params: { guidance_markdown: }
+    end
+
+    it "returns a JSDON object containing the converted HTML" do
```
aliuk2012, line 206:
```suggestion
    it "returns a JSON object containing the converted HTML" do
```

app/controllers/pages/additional_guidance_controller.rb
DavidBiddle: @aliuk2012 Good suggestion, have updated!

app/frontend/javascript/ajax-markdown-preview/index.js
```diff
@@ -0,0 +1,123 @@
...
+  retryButton.classList.add('govuk-button', 'govuk-button--secondary')
+  retryButton.innerHTML = 'Retry preview'
+  addEventListeners(retryButton, manuallyTriggerMarkdownPreview)
+  store.target.appendChild(retryButton)
+}
+
+const triggerAjaxMarkdownPreview = async () => {
+  try {
+    const response = await window.fetch(
+      '/forms/16/pages/new/additional-guidance-preview',
```
aliuk2012, line 30: should this be dynamic?

DavidBiddle: Yes! Good spot 😅

DavidBiddle: Should be fixed now

aliuk2012 (approved):
Nicely done 👍

PR #605, opened 2023-08-26
Fix lint error in paste-to-html-markdown

chao-xian:

Addresses issues reported by the linter:

```
8:40  error  Expected '===' and instead saw '=='
32:3  error  'event' is never reassigned. Use 'const' instead
```

### DIFF
--- a/app/frontend/javascript/paste-html-to-markdown/index.js
+++ b/app/frontend/javascript/paste-html-to-markdown/index.js
@@ -5,7 +5,7 @@ import htmlToMarkdown from './html-to-markdown'
 const insertTextAtCursor = (field, contentToInsert) => {
   const selectionStart = field.selectionStart
   const selectionEnd = field.selectionEnd
-  if (selectionStart || selectionStart == '0') {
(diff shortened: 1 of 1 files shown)


aliuk2012:
@chao-xian great catch, 🤔  we need to check why linting CI isn't catching js linting issues.

Also noticed @DavidBiddle fixing that error here [`592822e` (#604)](https://github.com/alphagov/forms-admin/pull/604/commits/592822e555615c97469ce9e70d3e2acf4304bcff)

DavidBiddle:
I think the issue is that `npm run lint` runs `standard | snazzy & stylelint 'app/**/*.scss'`. Because we're running them in parallel (single `&`) the exit code is still 0 as long as one of the linting tasks passes. 

Fix would be to change `&` to `&&` for this task in `package.json`. @chao-xian feel free to do that in this PR or I'm happy to do it in a separate one

aliuk2012:
@chao-xian can we close this now?

chao-xian:
@aliuk2012 I completely forgot I even started this! I'll take a look - thanks for the prompt

PR #608, opened 2023-08-31
Refactor make live validation and task status

jamie-o-wilkinson:
Currently we have validation in forms-admin that prevents making a form live until certain conditions are met. However the api will happily make a form live if those conditions aren’t met. 

Prior to adding the validation to forms-api, I wanted to refactor the existing validation in forms-admin to make it more straightforward to move to forms-api, and allow forms-admin to consume the validation result from forms-api.

 I'll add comments to the diff, but in summary this means:
- refactoring the task status service to be user agnostic
- moving accessing task status service functionality via the form model (in preparation for that data to be in the form data itself)

Trello card: https://trello.com/c/xEyusUhB

### DIFF
--- a/app/forms/forms/make_live_form.rb
+++ b/app/forms/forms/make_live_form.rb
@@ -30,28 +30,28 @@ private
     return unless made_live?
     return if form.ready_for_live?
 
-    task_status_service = TaskStatusService.new(form:)
+    missing_sections = form.missing_sections
 
-    if task_status_service.missing_sections.include?(:missing_pages)
+    if missing_sections.include?(:missing_pages)
       errors.add(:confirm_make_live, :missing_pages)
       return false
     end
 
-    if task_status_service.missing_sections.include?(:missing_submission_email)
+    if missing_sections.include?(:missing_submission_email)
       errors.add(:confirm_make_live, :missing_submission_email)
       return false
     end
 
-    if task_status_service.missing_sections.include?(:missing_privacy_policy_url)
+    if missing_sections.include?(:missing_privacy_policy_url)
       errors.add(:confirm_make_live, :missing_privacy_policy_url)
     end
 
-    if task_status_service.missing_sections.include?(:missing_contact_details)
+    if missing_sections.include?(:missing_contact_details)
       errors.add(:confirm_make_live, :missing_contact_details)
       return false
     end
 
-    if task_status_service.missing_sections.include?(:missing_what_happens_next)
+    if missing_sections.include?(:missing_what_happens_next)
       errors.add(:confirm_make_live, :missing_what_happens_next)
       false
     end
--- a/app/forms/forms/make_live_form.rb
+++ b/app/forms/forms/make_live_form.rb
@@ -30,26 +30,28 @@ private
     return unless made_live?
     return if form.ready_for_live?
 
-    if form.missing_sections.include?(:missing_pages)
+    task_status_service = TaskStatusService.new(form:)
+
+    if task_status_service.missing_sections.include?(:missing_pages)
       errors.add(:confirm_make_live, :missing_pages)
       return false
     end
 
-    if form.missing_sections.include?(:missing_submission_email)
+    if task_status_service.missing_sections.include?(:missing_submission_email)
       errors.add(:confirm_make_live, :missing_submission_email)
       return false
     end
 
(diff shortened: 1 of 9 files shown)


app/forms/forms/make_live_form.rb
jamie-o-wilkinson, line 1: The behaviour here will have changed slightly, as previously the function would return as soon as any error (except `missing_privacy_policy_url`?) would be encountered. Now it will add all errors to the errors array. The `gov_uk_error_summary` component looks like it handles multiple errors so this should hopefully be fine.

app/models/form.rb
```diff
@@ -44,20 +42,15 @@ def append_page(page)
   end
 
   def ready_for_live?
-    @missing_sections = []
+    get_task_status_service.mandatory_tasks_completed?
```
jamie-o-wilkinson, line 45: The idea behind having these methods in the form model is that once forms-api does the validation, these methods will be replaced by equivalent data on the form object itself.

app/service/form_task_list_service.rb
```diff
@@ -107,8 +107,24 @@ def section_4_tasks
...
       task_name: I18n.t("forms.task_list_#{create_or_edit}.section_4.make_live"),
       path: @form.ready_for_live? ? make_live_path(@form.id) : "",
-      status: @task_list_statuses.make_live_status,
-      active: @task_list_statuses.mandatory_tasks_completed?,
+      status: @task_statuses[:make_live_status],
+      active: @form.ready_for_live?,
     }]
   end
+
+  def statuses_by_user
```
jamie-o-wilkinson, line 115: This was necessary to remove any `User` concerns from task status service, so that it deals exclusively with form data and can be moved into forms-api

app/forms/forms/make_live_form.rb
lfdebrux: Would you mind adding a screenshot of what this looks like?

app/models/form.rb
```diff
@@ -101,4 +90,8 @@ def page_number(page)
   def has_routing_conditions
     pages.filter { |p| p.conditions.any? }.any?
   end
+
+  def get_task_status_service
```
lfdebrux, line 94: I think this should just be called `task_status_service`?

app/models/form.rb
```diff
@@ -44,20 +42,7 @@ def append_page(page)
...
-    @missing_sections << :missing_privacy_policy_url unless task_list_statuses.privacy_policy_status == :completed
-    @missing_sections << :missing_contact_details unless task_list_statuses.support_contact_details_status == :completed
-    @missing_sections << :missing_what_happens_next unless task_list_statuses.what_happens_next_status == :completed
-
-    if @missing_sections.any?
-      false
-    else
-      true
-    end
+    TaskStatusService.new(form: self).mandatory_tasks_completed?
```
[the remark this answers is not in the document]
lfdebrux: It's probably not worth it though 😅

app/forms/forms/make_live_form.rb
jamie-o-wilkinson: image

@lfdebrux even with multiple errors it just shows the first which would have been the original behaviour. I'm a little surprised it doesn't show a list with multiple errors...I'll keep looking

app/models/form.rb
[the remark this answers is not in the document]
jamie-o-wilkinson: Good point, have done this 👍

app/forms/forms/make_live_form.rb
jamie-o-wilkinson: @lfdebrux looking at the [docs for `govuk_error_summary` ](https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder%2FBuilder:govuk_error_summary) it only shows the first error for each attribute, so this is working as expected. If we had multiple questions in this page, each with errors, it would list all of them, e.g.: 

image

lfdebrux: Hmm, I guess that makes sense in most circumstances, it's kind of annoying because in our case we're showing multiple errors but they're not interrelated, they just are sort of gated by the question... but as we discussed in tech weekly this morning, it's rare that people will see those errors, so I'd say keep the existing behaviour.

lfdebrux (approved):
LGTM!

PR #615, opened 2023-09-05
Remove unnecessary console log

DavidBiddle:
Removes a console log I left in while debugging some tests.


aliuk2012 (approved):
Should our linter have caught this?

thomasiles (approved):
seems legit

PR #616, opened 2023-09-05
add persistent timout of 75 to puma

alice-carr:

Trello card: [](https://trello.com/c/YyrG1YHG/114-investigate-502-between-alb-and-forms-api)

This PR adds a persistent_time of 75 to the puma config so that it is higher than the ALB idle time as this was causing 502s


cadmiumcat (approved):
⏲️

chao-xian (approved):
LGTM

PR #618, opened 2023-09-05
Move page related form objects to pages directory

aliuk2012:

I noticed that our FormObjects had a mixture of objects relating to Form model and Page model in the same directory. Sorry about the large number of file changes(mostly just moving files around), I did run lint & rspec before each commit so it should be easier to review looking at each individual commit.

Trello card: 


SamJamCul (approved):
All looking good

DavidBiddle (approved):
Nice!

PR #619, opened 2023-09-06
Add length validations to question text, page heading and hint text

aliuk2012:

Trello card: https://trello.com/c/MAokcg4Y/1001-add-size-limit-to-question-text-pageheading-hint-text

This should match the validations that we have in forms-api and as a result the validations in forms-api should only be a safety net for us should any validations form forms-admin fail to catch invalid data


DavidBiddle (approved):
Nice one 🎉

PR #607, opened 2023-08-30
Markdown editor component

DavidBiddle:
### What problem does this pull request solve?

Trello card: https://trello.com/c/hxlbS85W/987-enhance-markdown-editing-with-tabs-component

This PR:
- Moves the whole markdown editor into a component
- Moves the preview into tabs (on larger screen widths)
- Does some general cleaning up after the other markdown JS PRs

### Things to consider when reviewing

- Does it work when run on your machine?
- Is it clear what the code is doing?
- Do the commit messages explain why the changes were made?
- Are there all the unit tests needed?
- Has all relevant documentation been updated?

PR #620, opened 2023-09-06
Fix markdown ID issue

DavidBiddle:

Trello card: 

The markdown compenentisation work (#607) introduced a bug, where if the page has a validation error the error summary message does not link to the field correctly.

The reason for this is that we created a custom ID for the guidance field, but the error summary component expects the ID to be in the govuk_form_builder defined format.

Resolved in this PR by using the govuk_form_builder's govuk_field_id helper function to generate the ID in our markdown component, meaning the field's ID will always be in sync with the error summary component's expectations.


aliuk2012 (approved):
Nice one

PR #624, opened 2023-09-08
Add logo for Auth0 widget to public dir

lfdebrux:

Trello card: https://trello.com/c/CLoA4BfF

For trust and branding purposes we want the GOV.UK crown to appear on our Auth0 login screens. Auth0 expects us to provide the logo via a URL. This commit adds a copy of [govuk-mask-icon.svg] from GOV.UK Frontend v4 to our `public` folder, so it will be served from the server root.

[govuk-mask-icon]: https://github.com/alphagov/govuk-frontend/blob/v4.7.0/src/govuk/assets/images/govuk-mask-icon.svg

### Rejected alternatives

We could tell Auth0 to use the copy of this file already in our `assets` folder, but that is fingerprinted by the Vite asset pipeline with a content hash [[1]], so we could accidentally break our logo when GOV.UK Frontend updates the crown logo by doing that.

[1]: https://vitejs.dev/guide/assets.html

I thought about whether we could tell Ruby to serve the existing asset file at another path, but this seemed like a bit of an inefficient and complex process, and would also be an impediment if we want to change the icon in future to something more custom. I also rejected the idea of writing a Vite plugin to copy the asset to the public folder for the same reason.

I did experiment with encoding the SVG as a [data URI], so saving users an extra network request, but found that the Auth0 dashboard didn't like that; they expect an absolute URL. Also the base64 encoded SVG was over 4 KiB, which seems a bit cheeky.

[data URI]: https://en.wikipedia.org/wiki/Data_URI_scheme


thomasiles (approved):
nice write up

PR #627, opened 2023-09-12
Allow page options service to handle existing pages that dont have guidance

aliuk2012:

This fixes a problem where there is an existing live form which the user is trying to view the forms pages.

It throws an undefined method error because the snapshot of the live form page did not contain "page_heading" or "guidance_markdown"

fixes
- https://govuk-forms.sentry.io/issues/4458650198/?project=6576190
- https://govuk-forms.sentry.io/issues/4469134101/?project=6576190

Trello card:  https://trello.com/c/mjqZND5D/1019-viewing-an-existing-live-form-pages-in-forms-admin-returns-500-error

### DIFF
--- a/app/views/pages/_form.html.erb
+++ b/app/views/pages/_form.html.erb
@@ -13,9 +13,8 @@
 
   <% if FeatureService.enabled? :detailed_guidance_enabled %>
     <h2 class="govuk-heading-m"><%= t("guidance.guidance") %></h2>
-    <% if page_object.page_heading.present? && page_object.guidance_markdown.present? %>
+    <% if (page_object.respond_to?(:page_heading) && page_object.page_heading.present?) && (page_object.respond_to?(:markdown_guidance) && page_object.guidance_markdown.present?) %>
       <%= govuk_summary_list(**PageSummaryData::GuidanceService.call(form: form_object, page: page_object).build_data) %>
-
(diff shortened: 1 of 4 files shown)


app/views/pages/_form.html.erb
```diff
@@ -13,9 +13,8 @@
 
   <% if FeatureService.enabled? :detailed_guidance_enabled %>
     <h2 class="govuk-heading-m"><%= t("guidance.guidance") %></h2>
-    <% if page_object.page_heading.present? && page_object.guidance_markdown.present? %>
+    <% if (page_object.respond_to?(:page_heading) && page_object.page_heading.present?) && (page_object.respond_to?(:page_heading) && page_object&.guidance_markdown&.present?) %>
```
lfdebrux, line 16: Should this be `guidance_markdown` the second time? Also, I think the extra ampersands here are unnecessary?

```suggestion
    <% if (page_object.respond_to?(:page_heading) && page_object.page_heading.present?) && (page_object.respond_to?(:guidance_markdown) && page_object.guidance_markdown.present?) %>
```

aliuk2012: Good spot, thanks @lfdebrux.

lfdebrux (approved):
Works for me, thanks for fixing this!

PR #628, opened 2023-09-12
Revert "allow add/edit page to handle existing pages that dont have guidance"

aliuk2012:

This reverts commit c6fe07544716e6ff15625b9228be70bdf849ab73.

This should never have been implemented as the c6fe075 commit itself said its unlikely to happen. The Page model resource itself will always have guidance attributes unless the columns were dropped by then more things would break and need to refactor.

Trello card: https://trello.com/c/mjqZND5D/1019-viewing-an-existing-live-form-pages-in-forms-admin-returns-500-error


jamie-o-wilkinson (approved):
Looks ok to me 👍

PR #630, opened 2023-09-13
Add cancel link to "Add guidance" page

aliuk2012:

Adding this link to allow users to "leave" the page without saving any changes they have made. It works exactly like the "back link" already on the page

Trello card:  https://trello.com/c/48YVQpfI/1024-add-cancel-link-to-add-guidance-page


jamie-o-wilkinson (approved):
Looks good to me 👍

PR #631, opened 2023-09-14
Use ActiveRecord Session Store instead of cookies or Redis

aliuk2012:
## What problem does this pull request solve?

We will switch out from using Rails default cookie based session storage, which has 4k data limit and use ActiveRecord db storage instead where we would no longer have this restriction.

As this would affect existing users, I plan on deploying this first thing tomorrow morning.

Trello card: https://trello.com/c/o1oPv62S/1052-switch-cookie-storage-to-activerecord-session-store-gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -118,3 +118,5 @@ group :test do
   # axe-core for running automated accessibility checks
   gem "axe-core-rspec"
 end
+
+gem "activerecord-session_store", "~> 2.1"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -78,6 +78,13 @@ GEM
     activerecord (7.0.8)
       activemodel (= 7.0.8)
       activesupport (= 7.0.8)
+    activerecord-session_store (2.1.0)
+      actionpack (>= 6.1)
+      activerecord (>= 6.1)
+      cgi (>= 0.3.6)
+      multi_json (~> 1.11, >= 1.11.2)
+      rack (>= 2.0.8, < 4)
+      railties (>= 6.1)
     activeresource (6.0.0)
       activemodel (>= 6.0)
       activemodel-serializers-xml (~> 1.0)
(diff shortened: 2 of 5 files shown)


chao-xian:
From our convo earlier:

> I think the best thing is to only focus on update “pages” part of a form which would involve:
> * creating a database tables for each one of these https://github.com/alphagov/forms-admin/tree/main/app/forms/pages
> * Update those classes to be ActiveRecord
> * ensure there is some way to  ring fence specific user flows (we would have access to user_id)
> * need to think what happens if a user leaves the add/edit page journey and how they would return
> * package up all the data from the various models and build a final “page” to forward on to forms-api

chao-xian:
Changes look good but I'd like to check that e2e are ok before approving

chao-xian (approved):

thomasiles:
This looks good but I have a couple of questions about how it works and if there are any knock-on effects it makes sense to think about now.

From the [gem docs](https://github.com/rails/activerecord-session_store) it looks like sessions aren't removed from the database by default.

I guess we need to run `bin/rake db:sessions:trim` every so often against all our environments which removes sessions older than 30 days by default. I think we can use the tool laurence made to do that, we just need to remember it :)

Does the gem handle session expiry? It mentions it in the readme but I'm not sure if it's something we need to add. If it's by default, what is standard session expiry time?

This might be important to document as the sessions will be storing auth details too.

It might be worth adding something to Cat's list of risks (I forget the name of it). I think developers would be able to copy access the session of all users and maybe impersonate them. They might have been able to do that before, I'm not sure. As sessions were in browser cookies, it was less of an issue.

The last question is about the increase in DB calls this will create. I don't think it will be awful as we don't have that many users. I guess there will be two DB calls now for every request with a session. Do we need to alert infra or monitor anything to see how that pans out?

aliuk2012:
Closing this as @chao-xian @thomasiles @lfdebrux  and myself had a chat about the very good points that @thomasiles raised in the previous comment. There didn't seem to be be a desperate need for this. It basically solved the 4k data limit of using cookie but ultimately we dont want pages data in sessions. Sessions should be reserved for login tokens and smaller amount of temporary information. instead we will focus on migrating forms objects over to activerecord

chao-xian:
Thanks @aliuk2012

PR #632, opened 2023-09-14
Sort users page by organisation, access, role, and name

lfdebrux:

Trello card: https://trello.com/c/3tyyHhoK

Currently the users table that lists all users is not ordered. That’s ok when there aren’t many users but it would probably make sense to order them by organisation, name, user type or a combination of these - so that it’s easier to find a specific user or get a view of a number of users in a group.

<details><summary>Screenshot</summary>

Screenshot of table of users at locahost:3000/users

</details>

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -11,7 +11,12 @@ class UsersController < ApplicationController
 
     roles = User.roles.keys
     @users = policy_scope(User).sort_by do |user|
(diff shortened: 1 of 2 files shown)


jamie-o-wilkinson (dismissed):
Looks good to me. I'm looking forward to much a neater users page 👏

hannahkc:
Looks good to me,

jamie-o-wilkinson (approved):

PR #635, opened 2023-09-15
Use setup-node action to install Node.js and cache npm downloads

lfdebrux:

Currently `npm ci` can take over 2 minutes in GitHub Actions, this commit uses the actions/setup-node action to cache the npm downloads to try and speed things up.

We also use the action to install the same version of Node as in the `.nvmrc` file, because we might as well.

This PR also makes the `.test.yaml` workflow file more consistent with the same workflow file in our other app repos.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,7 @@
 # run tests and linters.
 
 name: "Tests"
+
 on:
   push:
     branches: [main]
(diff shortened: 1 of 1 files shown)


jamie-o-wilkinson (dismissed):
Good idea 😄

jamie-o-wilkinson (approved):

PR #636, opened 2023-09-18
Use api task validation

jamie-o-wilkinson:
Currently we have validation in forms-admin that prevents making a form live until certain conditions are met. However the api will happily make a form live if those conditions aren’t met.

This PR consumes the task status data on the form object that has been moved from forms-admin to forms-api (in this PR: https://github.com/alphagov/forms-api/pull/332). It also replaces the existing `TaskStatusService` with and `EmailTaskStatusService`, as forms-api deals with everything `TaskStatusService` used to except for the submission email status, which depends on data in the form_submission_emails table in the forms-admin database.

It would be great to hear other people's opinions on whether this refactor (in effect splitting the validation across forms-api and forms-admin) is preferable to duplicating the validation and removing the need to pass data back via the form object.

Trello card: https://trello.com/c/xEyusUhB


app/models/form.rb
jamie-o-wilkinson, line 1: Here we have to combine data from forms-api with the submission email stuff from the refactored email task status service. This feels a bit messy to me and had I foreseen this requirement perhaps would have just duplicated the logic between admin and api to achieve the make live validation in the api. 

We could still do this very easily so I would value any thoughts on if this solution or duplication of the task status service is preferable.

app/service/email_task_status_service.rb
jamie-o-wilkinson, line 1: This follows exactly the same structure as the task status service in forms-api, so that if we refactor to merge the two in future it is straightforward.

thomasiles: I think this makes sense and is a pragmatic solution. I guess we also have the option of moving the submission email stuff into the API in the future.

app/models/form.rb
thomasiles: I think it works as we are moving in the right direction

thomasiles (dismissed):
good work 🏅

app/service/email_task_status_service.rb
jamie-o-wilkinson: Yeah that was my thought when doing this. That or vice-versa if we move form data into forms-admin and make forms-api concerned only with serving live forms.

thomasiles (approved):

PR #640, opened 2023-09-19
Add edit guidance link

DavidBiddle:

Trello card: https://trello.com/c/70ocEwDu/1044-add-edit-guidance-text-link-to-add-guidance-page-when-js-is-disabled

Adds a anchor link to the guidance page which will take the user back to the guidance field. This is useful if there's no JS or if the user's on a narrow screen, as in those cases the preview and the edit field are further apart than on wider screens with JS enabled. The link won't display if the tabs component is active.

Also adds a few view tests for the guidance view as there weren't any.

### DIFF
--- a/spec/views/pages/guidance.html.erb_spec.rb
+++ b/spec/views/pages/guidance.html.erb_spec.rb
@@ -1,0 +1,54 @@
+require "rails_helper"
+
+describe "pages/guidance.html.erb", type: :view do
+  let(:form) { build :form, id: 1 }
+  let(:page) { OpenStruct.new(conditions: [], answer_type: "number") }
+  let(:guidance_form) { Pages::GuidanceForm.new }
+  let(:preview_html) { I18n.t("guidance.no_guidance_added_html") }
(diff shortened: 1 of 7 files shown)


spec/views/pages/guidance.html.erb_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  it "contains the question number" do
+    expect(rendered).to have_content("Question #{question_number}")
+  end
+
+  it "has the correct heading and caption" do
+    expect(rendered).to have_selector("h1", text: "Question #{question_number}")
+    expect(rendered).to have_selector("h1", text: I18n.t("guidance.add_guidance"))
+  end
+
+  it "contains a form which submits to @type_of_answer_path" do
```
aliuk2012, line 51:
```suggestion
  it "contains a form which submits to @guidance_form_path" do
```

DavidBiddle: @aliuk2012 Good spot, have fixed!

aliuk2012 (approved):

PR #641, opened 2023-09-19
Update en.yml

C-Harry:
Updated all the hint text hints that display on 'Edit question' pages. Have specified that 'You can add a short hint to help people answer the question' to help people understand when to use hint text versus longer guidance text.

Also updated the guidance text on 'Edit question' pages to make this one sentence and specify the types of Markdown formatting that are allowed.

Trello card: [<https://trello.com/c/6X7sxxAg/1011-guidance-add-brief-clarification-to-hint-text-and-guidance-instructions>]

- Have I changed all instances of hint text hints?
- Any typos or grammatical errors?

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -188,7 +188,7 @@ en:
   guidance:
     add_guidance: Add guidance
     guidance: Guidance
-    instructions: Only add guidance if you need to give a longer explanation of how to answer the question, or to format your content. For example, you can use paragraphs, links or lists.
+    instructions: Only add guidance if you need to give a longer explanation of how to answer the question, or to format your content with paragraphs, headings, lists or links.
     markdown_editor:
       preview:
         description: Below is a preview of how your guidance content will be shown to the person completing your form.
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -216,21 +216,21 @@ en:
...
-          single_line: Use hint text to help people answer the question. For example, you could say what format the answer should be in or where to find it.
+          address: You can add a short hint to help people answer the question. For example, ‘We’ll send your licence to this address.’
+          checkbox: You can add a short hint to help people answer the question. For a question where people can select more than one answer you might want to use ‘Select all that apply’.
+          date: You can add a short hint to help people answer the question. For a date question you could use ‘For example, 27 3 2007’.
+          date_of_birth: You can add a short hint to help people answer the question. For a date of birth question you could use ‘For example, 27 3 1998’.
+          default: You can add a short hint to help people answer the question. For example, to explain the format the answer should be in, or where to find the information you’ve asked for.
+          email: You could add a short hint to tell people how you'll use their email address. For example, ‘We’ll only use your email address to contact you about your application.’
```
hannahkc, line 224: I know it's not really related to the change your making but I wonder if we should follow the same pattern we used for the address hint for this one?

```suggestion
          email: You can add a short hint to help people answer the question. For example, you could tell people how you'll use their email address.
```

C-Harry: Could do, yes, but I was thinking that this example text is quite useful and people might want to copy and paste it (more so than the address example we had).

hannahkc: Yeah possibly true. I think I'm more concerned that this hint is quite suggestive of one way to use the hint text and not as broad as the others that start with "You can add a short hint to help people answer the question". This was an existing issue with the hint though and not related to this update - so fine to stick with what you have. :)

C-Harry: Yeah, I agree it's suggestive, but it feels like it's usefully suggestive... I've made a note in the hint text doc that this is something to think about though.

aliuk2012 (approved):

DavidBiddle (approved):
Nice!

PR #643, opened 2023-09-20
remove obsolete Form methods

aliuk2012:

I had a quick review of our Form model and found a couple of obsolete methods that can be removed. 


lfdebrux (approved):
LGTM!

PR #644, opened 2023-09-21
Fix issue with pages not being able to read/update from sessions

aliuk2012:
Recently, we did a refactor to convert all our session hash strings to symbols for consistency reasons and for better performance. https://github.com/alphagov/forms-admin/pull/637

A user reported issues with adding/editing guidance. On closer inspection, this was because the session keys were symbols but the main `load_from session` method was being passed in keys as strings rather than symbols.

` self.load("#{key}": session.dig(:page, key) || send(key.to_sym))` This meant that session.dig was returning nil and therefore it was falling back to whatever was last saved in forms-api.

The problem was not identify for a couple of reasons.

- The specs matched the calls being made in the controller so were all passing in strings.
- There was an issue with the spec that was written to check `load_from_session` returns the value from the session, had mocked out a session that was using strings as its hash key and so it was returning a false positive.
- There was also a missing spec to test what would happen if `load_from_session` was passed an array of strings

Trello card: 

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -45,7 +45,7 @@ class Page < ActiveResource::Base
 
   def load_from_session(session, keys)
     keys.each do |key|
-      self.load("#{key}": session.dig(:page, key) || send(key.to_sym))
+      self.load("#{key}": session.dig(:page, key.to_sym) || send(key.to_sym))
     end
     self
   end
--- a/app/controllers/pages/address_settings_controller.rb
+++ b/app/controllers/pages/address_settings_controller.rb
@@ -21,7 +21,7 @@ class Pages::AddressSettingsController < PagesController
   end
 
   def edit
-    page.load_from_session(session, %w[answer_type answer_settings])
+    page.load_from_session(session, %i[answer_type answer_settings])
     input_type = @page&.answer_settings&.input_type
     uk_address = input_type&.uk_address
     international_address = input_type&.international_address
--- a/app/controllers/pages/date_settings_controller.rb
+++ b/app/controllers/pages/date_settings_controller.rb
@@ -20,7 +20,7 @@ class Pages::DateSettingsController < PagesController
   end
 
   def edit
-    page.load_from_session(session, %w[answer_type answer_settings])
+    page.load_from_session(session, %i[answer_type answer_settings])
     input_type = @page&.answer_settings&.input_type
     @date_settings_form = Pages::DateSettingsForm.new(input_type:, page: @page)
     @date_settings_path = date_settings_update_path(@form)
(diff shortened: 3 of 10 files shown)


app/models/page.rb
```diff
@@ -45,7 +45,7 @@ def submit
 
   def load_from_session(session, keys)
     keys.each do |key|
-      self.load("#{key}": session.dig(:page, key) || send(key.to_sym))
+      self.load("#{key}": session.dig(:page, key.to_sym) || send(key.to_sym))
```
aliuk2012, line 48: 🤔  I was in two minds about creating a new local variable just above the the call to `each` which would store a version of session but with its keys symbolized like https://github.com/alphagov/forms-admin/blob/main/app/controllers/pages_controller.rb#L122 so if the session passed in does have a mixture of strings/symbols for some odd reason then the values can still be found.

Any thoughts?

lfdebrux: What about using `ActiveSupport::HashWithIndifferentAccess`? 

https://api.rubyonrails.org/classes/ActiveSupport/HashWithIndifferentAccess.html

lfdebrux:
Could we add a feature spec for adding detailed guidance? Would that have caught this issue ealier?

app/models/page.rb
aliuk2012: Thats a good option too. Leave this with me

lfdebrux (approved):
I was able to reproduce the original bug with main branch and verify that this change fixes it.

app/models/page.rb
```diff
@@ -45,7 +45,7 @@ def submit
 
   def load_from_session(session, keys)
     keys.each do |key|
-      self.load("#{key}": session.dig(:page, key.to_sym) || send(key.to_sym))
+      self.load("#{key}": session.with_indifferent_access.dig(:page, key.to_sym) || send(key.to_sym))
```
lfdebrux, line 48: When testing locally I get the following error:

```
undefined method `with_indifferent_access' for #<ActionDispatch::Request::Session:0x000000011d384fc8 ...
```

aliuk2012:
"Could we add a feature spec for adding detailed guidance? Would that have caught this issue earlier?" 
I think this issue could have been picked up sooner with better unit tests as there were a couple of issues with them and it was a wider issue than just detailed guidance.

app/models/page.rb
aliuk2012: I'll fix this issue in another commit. lets get the original fix deployed

lfdebrux (approved):

app/models/page.rb
aliuk2012: https://github.com/alphagov/forms-admin/pull/645

PR #646, opened 2023-09-21
Add form for user to request editor upgrade

thomasiles:
# Add a form for users to request an upgrade from trial user to editor

Trello card: https://trello.com/c/cgDgv0Dw/1036-add-initial-account-upgrade-page

For early access, trial users need to be able to request an editor account from the engagements team.

This requirement is met with a new page `/upgrade`. It contains some text and a form with a confirmation box, which when filled in shows the user a screen `/upgrade/requested` telling the user that they will receive an email with further instructions.

These pages are not linked to by any pages yet and so should be safe to deploy to production.

## Notes

We don't store that a user has requested an upgrade and there is nothing to stop users accessing /upgrade/requested directly. Submitting the form multiple times would result in multiple emails.

I think this makes sense as an MVP and it matches the designs. We haven't specified what happens if a user doesn't receive the email etc and these things would need extra logic. It should be easy to add in the future if we decide it's something we want to add.

This commit does send the email but stubs out a simple service class to add this in.

I've made bits of text up for the page titles and error message as they weren't on the design. I'm happy to make adjustments to the content in later commits.

There also aren't many tests beyond the feature tests. I think this is ok because there isn't much happening here - we can add further tests when we flesh out the service class. I could be convinced otherwise if there are strong opinions though.

## Screenshots

The upgrade page:
- url: /upgrade
- page title: Request upgrade to editor account
image

If the checkbox is not ticked when submitting the form:
- error message: You must meet the requirements
image

The confirmation page:
- url: /upgrade/requested
image


config/locales/en.yml
```diff
@@ -675,6 +685,32 @@ en:
...
+
+        <p>If you’re from a central government organisation you might be able to upgrade to an editor account early. You’ll need to meet some requirements.</p>
+
+        <p>You may be able to upgrade to an editor account if:</p>
+
+        <ul class="govuk-list govuk-list--bullet">
+          <li>you need to make an online form to be published on GOV.UK</li>
+          <li>you believe your form can be created with the features already available — <a href="https://www.forms.service.gov.uk/features" class="govuk-link">read about the features available already</a> and try to make the form to check this</li>
+          <li>your form is expected to get between 250 and 10,000 submissions a year</li>
+          <li>you're happy to be asked to provide feedback and take part in user research</li>
```
DavidBiddle, line 701: Tiny thing but we can use curly quotes here:
```suggestion
          <li>you’re happy to be asked to provide feedback and take part in user research</li>
```

config/locales/en.yml
```diff
@@ -675,6 +685,32 @@ en:
...
+        </ul>
+
+        <p>If you believe you meet these requirements, we'll get in touch with you to discuss the next steps.</p>
```
DavidBiddle, line 704:
```suggestion
        <p>If you believe you meet these requirements, we’ll get in touch with you to discuss the next steps.</p>
```

config/locales/en.yml
```diff
@@ -675,6 +685,32 @@ en:
...
+      heading: Requirements to upgrade to an editor account
+      title: Request upgrade to editor account
+    show:
+      body_html: |
+        <h2 class="govuk-heading-l">What happens next</h2>
+        <p class="govuk-body">Check your email and reply with the information requested.</p>
+        <p class="govuk-body">We'll then get in contact with you to discuss the next steps. It may take up to 2 weeks.</p>
+      panel_title: We've sent you an email to request more information
```
DavidBiddle, line 712:
```suggestion
        <p class="govuk-body">We’ll then get in contact with you to discuss the next steps. It may take up to 2 weeks.</p>
      panel_title: We’ve sent you an email to request more information
```

spec/features/users/user_upgrade_request_spec.rb
lfdebrux, line 1: Yay feature tests! 🙌

spec/features/users/user_upgrade_request_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+private
+
+  def visit_upgrade_page
+    visit new_user_upgrade_request_path
+    expect(page).to have_text I18n.t("user_upgrade_request.new.heading")
+  end
+
+  def submit_request
+    click_button I18n.t('continue')
```
DavidBiddle, line 27: Linting fix:
```suggestion
    click_button I18n.t("continue")
```

DavidBiddle:
Looks good overall, have just raised a small curly quote suggestion. Another very minor thing but think the PR message should say:
> This commit **doesn't** send the email but stubs out a simple service class to add this in.

Instead of:
> This commit does send the email but stubs out a simple service class to add this in.

lfdebrux (commented):
A couple of questions, nothing super important though, I'm mainly asking out of my own curiosity so feel free to ignore 😅

app/views/user_upgrade_requests/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<% set_page_title(title_with_error_prefix(t("user_upgrade_request.new.title"), @user_upgrade_request.errors.any?)) %>
+<% content_for :back_link, govuk_back_link_to(root_path, t("back_link.forms")) %>
```
lfdebrux, line 2: If we're going to be linking to this page from everywhere the banner is, this should probably be a plain back link?

lfdebrux: Any reason we can't use the back link helper in the govuk-components gem?

https://govuk-components.netlify.app/components/back-link/

app/controllers/user_upgrade_requests_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+    @user_upgrade_request = UserUpgradeRequest.new
+  end
+
+  def show; end
+
+  def create
+    @user_upgrade_request = UserUpgradeRequest.new(user_upgrade_request_params)
+
+    if @user_upgrade_request.valid?
+      UserUpgradeRequestService.new(current_user).request_upgrade
```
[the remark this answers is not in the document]
thomasiles: I went down this route because the form is so simple and I think the code in the service is likely to be triggered in other ways in the future. Maybe it's a YAGNI thing, but while we don't have a requirement for it yet but it doesn't seem too far fetched that the engagement team might want to trigger an email from the /users screen for example.

I guess either way works, as long as the functionality is kept out of the controller and testable.

I nearly used a form object, like we do elsewhere to keep things consistent. But the form is really so small that it feels hard to justify it. The checkbox isn't really used anywhere else, it's just a guard to ensure the the user has ticked it. Because of this I was tempted to do the form totally in the controller. We only need to make sure the checkbox is true, which comes down to checking a param. The only which stopped me was that I needed the error state, which winds up being more lines of code because I don't get to use the form helpers.

config/routes.rb
```diff
@@ -113,6 +113,8 @@
   end
 
   resources :users, only: %i[index edit update]
+  resource :user_upgrade_request, only: %i[new show create], path_names: { new: "/" }, path: "/upgrade"
```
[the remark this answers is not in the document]
thomasiles: Yeah, I've tied myself in knots around this one.

When we started this project I was much more used to writing small javascript apps in express, where i really like explicit routing and handlers - it makes it clear exactly whats mounted on URLs. You can see that reflected in our routes file.

It's been pointed out to me that this isn't how things are done in Rails 😄 

I've ended up here because I wanted to make things more "railsy". I think the railsy way is to leverage "resources", in routing, controllers and models, even for things which aren't saved in the DB like this case.

I like the idea of only using the standard rails action, not adding custom resource actions if its possible to avoid it and using extra routes for custom URLS. This makes reading the routes file useful for people to work out whats available and what it's likely to do.

This has got messy because we're not storing the user_upgrade_request so the confirmation page (under #show) is always available. When the user hits /new, clicks continue without checking the met_requirements box, they are   redirected to the show route by the #create handler. Refreshing that, gives the confirmation screen.

I think you are right, it's probably better to get rid of show and create a custom resource action for the confirmation page. It's clearer than what I've ended up with.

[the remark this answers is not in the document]
thomasiles: Yeah, I've tied myself in knots around this one.

When we started this project I was much more used to writing small javascript apps in express, where i really like explicit routing and handlers - it makes it clear exactly whats mounted on URLs. You can see that reflected in our routes file.

It's been pointed out to me that this isn't how things are done in Rails 😄 

I've ended up here because I wanted to make things more "railsy". I think the railsy way is to leverage "resources", in routing, controllers and models, even for things which aren't saved in the DB like this case.

I like the idea of only using the standard rails action, not adding custom resource actions if its possible to avoid it and using extra routes for custom URLS. This makes reading the routes file useful for people to work out whats available and what it's likely to do.

This has got messy because we're not storing the user_upgrade_request so the confirmation page (under #show) is always available. When the user hits /new, clicks continue without checking the met_requirements box, they are   redirected to the show route by the #create handler. Refreshing that, gives the confirmation screen.

I think you are right, it's probably better to get rid of show and create a custom resource action for the confirmation page. It's clearer than what I've ended up with.

app/controllers/user_upgrade_requests_controller.rb
```diff
@@ -0,0 +1,23 @@
+class UserUpgradeRequestsController < ApplicationController
+  def new
+    @user_upgrade_request = UserUpgradeRequest.new
+  end
+
+  def show; end
```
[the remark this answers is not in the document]
thomasiles: this is answered below, where I decided to change it.

app/views/user_upgrade_requests/new.html.erb
thomasiles: We are using the component to display the back link, it's the url needed by the component which our own application helper was intended to wrap, but I guess we have never needed.

I think above you are suggesting that the link should return to the previous page, either tasklist or forms index, which I think requires using the referer header, javascript, or setting something in the user's cookie (unless there is a better way).

That's the bit missing from our current helper and I don't think it's provided by the govuk-components gem.

lfdebrux: Ah yes, I see what you're talking about needing to fix in the helper now, the `back_link_url` method in `govuk_back_link_to` has no implementation 🤔 

I wonder if a compromise might be using a breadcrumb to the root page instead might be preferable?

Compromise in the sense that everyone is unhappy...

app/controllers/user_upgrade_requests_controller.rb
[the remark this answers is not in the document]
lfdebrux: Yes, I also tried initially to get away without a form object and also quickly found that the errors are difficult to deal with without it!

app/views/user_upgrade_requests/new.html.erb
lfdebrux: Ah, interesting! Thanks for explaining all that.

The main scenario I can think of where using the referrer might be an issue is if a search engine indexes the page and the suggests it to someone... but that shouldn't be an issue here because the page is behind a login, right?

The dfe `back_link_url` helper has a certain neatness to it, I like how it checks that the referrer is from the same host. It's a shame that the Rails [:back URL logic](https://github.com/rails/rails/blob/fc734f28e65ef8829a1a939ee6702c1f349a1d5a/actionview/lib/action_view/helpers/url_helper.rb#L52) doesn't do the same. But it is a pain that we have to then include some extra code in all our apps we want to use it.

I think if a pentest labels this as an open redirect and we're confident that there's no issue we can push back on it.

lfdebrux (approved):

hannahkc:
Content here is fine for now. I need to iterate all of it following user research, but that will take me a few days and I should be able to manage that myself once the files are there for me to edit. Thanks Tom!

PR #651, opened 2023-09-26
Update preview area with border

DavidBiddle:

Trello card: https://trello.com/c/ftaYvI4q/1045-change-preview-on-add-guidance-page-when-js-is-disabled

This PR contains changes to make it easier to tell which parts of the preview UI are previewed content (i.e the user's content) and which are normal parts of the add guidance page.

We make the distinction visually by adding a 2px border around the preview area. This change will be visible on narrow screens, or when javascript is not enabled. On wider screens, with JS enabled, the tab component will be active, so the visual distinction is already quite clear. 
We've also made the distinction programmatically by giving the preview container the `region` ARIA role. This should make it easier for screen reader users to:
- navigate to their preview 
- realise that they're inside the preview box if they navigate into it from a rotor or other navigation UI.

### DIFF
--- a/app/components/markdown_editor_component/_index.scss
+++ b/app/components/markdown_editor_component/_index.scss
@@ -2,6 +2,12 @@
   @include govuk-visually-hidden;
 }
 
+.app-markdown-editor__preview-area {
+  border: 2px solid $govuk-border-colour;
+  padding: 30px;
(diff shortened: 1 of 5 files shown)


thomasiles (approved):
nice 🏅

PR #652, opened 2023-09-26
Content updates to 'Add guidance' page

C-Harry:

Trello card: https://trello.com/c/gVtM7i1U/1046-guidance-clarify-why-a-page-heading-is-needed-for-guidance

I've made minor content updates in a couple of files to help users understand why they need to add a page heading when using guidance. 

Both updates will appear on the 'Add guidance' page in a form - this is the page users will see when they click the 'Add guidance' button on an 'Edit question' page.

- Any typos or mishaps with the html?!


DavidBiddle (approved):
Looks great! 🎉

PR #653, opened 2023-09-27
WIP - add/edit question steps saving to ActiveRecord instead of overloading the session

aliuk2012:

This PR is a major refactor of how add/edit question user journey works.  We were temporarily storing question settings in the session (Session was defaulted to Rails cookie storage). Cookie storage has a 4kb limit and its not maintainable to keep using this for the page as more options get added to it. 

Ultimately, the problem is that we didn't have a place to store the data as a user went through adding/editing a question wizard. In this PR, we introduce a new ActiveRecord class called `DraftQuestion` which will be used to store the question settings as a user supplies them and will then delete that record once it has been sent and stored in `forms-api`. 

Each part of the wizard uses a Form Object to break up the various parts of the question. These parts run validations for each page and will not save to DraftQuestion until the part is validated. When the Form Object saves its data to DraftQuestion it does not run DraftQuestion validations as part of the save (i.e `.save!(validate: false)`).

Once the user has gone through each step and has successfully  completed all the parts and has a question. `DraftQuestion.valid?` will run to check that the final version of the record is valid, if it is then it goes on to getting creating/updating a `Page` and sending on that info to `forms-api`.

Trello card: 

## TODO:

- [ ] Identify why `is_optional` is being set as `nil` in forms-api. the params forms-admin received is set to true  but the record ends up being created with `nil`
- [ ] Remove `load_from_session` page method. Theres no reason to load from a session.

### DIFF
--- a/app/forms/pages/type_of_answer_form.rb
+++ b/app/forms/pages/type_of_answer_form.rb
@@ -1,11 +1,28 @@
 class Pages::TypeOfAnswerForm < BaseForm
-  attr_accessor :answer_type, :form, :page
+  attr_accessor :answer_type, :draft_question
 
   validates :answer_type, presence: true, inclusion: { in: Page::ANSWER_TYPES }
 
-  def submit(session)
+  def submit
     return false if invalid?
 
-    session[:page] = { answer_type:, answer_settings: nil }
+    draft_question.answer_type = answer_type
+    draft_question.answer_settings = default_answer_settings_for_answer_type # TODO: We should only clear this if the answer_type changes
+    draft_question.save!(validate: false)
+  end
+
+private
+
+  def default_answer_settings_for_answer_type
+    case answer_type
+    when "selection"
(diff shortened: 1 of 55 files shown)


app/forms/pages/type_of_answer_form.rb
```diff
@@ -1,11 +1,28 @@
...
-  attr_accessor :answer_type, :form, :page
+  attr_accessor :answer_type, :draft_question
 
   validates :answer_type, presence: true, inclusion: { in: Page::ANSWER_TYPES }
 
-  def submit(session)
+  def submit
     return false if invalid?
 
-    session[:page] = { answer_type:, answer_settings: nil }
```
aliuk2012, line 9: this line is extremely destructive and can leave a session in a broken state, if the user is clicking back and changing previous answer types and then decides to go back to "Edit question" and saves theres nothing to catch that answer settings is nil

PR #654, opened 2023-09-27
Add mailer and logging for upgrade requests

thomasiles:
Add sending of email for user_upgrade_request and logging

When a user requests to upgrade from a trail user to an editor, they should receive an email with more instructions.

The email should have a reply-to of our teams zendesk account so that when the user replies, the answer is passed on to the engagement team.

An event is also logged.

a screenshot of the email:
image

Trello card: https://trello.com/c/c4WHxHZi/1039-send-an-email-to-the-account-upgrade-requestor-to-ask-for-more-information-and-create-account-upgrade-requested-confirmation-pag

### DIFF
--- a/spec/features/users/user_upgrade_request_spec.rb
+++ b/spec/features/users/user_upgrade_request_spec.rb
@@ -1,16 +1,6 @@
 require "rails_helper"
 
 describe "Request an upgrade from trial user to editor", type: :feature do
-  before do
-    # prevent notify being called
-    mailer = instance_double(ActionMailer::MessageDelivery)
-    allow(mailer).to receive(:deliver_now)
-
-    allow(UserUpgradeRequestMailer).to receive(:upgrade_request_email).with(
-      user_email: trial_user.email,
-    ).and_return(mailer)
-  end
-
   it "A trial user can request an upgrade when they declare they meet the requirements" do
     login_as trial_user
     visit_upgrade_page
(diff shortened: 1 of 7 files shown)


spec/mailers/user_upgrade_request_mailer_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+
+describe UserUpgradeRequestMailer do
+  describe "sending an email to a user" do
+    let(:mail) { described_class.upgrade_request_email(user_email: "test@example.gov.uk") }
+
+    before { mail }
```
aliuk2012, line 7: I dont think this is needed, the first time you call `mail` in the test it will automatically call line 5 and initialise it and its value

DavidBiddle (approved):
Looks good! 🎉

spec/features/users/user_upgrade_request_spec.rb
```diff
@@ -1,6 +1,16 @@
 require "rails_helper"
 
 describe "Request an upgrade from trial user to editor", type: :feature do
+  before do
+    # prevent notify being called
```
lfdebrux, line 5: I don't _think_ we need to worry about preventing Notify being called... in our Rails test environment we use the `:test` mailer, not the Notify one:

https://github.com/alphagov/forms-admin/blob/7b09e0005c53242a17f30e207633ad9091be8b65/config/environments/test.rb#L41-L44

At least, that's what I've been relying on when I've been writing tests of sending mail in forms-runner 😅 

https://github.com/alphagov/forms-runner/blob/0fc0a5eee0d6f0bfa0e0e6db3005cf4d8eec9e82/spec/requests/forms/submit_answers_controller_spec.rb#L103-L120

lfdebrux (approved):

lfdebrux:
Tested this locally, all looks good to me 🎉

PR #655, opened 2023-09-27
Add feature flag for metrics

DavidBiddle:

Trello card: https://trello.com/c/KRuPX20O/1037-add-a-feature-flag-for-metrics-for-form-creators

Adds a feature flag for the metrics feature. We're not actually doing anything with the flag yet. See also: https://github.com/alphagov/forms-deploy/pull/318

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,6 +1,7 @@
 # Used to add feature flags in the app to control access to certain features.
 features:
   detailed_guidance_enabled: false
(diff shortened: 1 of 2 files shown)


aliuk2012:
Could you add a spec for the feature flag?
https://github.com/alphagov/forms-admin/blob/main/spec/config/settings_spec.rb#L21-L25

aliuk2012 (approved):

PR #657, opened 2023-09-29
Add client side validation for guidance

DavidBiddle:

Trello card: https://trello.com/c/aQ09lXJw/1041-guidance-build-solution-to-add-client-side-validation-to-markdown-editor

For the guidance editor, we have distinct non-JS and JS-enhanced experiences. In the non-JS experience, when the user presses the 'Preview' button their input is validated on the server side. In the JS-enhanced journey, the preview currently happens on the client-side without a page refresh, so the input is not validated.

In this PR, we add an `errors` key to the response from the render_preview endpoint, which can have 0 or more errors as its value:
```jsonc
// input: '# A second-level heading'
{
  "preview_html": "\u003ch2 class=\"govuk-heading-m\"\u003e# A second-level heading\u003c/h2\u003e", 
  "errors": []
}
```

```jsonc
// input: '# A top-level heading'
{
  "preview_html": "\u003cp class=\"govuk-body\"\u003eA top-level heading\u003c/p\u003e", 
  "errors": ["Guidance text can only contain formatting for links, subheadings (##), bulleted lists (*), or numbered lists (1.)"]
}
```

We then handle this in the UI by:
- checking if there's a server-side error message present already, and creating a client-side one if not
- adding the `.govuk-form-group--error` class to the form group element if there is an error
- removing the error message and the `.govuk-form-group--error` class if the user updates the content and fixes the error

As devloped here the validation script won't do anything with the error summary at the top of the page - it's only concerned with the inline error message and the error class on the markdown field. This means that:

- if the page has a server-side error, the error summary will remain present even if the user fixes the error.
- if the page doesn't have a server-side error, but the user introduces an error and the error message appears on the client-side, the error summary won't be displayed.

This may not be ideal behaviour (but if we did make the error summary appear and disappear, this would likely cause annoying and confusing vertical layout shifts for the user).

This will stay in Draft for now because it's speculative work - the design hasn't been completed yet (that's covered in this card: https://trello.com/c/Gk4wELR5/1040-guidance-design-add-client-side-validation-to-markdown-editor ). I will review with @christophercameron-ixd when he's back and we can decide what to do with it.

### DIFF
--- a/app/components/markdown_editor_component/_index.scss
+++ b/app/components/markdown_editor_component/_index.scss
@@ -93,3 +93,7 @@
     font-family: inherit;
   }
 }
+
+.app-markdown-editor__error-message {
+  margin: 0;
+}
--- a/app/controllers/pages/guidance_controller.rb
+++ b/app/controllers/pages/guidance_controller.rb
@@ -55,8 +55,9 @@ class Pages::GuidanceController < PagesController
 
   def render_preview
     guidance_form = Pages::GuidanceForm.new(guidance_markdown: params[:guidance_markdown])
+    guidance_form.validate
 
-    render json: { preview_html: preview_html(guidance_form) }.to_json
+    render json: { preview_html: preview_html(guidance_form), errors: guidance_form.errors[:guidance_markdown] }.to_json
   end
 
 private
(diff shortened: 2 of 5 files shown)


thomasiles (approved):
Nice! Looks good to me

PR #663, opened 2023-10-04
Add metrics section to live form page

DavidBiddle:

Trello card: https://trello.com/c/n9Xq0v5L/1071-render-on-a-forms-info-page-the-metric-for-total-submissions-for-the-last-7-days

This PR creates a new 'Metrics summary' component that displays the number of submissions in the last week. The component also handles rendering the error states for metrics when a form is too new to have any metrics, or when the metrics data cannot be fetched from CloudWatch. This component will be rendered on the live form page if the `metrics_for_form_creators_enabled` feature flag is set to `true`.

This PR also includes [a commit updating the browserslist in package.json](https://github.com/alphagov/forms-admin/pull/663/commits/c5d1536e9a80a0f4cdae77429eb1b797f653c21d) to fix a layout issue spotted in this work. See the inidividual commit message for more information about what this change means and why it's necessary.

If running locally you'll need to:
- have `metrics_for_form_creators_enabled` set to `true`
- run it with `aws vault` to be able to read from CloudWatch
  - i.e. run `bin/setup` then `aws-vault exec gds-forms-dev-admin -- bin/rails s`
- have a live form set up locally whose `live_at` date is before today (you may need to fake this in the database)

#### When metrics are available
A large heading that says 'A live form with metrics, followed by a smaller heading that says 'Form metrics for the past 7 days: 4 October to 10 October 2023'.  Under the headings there is inset text that says 'If you want to track metrics over a longer period you'll need to make a note of these on the same day each week', and under that is the text 'Forms submitted' follwed by the number 1032 in large font.

#### When the form was made today
A large heading that says 'A live form with metrics, followed by a smaller heading that says 'Form metrics for the past 7 days: 4 October to 10 October 2023'. Under the headings there is inset text that says 'No metrics are available yet as no-one has started or submitted a form since it went live. Once they have, you'll be able to see the following metrics for the past 7 days: Number of forms completed, Number of forms started but not completed, Completion rate'

#### When Cloudwatch returns an Aws::CloudWatch::Errors::ServiceError
A large heading that says 'A live form with metrics, followed by a smaller heading that says 'Form metrics for the past 7 days: 4 October to 10 October 2023'. Under the headings there is inset text that says 'Sorry, there's a problem getting your form's metrics. Try again later.'

### DIFF
--- a/app/components/metrics_summary_component/_index.scss
+++ b/app/components/metrics_summary_component/_index.scss
@@ -1,0 +1,36 @@
+.app-metrics__container {
+  margin-bottom: govuk-spacing(6);
+
+  @include govuk-media-query(tablet) {
+    display: grid;
+    grid-template-columns: repeat(3, minmax(govuk-px-to-rem(100), 1fr));
+    gap: $govuk-gutter;
+  }
+}
+
+.app-metrics__big-number {
+  @include govuk-typography-responsive(19);
+  @include govuk-typography-common;
+
+  padding: govuk-spacing(1) govuk-spacing(0);
+  margin-bottom: govuk-spacing(2);
+
+  @include govuk-media-query(tablet) {
+    @supports (display: grid) {
+      display: flex;
+      flex-flow: column nowrap;
+      justify-content: space-between;
+
+      border-top: $govuk-border-width solid $govuk-border-colour;
+      border-bottom: $govuk-border-width solid $govuk-border-colour;
+    }
+  }
+}
+
+.app-metrics__big-number-number {
+  @include govuk-typography-responsive(48);
+  @include govuk-typography-weight-bold;
+
+  display: block;
+  margin-top: govuk-spacing(1);
+}
--- a/app/components/metrics_summary_component/view.rb
+++ b/app/components/metrics_summary_component/view.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module MetricsSummaryComponent
+  class View < ViewComponent::Base
+    attr_accessor :start_date, :end_date, :weekly_submissions, :weekly_starts, :weekly_started_but_not_completed, :weekly_completion_rate, :form_has_metrics, :error_message
+
+    def initialize(metrics_data)
+      super
+      set_dates
+
+      if metrics_data.nil?
+        @error_message = I18n.t("metrics_summary.errors.error_loading_data_html")
+      elsif metrics_data[:form_is_new]
+        @error_message = I18n.t("metrics_summary.errors.new_form_html")
+      else
+        @weekly_submissions = metrics_data[:weekly_submissions]
+      end
+    end
+
+    def render?
+      FeatureService.enabled?(:metrics_for_form_creators_enabled)
+    end
+
+    def formatted_date_range
+      start_date_format_string = start_date.year == end_date.year ? "%e %B" : "%e %B %Y"
+      I18n.t("metrics_summary.date_range", start_date: start_date.strftime(start_date_format_string).strip, end_date: end_date.strftime("%e %B %Y").strip)
+    end
+
+  private
+
+    def set_dates
+      @start_date = 1.week.ago.to_date
+      @end_date = 1.day.ago.to_date
+    end
+  end
+end
--- a/app/controllers/forms/live_controller.rb
+++ b/app/controllers/forms/live_controller.rb
@@ -2,7 +2,7 @@ class Forms::LiveController < Forms::BaseController
   after_action :verify_authorized
   def show_form
     authorize current_form, :can_view_form?
-    render template: "live/show_form", locals: { form_metadata: current_form, form: current_live_form }
+    render template: "live/show_form", locals: { form_metadata: current_form, form: current_live_form, metrics_data: }
   end
 
   def show_pages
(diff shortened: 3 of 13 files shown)


app/controllers/forms/live_controller.rb
```diff
@@ -15,4 +15,16 @@ def show_pages
   def current_live_form
     Form.find_live(params[:form_id])
   end
+
+  def metrics_data
+    today = Time.zone.today
+    form_is_new = (today.to_date - Time.zone.parse(current_form.live_at).to_date).to_i < 1
+    # TODO: fetch real data from CloudWatch
```
SamJamCul, line 22: This should work fine, if we get an error from the cloudwatch service we might need to put some sort of handler here

DavidBiddle: Have rebased onto your branch and made it read real data - included error handling as part of that.

app/components/metrics_summary_component/view.rb
```diff
@@ -0,0 +1,37 @@
...
+      start_date_format_string = start_date.year == end_date.year ? "%e %B" : "%e %B %Y"
+      I18n.t("metrics_summary.date_range", start_date: start_date.strftime(start_date_format_string).strip, end_date: end_date.strftime("%e %B %Y").strip)
+    end
+
+  private
+
+    def set_dates
+      today = Time.zone.today
+      @start_date = today - 8.days
+      @end_date = today - 1.day
```
aliuk2012, line 34:
```suggestion
      @start_date = 1.week.ago
      @end_date = 1.day.ago
```

DavidBiddle: Had to tweak this slightly but good spot! Also made me realise that I was originally off by one on the start date

app/views/live/show_form.html.erb
```diff
@@ -4,51 +4,53 @@
...
-    <h1 class="govuk-heading-l govuk-!-margin-bottom-2">
+    <h1 class="govuk-heading-xl govuk-!-margin-bottom-2">
       <%= form.name %>
     </h1>
-
     <%= render FormStatusTagDescriptionComponent::View.new(status: :live) %>
 
+    <%= render MetricsSummaryComponent::View.new(metrics_data) %>
+
+    <h2 class="govuk-heading-l">Your form</h2>
```
aliuk2012, line 14: Does this need to be behind feature flag or are we happy for this to show even if metric feature is disabled?

app/components/metrics_summary_component/_index.scss
```diff
@@ -0,0 +1,36 @@
...
+  margin-bottom: govuk-spacing(2);
+
+  @include govuk-media-query(tablet) {
+    @supports (display: grid) {
+      display: flex;
+      flex-flow: column nowrap;
+      justify-content: space-between;
+
+      border-top: 5px solid $govuk-border-colour;
+      border-bottom: 5px solid $govuk-border-colour;
```
aliuk2012, line 25: Could we use govuk spacing rather than hardcoding our own px units here? https://design-system.service.gov.uk/styles/spacing/

app/components/metrics_summary_component/_index.scss
```diff
@@ -0,0 +1,36 @@
...
+    gap: $govuk-gutter;
+  }
+}
+
+.app-metrics__big-number {
+  @include govuk-typography-responsive(19);
+  @include govuk-typography-common;
+
+  padding: 5px 0;
+  margin-bottom: govuk-spacing(2);
```
aliuk2012, line 16: could we use govuk spacing? maybe using responsive spacing? lhttps://design-system.service.gov.uk/styles/spacing/#applying-spacing-in-your-own-css

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "app",
   "private": "true",
   "browserslist": [
-    ">0.1%",
+    ">0.1% and not dead",
```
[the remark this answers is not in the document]
aliuk2012: Might be worth sync with forms-runner

app/views/live/show_form.html.erb
DavidBiddle: Good question! We've changed the whole heading structure below to account for this new `h2` (basically everything has been bumped down one heading level) so I think our options are (in my order of preference):

1. keep this h2 regardless of whether the feature flag is on or off (the current solution - slightly changes the design when the feature flag is off)
2. use the feature flag in each individual heading below this to determine its level (a bit complex)
3. feature flag the whole area below (simpler to feature flag but means significant duplication beetween the two versions)
4. not feature flag the headings below, so we'd have a h1 followed by h3s with the feature flag turned off (not ideal from an accessibility/HTML structure perspective as it makes the document heirarchy harder to understand)

For me option 1 made the most sense, since it's less complex to implement and the heading still makes sense without the metrics present.

app/components/metrics_summary_component/_index.scss
DavidBiddle: Yes, will use govuk-spacing here. Don't think it's worth using responsive spacing here at the moment, since the small spacing levels we're using here don't change between screen sizes in the responsive mixin.

app/components/metrics_summary_component/_index.scss
DavidBiddle: Since it's border I don't think we should use `govuk-spacing` as I think that's mainly intended for margin and padding, but we can use the $govuk-border-width variable here, so good spot!

app/components/metrics_summary_component/_index.scss
DavidBiddle: Done!

app/components/metrics_summary_component/_index.scss
DavidBiddle: Done!

package.json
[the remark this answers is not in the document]
DavidBiddle: Good pont, have done that here: https://github.com/alphagov/forms-runner/pull/442

app/views/live/show_form.html.erb
aliuk2012: 1 for the winner, i think

aliuk2012 (approved):

PR #665, opened 2023-10-05
Allow super admins to add and change a user's name

lfdebrux:

Trello card: https://trello.com/c/tUxg95jK/1068-allow-super-admins-to-add-and-change-a-users-name 

When we switch to the Auth0 passwordless email flow we won't get a name for a user when they first sign up. We've agreed for now that super admins will be able to set a user's name, and will do so when upgrading their account from the trial role. This PR adds a name field to the edit user form.

<details><summary>

### Screenshots

</summary>

#### Users list

Screenshot of users list page `/users`

#### Edit a user with no name set

Screenshot of edit user page `/users/:id/edit` when user has no name set

#### Error if trying to give user with no name or organisation set an editor role

Screenshot of edit user page in error state

</details>

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -67,10 +67,10 @@ class User < ApplicationRecord
 private
 
   def requires_name?
-    name_was.present? || role_changed?(from: :trial)
+    name_was.present? || !trial?
   end
 
   def requires_organisation?
-    organisation_id_was.present? || role_changed?(to: :editor)
+    organisation_id_was.present? || editor?
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,6 +14,7 @@ class User < ApplicationRecord
     trial: "trial",
   }
 
+  validates :name, presence: true, if: :requires_name?
   validates :role, presence: true
   validates :organisation_id, presence: true, if: :requires_organisation?
   validates :has_access, inclusion: [true, false]
@@ -64,6 +65,10 @@ class User < ApplicationRecord
   end
 
 private
+
+  def requires_name?
+    name_was.present? || role_changed?(from: :trial)
+  end
 
   def requires_organisation?
     organisation_id_was.present? || role_changed?(to: :editor)
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -756,6 +756,7 @@ en:
         description: Will be able to use GOV.UK Forms
         name: Permitted
     index:
+      name_blank: No name set
       organisation_blank: No organisation set
       table_headings:
         access: Access
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -33,6 +33,8 @@ en:
               too_long: Question text must be %{count} characters or less
         user:
           attributes:
+            name:
+              blank: You must enter a name for editor or super admin users
             organisation_id:
               blank: Select the user’s organisation
             role:
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -188,4 +188,30 @@ describe User, type: :model do
       end
     end
   end
+
+  context "when updating name" do
+    it "is valid to leave name unset" do
+      user = create :user, :with_no_name
+      user.name = nil
+      expect(user.valid?).to be true
+    end
+
+    it "is not valid to unset name if it is already set" do
+      user = create :user, name: "Test User"
+      user.name = nil
+      expect(user.valid?).to be false
+    end
+
+    it "is not valid to leave name unset if changing role to editor" do
+      user = create :user, :with_no_name, role: :trial
+      user.role = :editor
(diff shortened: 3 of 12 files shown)


config/locales/en.yml
```diff
@@ -33,6 +33,8 @@ en:
               too_long: Question text must be %{count} characters or less
         user:
           attributes:
+            name:
+              blank: You must enter a name for editor or super admin users
             organisation_id:
               blank: Select the user’s organisation
```
lfdebrux, line 39: @hannahkc as well as making it so that the user has to have a name set unless they have a trial account, I also made it an error to unset their name if they already have one set (we do the same thing for the organisation, so I thought it would be best to be consistent). Do you think we need a different error message for that scenario?

Also, there's a bit of inconsistency between the different error messages for this page, do we want to do anything about that?

Screenshot 2023-10-05 at 08 46 40

spec/models/user_spec.rb
```diff
@@ -188,4 +188,24 @@
       end
     end
   end
+
+  context "when updating name" do
+    it "is valid to leave name unset" do
+      user = create :user, :with_no_name
+      user.name = nil
+      expect(user.valid?).to be true
```
aliuk2012, line 196:
```suggestion
      user = build :user, :with_no_name
      expect(user).to be_valid
```

I switched the `create` to `build` because every time you use `create` it creates an actual entry in the DB which is not always necessary to check if a class instance is valid. For this reason `build` and `build_stub` are performant than `create`. 

Also could we not use `let` statements at the top of the specs to create instances and pass in different contexts as we test them? For example https://github.com/alphagov/forms-admin/blob/main/spec/models/draft_question_spec.rb#L37-L56 you will see we testing `question_text` but the instance of DraftQuestion is only defined once at the top of the spec file. Sometimes you do need a a different instance but I dont think we need to repeat creating instances for each specification

aliuk2012: https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#build-strategies Not one for now but we should look at our build strategies that we are using. I would try to use `build` more than `create` because it doesn't "touch" the database like "create" does but also sometimes `build_stubbed` might be a better option for testing and would be more performant. what we dont want is for our test suite to grind to a halt because even test is constantly reading/writing to the database. I've been in teams where test suites have taken like 30mins to run and by switching build strategies for tests helped almost half the time tests took to run.

lfdebrux: Unfortunately with this test group `build` doesn't always work, because the validation we're testing requires a persistence layer:

https://github.com/alphagov/forms-admin/blob/ce694cfae204154cbfe09008244a38b54a7a994f/app/models/user.rb#L69-L71

Using `build` instead of `create` causes the tests as written to fail.

aliuk2012: oh I see, what about the suggestion of using lets?

lfdebrux: I had a look at using `let`, I think for this set of tests we'd need a different context for each test... I'm not sure if that's better than doing the `create` in the test, what do you think?

app/models/user.rb
```diff
@@ -65,6 +66,10 @@ def update_user_forms
 
 private
 
+  def requires_name?
+    name_was.present? || role_changed?(to: :editor)
```
aliuk2012, line 70: Can I double check this?

So we want to make `name` required if the user already had a name or if their role is being upgraded to :editor?

```suggestion
    name_was.present? || editor?
```
This would make it required to always have a name if the user previously had a name set or if they are actually an editor. I dont think we need to check if their role is changing to make it required.

lfdebrux: Thanks for flagging this up... it made me realise it should be `role_changed?(from: :trial)` 😅

lfdebrux: I guess we could also do `name_was.present? || editor? || super_admin?`, or `name_was.present? || !trial?`. Do you have a preference from those three options?

aliuk2012: I think the last option makes the most sense and clear that all users should have the name except if they are a trial user

lfdebrux: Okay, I've changed this and the other use of `role_changed` in a new commit: https://github.com/alphagov/forms-admin/pull/665/commits/2a83c29dfa3db3bc932a328a70208d6066c5f0c7

spec/models/user_spec.rb
aliuk2012: Happy to just keep it as is for the time being.  We can always come back to refactor this in the future.

aliuk2012 (approved):

config/locales/en.yml
lfdebrux: Going to resolve this in a separate PR.

PR #668, opened 2023-10-06
Log page ID with requests

lfdebrux:

When searching Splunk for requests related to a particular page of a form it would be nice to be able to construct a query with `page_id=n`.

This PR adds the page ID to the log event JSON payload so that this will be possible in future.

See https://github.com/alphagov/forms-runner/pull/431 for a similar PR for the runner app.


thomasiles (approved):
nice

PR #670, opened 2023-10-09
1070 fetch from cloudwatch the total submissions of a form for the last 7 days in forms admin

SamJamCul:

Trello card: https://trello.com/c/73R1GLQV/1070-fetch-from-cloudwatch-the-total-submissions-of-a-form-for-the-last-7-days-in-forms-admin

This adds a service for obtaining submission stats for a given form. If there are stats available in the appropriate namespace in cloudwatch (based on the environment the request is coming from), then it should end up giving a number. 

Bit of a pain to test, but I can go through it with you. It will involve running things through the gds-cli as an aws role, which will enable submissions being logged with cloudwatch. This in turn will mean you can get data back from cloudwatch that isn't just the default 0. 

I'm not sure if the default response should be 0 if there's no data for a form - if haven't been any metrics sent to cloudwatch within the 1 week timeframe, it just doesn't send anything back. It's not possible to do the SUM of no data, so defaulting to 0 is assuming that no data == no submissions.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,6 +72,9 @@ gem "validate_url"
 # For auditing tables
 gem "paper_trail"
 
+# For obtaining CloudWatch metrics
+gem "aws-sdk-cloudwatch", "~> 1.81"
+
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem "debug", platforms: %i[mri mingw x64_mingw]
--- a/app/service/cloud_watch_service.rb
+++ b/app/service/cloud_watch_service.rb
@@ -1,0 +1,33 @@
+class CloudWatchService
+  def self.week_submissions(form_id:)
+    region = "eu-west-2"
+    a_week = 604_800
+    cloudwatch_client = Aws::CloudWatch::Client.new(region:)
+
+    response = cloudwatch_client.get_metric_statistics({
+      metric_name: "submitted",
+      namespace: metric_namespace.to_s,
+      dimensions: [
+        {
+          name: "form_id",
+          value: form_id.to_s,
+        },
+      ],
+      start_time: start_of_today - 7.days,
+      end_time: start_of_today,
+      period: a_week,
+      statistics: %w[Sum],
+      unit: "Count",
+    })
+
+    response.datapoints[0]&.sum.to_i || 0
+  end
+
+  def self.metric_namespace
+    "forms/#{Settings.forms_env}".downcase
+  end
+
+  def self.start_of_today
+    Time.zone.now.midnight
+  end
+end
--- a/spec/service/cloud_watch_service_spec.rb
+++ b/spec/service/cloud_watch_service_spec.rb
@@ -1,0 +1,69 @@
+require "rails_helper"
+
+describe CloudWatchService do
+  describe "#week_submissions" do
+    let(:forms_env) { "test" }
+    let(:form_id) { 3 }
+    let(:datapoints) { [{ sum: total_submissions }] }
+    let(:total_submissions) { 3.0 }
+
+    around do |example|
+      Timecop.freeze(Time.zone.local(2021, 1, 1, 4, 30, 0)) do
+        example.run
+      end
+    end
+
+    before do
+      allow(Settings).to receive(:forms_env).and_return(forms_env)
+    end
+
+    it "calls the cloudwatch client with get_metric_statistics" do
+      cloudwatch_client = Aws::CloudWatch::Client.new(stub_responses: true)
+      metric_response = cloudwatch_client.stub_data(:get_metric_statistics, datapoints:)
+
+      allow(Aws::CloudWatch::Client).to receive(:new).and_return(cloudwatch_client)
+
+      allow(cloudwatch_client).to receive(:get_metric_statistics).with({
+        metric_name: "submitted",
+        namespace: "forms/#{forms_env}",
+        dimensions: [
+          {
+            name: "form_id",
+            value: form_id.to_s,
+          },
+        ],
+        start_time: Time.zone.now.midnight - 7.days,
+        end_time: Time.zone.now.midnight,
+        period: 604_800,
+        statistics: %w[Sum],
+        unit: "Count",
+      }).and_return(metric_response)
+
+      described_class.week_submissions(form_id:)
+    end
+
+    it "returns the sum for the datapoint in the response" do
+      cloudwatch_client = Aws::CloudWatch::Client.new(stub_responses: true)
+      metric_response = cloudwatch_client.stub_data(:get_metric_statistics, datapoints:)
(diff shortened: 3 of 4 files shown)


app/service/cloud_watch_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      namespace: metric_namespace.to_s,
+      dimensions: [
+        {
+          name: "form_id",
+          value: form_id.to_s,
+        },
+      ],
+      start_time: Time.zone.now.midnight - 7.days,
+      end_time: Time.zone.now.midnight,
+      period: 604_800,
```
chao-xian, line 17: Can you set this to a variable/constant describing what the period means please? (it looks like a magic number - sorry if I missed this in forms-runner)

spec/service/cloud_watch_service_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+    end
+
+    before do
+      allow(Settings).to receive(:forms_env).and_return(forms_env)
+    end
+
+    it "calls the cloudwatch client with get_metric_statistics" do
+      cloudwatch_client = Aws::CloudWatch::Client.new(stub_responses: true)
+
```
chao-xian, line 22: Nitpick: can we clump together the setup stuff without line breaks before the `expect` pls?

spec/service/cloud_watch_service_spec.rb
```diff
@@ -0,0 +1,71 @@
+require "rails_helper"
+
+describe CloudWatchService do
+  describe "#week_submissions" do
+    let(:forms_env) { "test" }
+    let(:form_id) { 3 }
+    let(:datapoints) { [ { sum: } ] }
+    let(:sum) { 3.0 }
```
chao-xian, line 8: Can we rename this as total submissions or something please?

spec/service/cloud_watch_service_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        end_time: Time.zone.now.midnight,
+        period: 604_800,
+        statistics: %w[Sum],
+        unit: "Count",
+      }).and_return(metric_response)
+
+      described_class.week_submissions(form_id:)
+    end
+
+    it "returns the sum for the datapoint in the response" do
```
chao-xian, line 46: Can we rename `sum` to total submissions please?

chao-xian (commented):
Looks good! Just a few tiny points.

app/service/cloud_watch_service.rb
SamJamCul: You didn't miss it, this is just representing the previous week. I'll parse it out into something with proper names

SamJamCul: I've created some methods that better describe what points in time they're referring to

spec/service/cloud_watch_service_spec.rb
SamJamCul: Yep, renamed to explain what that number actually is. It was called sum as that's what the response uses, but we can give it more detail :D

spec/service/cloud_watch_service_spec.rb
SamJamCul: as above, `sum` is now `total_submissions` throughout the tests

Gemfile
```diff
@@ -118,3 +118,5 @@ group :test do
   # axe-core for running automated accessibility checks
   gem "axe-core-rspec"
 end
+
+gem "aws-sdk-cloudwatch", "~> 1.81"
```
aliuk2012, line 122: Could we move this up to where our other prod gems are listed?

app/service/cloud_watch_service.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    response.datapoints[0]&.sum.to_i || 0
+  end
+
+  def self.metric_namespace
+    "forms/#{Settings.forms_env}".downcase
+  end
+
+  def self.one_week_ago
+    Time.zone.now.midnight - 7.days
```
aliuk2012, line 31: I'm not sure we need this method specifically
```suggestion
    1.week.ago.beginning_of_day
```

or 
```suggestion
    1.week.ago.midnight
```

both options would return the same results.

Alternatively, reusing `start_of_today` method

```suggestion
    start_of_today - 7.days
```

app/service/cloud_watch_service.rb
```diff
@@ -0,0 +1,37 @@
+class CloudWatchService
+  def self.week_submissions(form_id:)
+    region = "eu-west-2"
+    a_week = 604_800
```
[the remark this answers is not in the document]
chao-xian: Does it really matter? As long as it still returns a good enough answer, it'll be fine.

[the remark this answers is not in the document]
DavidBiddle: The accuracy doesn't bother me too much as long as we're confident CloudWatch won't do anything unexpected in this case (I don't really know what the CloudWatch response looks like or if it's different when the `period` isn't the same as the time between the start and end dates)

Gemfile
chao-xian: Ooh good spot...

app/service/cloud_watch_service.rb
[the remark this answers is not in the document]
DavidBiddle: I suppose it's fairly low risk since:
- we're building this feature to handle cloudwatch errors gracefully
- the clocks go back later this month so we'll find out soon enough

Gemfile
SamJamCul: Done!

app/service/cloud_watch_service.rb
SamJamCul: I've gone with the last one, as it removes the most gubbins

app/service/cloud_watch_service.rb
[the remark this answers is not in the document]
SamJamCul: Just tested this, a period that overruns the start of the query isn't an issue - datapoints that aren't within the start and end times just aren't included in the data.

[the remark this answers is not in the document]
SamJamCul: And yeah for the scenario where the period doesn't cover the first hour of the saved day, we'll miss stats from that hour as it'll be part of another datapoint which we're not using. 

We could fix this by checking the length of the diff between the two points in time and seeing if they match the period. I'm leaning towards not bothering right now!

DavidBiddle (approved):
Looks good, and works when I rebase my branch onto it 🎉

app/service/cloud_watch_service.rb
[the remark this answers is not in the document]
DavidBiddle: That sounds alright to me - thanks for checking!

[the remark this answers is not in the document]
chao-xian: Yay! Nice one! Thanks for checking Sam

chao-xian:
Thanks for your work on this Sam! 👏

PR #672, opened 2023-10-10
Remove Alpine edge repositories from Dockerfile

lfdebrux:

Remove Alpine edge repositories from Dockerfile

In commit f360d84 we used both the stable and edge repos for Alpine because we wanted both Node 16 and OpenSSL 3 in our app image. However, today we no longer need any version of Node in our deployment image, and OpenSSL 3 is the default version of OpenSSL in Alpine 3.17.0 [[1]]. So we no longer have a need for edge.

With Alpine 3.17 at the time of this commit fetching the edge repo index fails, so building this Dockerfile fails with the error:

```
Step 24/35 : RUN apk update
 ---> Running in 247913f7587d
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/main/aarch64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/community/aarch64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/edge/main/aarch64/APKINDEX.tar.gz
ERROR: http://dl-cdn.alpinelinux.org/alpine/edge/main: BAD signature
WARNING: Ignoring http://dl-cdn.alpinelinux.org/alpine/edge/main: No such file or directory
v3.17.5-107-g07fd110f6bf [https://dl-cdn.alpinelinux.org/alpine/v3.17/main]
v3.17.5-108-gfe9a155566d [https://dl-cdn.alpinelinux.org/alpine/v3.17/community]
1 errors; 17699 distinct packages available
The command ‘/bin/sh -c apk update’ returned a non-zero code: 1
```

The Alpine wiki says that

> Warning: Do not enable stable and edge repos at the same time. This can break your system. Either use edge or stable.

[[2]].

As we no longer need the edge repos, and it does not appear to work, this commit removes the use of edge from the Dockerfile.

[1]: https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.17.0
[2]: https://wiki.alpinelinux.org/wiki/Repositories

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,8 +44,6 @@ ENV RAILS_ENV="${RAILS_ENV:-production}" \
 
 WORKDIR /app
 
-# Edge repo is necessary for Node 16 and openssl 3
-RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories
 RUN apk update
 RUN apk upgrade --available
 RUN apk add libc6-compat openssl-dev libpq


jamie-o-wilkinson (approved):
Nice one, bit of an eeek reading that warning in the Alpine wiki 😅

PR #676, opened 2023-10-11
Revert "Refactor predicate methods to simplify"

lfdebrux:

The refactor of the validation predicates in commit 2a83c29 caused unexpected side-effects. Although in most cases checking the current role is equivalent to checking whether the role has been changed, the two conditions are not exactly identical. We found that after applying this change we started getting errors from GOV.UK Signon because the Signonotron API user had an editor role and no organisation set.

Although the original logic was more complex, it more precisely describes the behaviour we want and has fewer bugs, so let's revert to that.

This reverts commit 2a83c29dfa3db3bc932a328a70208d6066c5f0c7.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -5,7 +5,7 @@ describe User, type: :model do
   subject(:user) { described_class.new }
 
   it "validates" do
-    expect(user.valid?).to be true
+    expect(user).to be_valid
   end
 
   describe "versioning", versioning: true do
@@ -27,7 +27,7 @@ describe User, type: :model do
     it "is invalid if blank" do
       user.role = nil
 
-      expect(user.valid?).to be false
+      expect(user).to be_invalid
     end
   end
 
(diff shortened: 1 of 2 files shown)


jamie-o-wilkinson (dismissed):

aliuk2012:
Interesting, should there not be some tests affected by this change?

lfdebrux:
> Interesting, should there not be some tests affected by this change?

The unwanted behaviour wasn't caught by any tests; I guess we could add some?

aliuk2012:
I think we should add some tests to catch the unwanted behavior, otherwise, if someone comes along and refactors it to what we had it's not going to be caught until it's in prod and might go unnoticed.

lfdebrux:
> I think we should add some tests to catch the unwanted behavior, otherwise, if someone comes along and refactors it to what we had it's not going to be caught until it's in prod and might go unnoticed.

Good point. I've added tests now.

spec/models/user_spec.rb
```diff
@@ -190,28 +209,44 @@
...
       user.role = :editor
-      expect(user.valid?).to be false
+      expect(user).not_to be_valid
     end
 
     it "is not valid to leave name unset if changing role to super admin" do
-      user = create :user, :with_no_name, role: :trial
       user.role = :super_admin
-      expect(user.valid?).to be false
+      expect(user).not_to be_valid
```
aliuk2012, line 249: Only a suggestion, any thoughts on using this instead? I find it easier to read but its a personal preference, either way is totally fine.

```suggestion
      expect(user).to be_invalid
```

aliuk2012 (approved):
Looks good and its also good that we filled in some missing specs.

Only a minor suggestion but otherwise happy to approve it

spec/models/user_spec.rb
lfdebrux: Works for me, I'll do that 👍

aliuk2012 (approved):

PR #677, opened 2023-10-11
Render form starts and completion percentage

DavidBiddle:

Trello card: https://trello.com/c/sXIsnAqj/1075-render-total-started-forms-for-the-last-7-days-and-forms-completion-rate-for-last-7-days-in-forms-admin

This PR adds two new metrics to the metrics summary component:
- the number of times a form has been started but not completed (calculated as the number of starts that week - the number of submissions)
- the percentage of form journeys that are completed (calculated as the number of submissions as a percentage of the total number of starts, rounded to the nearest whoile percentage point)

It also includes:
- a new error message to be displayed if the form hasn't had any starts that week
- a fix for an unhandled error when the AWS credentials have not been set (this is
most likely to happen on a developer's local machine)

See the individual commit messages for more detail.

If running locally you'll need to:
- have `metrics_for_form_creators_enabled` set to `true`
- run it with `aws vault` to be able to read from CloudWatch
  - i.e. run `bin/setup` then `aws-vault exec gds-forms-dev-admin -- bin/rails s`
- have a live form set up locally whose `live_at` date is before today (you may need to fake this in the database)

#### Message shown for a form with no starts in the last week
A large heading that says 'A live form with metrics, followed by a smaller heading that says 'Form metrics for the past 7 days: 5 October to 11 October 2023'. Under the headings there is inset text that says 'No metrics are available as no-one has started or submitted a form in the past 7 days.This may be because the form is no longer published on GOV.UK.'

#### Form with available stats in the last week
A large heading that says 'A live form with metrics, followed by a smaller heading that says 'Form metrics for the past 7 days: 5 October to 11 October 2023'.  Under the headings there is inset text that says 'If you want to track metrics over a longer period you'll need to make a note of these on the same day each week', and under that are three boxes containing form metrics. The metrics are 'Completion rate 66%', 'Forms submitted 1032' and 'Forms started but not completed 536' where in each acase the number in the box is in a large font size.

### DIFF
--- a/spec/components/metrics_summary_component/view_spec.rb
+++ b/spec/components/metrics_summary_component/view_spec.rb
@@ -41,13 +41,25 @@ RSpec.describe MetricsSummaryComponent::View, type: :component, feature_metrics_
     end
   end
 
+  describe "#calculate_percentage" do
+    it "returns nil if the total argument is zero" do
+      expect(metrics_summary.calculate_percentage(168, 0)).to eq(nil)
+    end
+
+    it "returns the percentage rounded to the nearest integer" do
+      expect(metrics_summary.calculate_percentage(168, 1000)).to eq(17)
+      expect(metrics_summary.calculate_percentage(253, 1000)).to eq(25)
+      expect(metrics_summary.calculate_percentage(965, 1000)).to eq(97)
+    end
+  end
+
   context "when metrics_data is null" do
     it "returns the 'error loading data' message" do
       expect(metrics_summary.error_message).to eq(I18n.t("metrics_summary.errors.error_loading_data_html"))
     end
 
     it "renders the error message" do
-      expect(page).to have_text(Nokogiri::HTML(metrics_summary.error_message).text)
+      expect(page).to have_css(".govuk-inset-text", text: Nokogiri::HTML(metrics_summary.error_message).text)
     end
   end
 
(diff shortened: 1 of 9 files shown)


app/service/cloud_watch_service.rb
DavidBiddle, line 1: This file is from Sam's branch, if you've got any comments put them in #678 to keep things neat

spec/controller/forms/live_controller_spec.rb
DavidBiddle, line 1: This file is from Sam's branch, if you've got any comments put them in #678 to keep things neat

spec/service/cloud_watch_service_spec.rb
DavidBiddle, line 1: This file is from Sam's branch, if you've got any comments put them in #678 to keep things neat

spec/components/metrics_summary_component/view_spec.rb
```diff
@@ -63,19 +75,45 @@
...
-    let(:metrics_data) { { weekly_submissions: 1235, form_is_new: false } }
+  context "when weekly starts is zero" do
+    let(:metrics_data) { { weekly_submissions: 0, form_is_new: false, weekly_starts: 0 } }
+
+    it "returns the 'no starts' message" do
+      expect(metrics_summary.error_message).to eq(I18n.t("metrics_summary.errors.no_submissions_html"))
+    end
+
+    it "renders the error message" do
+      expect(page).to have_text(Nokogiri::HTML(metrics_summary.error_message).text)
```
aliuk2012, line 86: Why are we using Nokogiri here?

DavidBiddle: As far as I can tell Capybara doesn't have matcher for checking whether the page contains a specific bit of HTML (these are HTML translations). So the best solution I could think of was to use the `has_text` matcher and use Nokogiri to get the translation's text value.

aliuk2012: 🤔  you should be able to `find` html using selectors but this specific test looks like it just looking for specific text on the page rather than specific HTML with text.

DavidBiddle: I think the finders would be on the `expect` side of the assertion (and I should probably add these). My issue is that I don't think Capybara provides a matcher to compare whatever's in the `expect` to the translation.

So I can do:

```ruby
  expect(page).to have_text(Nokogiri::HTML(metrics_summary.error_message).text)
```
or: 
```ruby
  expect(page.find(".govuk-inset-text")).to have_text(Nokogiri::HTML(metrics_summary.error_message).text)
```
or:
```ruby
  expect(page).to have_css(".govuk-inset-text", text: Nokogiri::HTML(metrics_summary.error_message).text)
```

but not:

```ruby
  expect(page).to have_text(metrics_summary.error_message)
```
or: 
```ruby
  expect(page.find(".govuk-inset-text")).to have_text(metrics_summary.error_message)
```
or:
```ruby
  expect(page).to have_css(".govuk-inset-text", text: metrics_summary.error_message)
```
because in this case `error_message` is a HTML string `"<p>Sorry, there's a problem getting your form's metrics.</p><p>Try again later.</p>"` and Capybara's matchers will strip the HTML tags from whatever's inside the `expect()`. So unless I'm missing something I think I have to either strip out the HTML tags on the right hand side as I'm doing here (either with Nokogiri or something else), or I'll have to hard-code text into the tests to get around the HTML tag stripping:
```ruby
  expect(page.find(".govuk-inset-text > p")).to have_text("Sorry, there's a problem getting your form's metrics.")
  expect(page.find(".govuk-inset-text > p")).to have_text("Try again later.")
```

aliuk2012: Ok, not to worry lets leave it as is

aliuk2012:
@DavidBiddle  based on the screenshots, could we keep the day and month together on the same line it looks weird to have 5 and then October wrapped on the next line? apart from that it looks good

spec/components/metrics_summary_component/view_spec.rb
DavidBiddle: I'll change them to `expect(page).to have_css(".govuk-inset-text", text: metrics_summary.error_message)` now, since I think that's a better, more specific test. Thanks for raising this!

aliuk2012 (approved):

PR #678, opened 2023-10-11
Add forms started CloudWatch metrics

SamJamCul:

Trello card: [1076-fetch-from-cloudwatch-the-total-started-forms-for-the-last-7-days-in-forms-admin
](https://trello.com/c/pwgdk1Tm/1076-fetch-from-cloudwatch-the-total-started-forms-for-the-last-7-days-in-forms-admin)
Continuing on from https://github.com/alphagov/forms-runner/pull/443, this change creates a new class method for obtaining the number of times the form has been started in the last week. There's a little bit of refactoring to remove some duplication. There is still a lot of duplication, but removing it seemed a bit unclean, and it's nice to have each method be in control of its own API call. 

### DIFF
--- a/app/service/cloud_watch_service.rb
+++ b/app/service/cloud_watch_service.rb
@@ -1,8 +1,9 @@
 class CloudWatchService
+  REGION = "eu-west-2".freeze
+  A_WEEK = 604_800
+
   def self.week_submissions(form_id:)
-    region = "eu-west-2"
-    a_week = 604_800
-    cloudwatch_client = Aws::CloudWatch::Client.new(region:)
+    cloudwatch_client = Aws::CloudWatch::Client.new(region: REGION)
 
     response = cloudwatch_client.get_metric_statistics({
       metric_name: "submitted",
@@ -15,7 +16,29 @@ class CloudWatchService
       ],
       start_time: start_of_today - 7.days,
       end_time: start_of_today,
-      period: a_week,
+      period: A_WEEK,
+      statistics: %w[Sum],
+      unit: "Count",
+    })
+
+    response.datapoints[0]&.sum.to_i || 0
+  end
+
+  def self.week_starts(form_id:)
+    cloudwatch_client = Aws::CloudWatch::Client.new(region: REGION)
+
+    response = cloudwatch_client.get_metric_statistics({
+      metric_name: "started",
+      namespace: metric_namespace.to_s,
+      dimensions: [
+        {
+          name: "form_id",
+          value: form_id.to_s,
+        },
+      ],
+      start_time: start_of_today - 7.days,
+      end_time: start_of_today,
+      period: A_WEEK,
       statistics: %w[Sum],
       unit: "Count",
     })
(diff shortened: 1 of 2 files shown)


spec/service/cloud_watch_service_spec.rb
```diff
@@ -66,4 +67,68 @@
...
+          },
+        ],
+        start_time: Time.zone.now.midnight - 7.days,
+        end_time: Time.zone.now.midnight,
+        period: 604_800,
+        statistics: %w[Sum],
+        unit: "Count",
+      }).and_return(metric_response)
+
+      described_class.week_starts(form_id:)
```
chao-xian, line 107: is there meant to be an expectation in this spec?

app/service/cloud_watch_service.rb
```diff
@@ -1,8 +1,9 @@
 class CloudWatchService
+  REGION = "eu-west-2".freeze
+  A_WEEK = 604_800
```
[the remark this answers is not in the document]
SamJamCul: Ah I got rubocopped as it was apparently pointless freezing it as its already immutable (numbers are immutable?).

https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/RedundantFreeze

spec/service/cloud_watch_service_spec.rb
SamJamCul: I was confused by this - I originally had line 91 start with an `expect` instead of `allow`. Rubocop told me off, and it turns out that the tests will still fail if the `allow` isn't triggered. I'm not a fan...

SamJamCul: You know what, after reading the docs for the cop, it looks like it's best practice to have both an allow and an expect. It comes out a bit weird imo, but at least there's an expect to hold on to.

https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/StubbedMock

SamJamCul: I've reworked it to use an expect as well as the allow, and I've also deduped some test setup bits

spec/service/cloud_watch_service_spec.rb
```diff
@@ -39,6 +40,22 @@
...
+            name: "form_id",
+            value: form_id.to_s,
+          },
+        ],
+        start_time: Time.zone.now.midnight - 7.days,
+        end_time: Time.zone.now.midnight,
+        period: 604_800,
+        statistics: %w[Sum],
+        unit: "Count",
+      })
```
DavidBiddle, line 57: Entirely optional but potentially you could use:

```suggestion
      }).once
```
here (and for the week_starts test below) to make the tests a bit more strict. Doesn't look like it'll make a huge difference with the code as it stands, but could catch double requests if we try any fancy refactoring in future.

DavidBiddle (dismissed):
Looks really good! Have raised one small thing in the tests but it's entirely optional, so don't let it block you 😄

spec/service/cloud_watch_service_spec.rb
SamJamCul: I do like this, especially as its an external API call so hitting it too hard accidentally could spell trouble!

DavidBiddle (approved):

PR #682, opened 2023-10-13
Prevent unnecessary CloudWatch request

DavidBiddle:

Trello card: None

We have a feature flag (:metrics_for_form_creators_enabled) to determine whether the metric component should be displayed on the live form page.

Currently we only use this flag in the component view file, to detemine whether the component should be rendered, but we don't use the flag in the controller. This means that we still send the get_metric_statistics request to CloudWatch, even if we're not going to render the metrics component.

This PR adds a guard clause to the controller's metrics_data method, so that the CloudWatch requests will not be sent if the flag is off.

### DIFF
--- a/app/controllers/forms/live_controller.rb
+++ b/app/controllers/forms/live_controller.rb
@@ -11,6 +11,8 @@ class Forms::LiveController < Forms::BaseController
   end
 
   def metrics_data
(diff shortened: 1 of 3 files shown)


jamie-o-wilkinson (approved):
👍

PR #683, opened 2023-10-13
Prevent metrics summary's start and end dates from breaking across lines

DavidBiddle:

Trello card: https://trello.com/c/eZZKFbZ8/1091-stop-dates-breaking-across-lines-in-metrics-summary-component

Adds a new class to the dates in the `formatted_date_range` method and uses CSS `white-space: nowrap` to stop the dates from breaking across lines.

Also includes:

- a small refactor to make the above easier to implement. 
- an update to the tests to check the HTML rendered by `render_inline` rather than using Nokogiri's `text` method.

See the individual commit messages if you'd like further detail on these.

#### Before
Text reading 'Form metrics for the past 7 days: 6 October to 12 October 2023' where the line breaks between the colon after the word the number 6 and the word 'October'.

#### After 
Text reading 'Form metrics for the past 7 days: 6 October to 12 October 2023' where the line breaks between the colon after the word 'days' and the number 6.

### DIFF
--- a/spec/components/metrics_summary_component/view_spec.rb
+++ b/spec/components/metrics_summary_component/view_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe MetricsSummaryComponent::View, type: :component, feature_metrics_
   end
 
   it "renders the start and end dates" do
-    expect(page).to have_css("h2", text: Nokogiri::HTML(metrics_summary.formatted_date_range).text)
+    expect(render_inline(metrics_summary).to_html).to include(metrics_summary.formatted_date_range)
   end
 
   describe "#formatted_date_range" do
@@ -59,7 +59,7 @@ RSpec.describe MetricsSummaryComponent::View, type: :component, feature_metrics_
     end
 
     it "renders the error message" do
-      expect(page).to have_css(".govuk-inset-text", text: Nokogiri::HTML(metrics_summary.error_message).text)
+      expect(render_inline(metrics_summary).to_html).to include(metrics_summary.error_message)
     end
   end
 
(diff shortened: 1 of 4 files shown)


spec/components/metrics_summary_component/view_spec.rb
```diff
@@ -14,7 +14,7 @@
   end
 
   it "renders the start and end dates" do
-    expect(page).to have_css("h2", text: metrics_summary.formatted_date_range)
+    expect(page).to have_css("h2", text: Nokogiri::HTML(metrics_summary.formatted_date_range).text)
```
chao-xian, line 17: I missed this in the other PR but saw your discussion with Al. 

Could we replace this with?:
```
expect(page).to have_css("p", text: "Sorry, there's a problem getting your form's metrics.")
expect(page).to have_css("p", text: "Try again later.")
```

chao-xian (dismissed):
LGTM thanks for this David! Just one tiny suggestion

spec/components/metrics_summary_component/view_spec.rb
DavidBiddle: I can do this, but the downside is our tests will break if we change the content. Been playing around with it and found that I can do this, which I think might be less fragile:
```suggestion
    expect(render_inline(metrics_summary).to_html).to include(metrics_summary.formatted_date_range)
```
and in the error cases that would look like: 
```
      expect(render_inline(metrics_summary).to_html).to include(metrics_summary.error_message)
```

DavidBiddle: I've updated to do the above @chao-xian

chao-xian: Nice! 

(As an aside: I'm personally not against tests breaking from content changes as to me being that explicitness and better readability is easier to parse but I'm not going to start that crusade)

chao-xian (approved):

spec/components/metrics_summary_component/view_spec.rb
DavidBiddle: Definitely worth having the conversation at some point, because there are ways it would make life easier

PR #684, opened 2023-10-16
Add view for incomplete week of metrics

DavidBiddle:

Trello card: https://trello.com/c/JSIJWVBM/1095-add-a-variation-of-the-metrics-component-for-when-the-form-has-less-than-a-weeks-worth-of-data

This PR adds logic to vary the heading and description text in the metrics component if the form went live less than a week ago. Specifically this PR:
 - Makes the number of days in the heading text ("Form metrics for the past 7 days: 9 October to 15 October 2023") vary (e.g. "Form metrics for the past 3 days: 13 October to 15 October 2023")
 - Adds a new translation for the description (`t("metrics_summary.description.incomplete_week")`)
 - Adds some logic to vary the heading depending on when the form went live. This is necessary to prevent the following bug cases:
   - a form that went live today showing the heading "Form metrics for the past 0 days: 17 October to 16 October 2023."
   - a form that went live yesterday showing the heading "Form metrics for the past 1 days: 16 October to 16 October 2023."

To make this work we have to pass the form's `live_at` date into the component, so this PR includes that too.

### Screenshots

Previously we'd see this view, even if the form went live after the start date the heading is claiming: 
A heading with text "Form metrics for the past 7 days: 10 October to 16 October 2023." and a description with the text "If you want to track metrics over a longer period you'll need to make a note of these on the same day each week."

#### For a form that went live less than a week ago, but not today or yesterday
A heading with text "Form metrics for the past 3 days: 14 October to 16 October 2023." and a description with the text "The metrics shown here are for less than a week. To see a full week’s metrics, check again when your form’s been live for 7 days."

#### For a form that went live yesterday
A heading with text "Form metrics for 16 October 2023." and a description with the text "The metrics shown here are for less than a week. To see a full week’s metrics, check again when your form’s been live for 7 days."

#### For a form that went live today
A heading with text "Form metrics" and inset text with the text "No metrics are available yet as no-one has started or submitted a form since it went live."

### DIFF
--- a/app/components/metrics_summary_component/view.html.erb
+++ b/app/components/metrics_summary_component/view.html.erb
@@ -1,5 +1,5 @@
 <div class="app-metrics">
-  <h2 class="govuk-heading-l"><%= t("metrics_summary.heading", number_of_days:, formatted_date_range:).html_safe %></h2>
+  <h2 class="govuk-heading-l"><%= heading %></h2>
   <% if error_message.present? %>
     <div class="govuk-inset-text">
         <%= error_message.html_safe %>
--- a/app/components/metrics_summary_component/view.html.erb
+++ b/app/components/metrics_summary_component/view.html.erb
@@ -5,7 +5,9 @@
         <%= error_message.html_safe %>
     </div>
   <% else %>
-    <p><%= t("metrics_summary.description") %></p>
+    <p>
+      <%= description %>
+    </p>
     <div class="app-metrics__container">
       <div class="app-metrics__big-number">
         <%= t("metrics_summary.completion_rate") %> <span class="app-metrics__big-number-number"><%= weekly_completion_rate %></span>
--- a/app/components/metrics_summary_component/view.html.erb
+++ b/app/components/metrics_summary_component/view.html.erb
@@ -1,5 +1,5 @@
 <div class="app-metrics">
-  <h2 class="govuk-heading-l"><%= t("metrics_summary.heading", formatted_date_range:).html_safe %></h2>
+  <h2 class="govuk-heading-l"><%= t("metrics_summary.heading", number_of_days:, formatted_date_range:).html_safe %></h2>
   <% if error_message.present? %>
     <div class="govuk-inset-text">
         <%= error_message.html_safe %>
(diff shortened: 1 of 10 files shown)


app/controllers/forms/live_controller.rb
```diff
@@ -2,20 +2,25 @@ class Forms::LiveController < Forms::BaseController
...
   end
 
   def show_pages
     authorize current_form, :can_view_form?
     render template: "live/show_pages", locals: { form: current_live_form }
   end
 
+  def form_live_date
+    Time.zone.parse(current_live_form.live_at.to_s).to_date
+  end
```
aliuk2012, line 15: This doesn't feel like the right place for this method.

1. This is essentially a decorated value we would like for the view so should be in model or somewhere other than a controller.
2. Are we going to use this anywhere else at the moment? if not maybe we could just inline it in the view if theres no clear place to put this.

app/components/metrics_summary_component/view.html.erb
```diff
@@ -1,11 +1,17 @@
...
         <%= error_message.html_safe %>
     </div>
   <% else %>
-    <p><%= t("metrics_summary.description") %></p>
+    <p>
+      <% if complete_week? %>
+        <%= t("metrics_summary.description.complete_week") %>
+      <% else %>
+        <%= t("metrics_summary.description.incomplete_week") %>
+      <% end %>
```
aliuk2012, line 13: could this not be a method in the View.rb  to return a summary description and then all you need to do is call it here. It would be easier to unit the hell out of it to make sure it returns what we expect it to or handle cases we dont expect.

app/controllers/forms/live_controller.rb
DavidBiddle: Fair point - I feel like putting it in the `form` model would probably make most sense. 

Part of me wonders if it would make sense to do the same with `metrics_data`, and then just pass `form` into the component (but that's a bigger change that I'd want to do in a separate PR, if I do it)

app/components/metrics_summary_component/view.html.erb
DavidBiddle: Done

app/controllers/forms/live_controller.rb
DavidBiddle: Have moved it onto the form model - think it's much better now so thanks!

PR #687, opened 2023-10-17
Add MOU signature model

thomasiles:
# Add a MouSignature resource

Trello card: https://trello.com/c/bj5SLNgQ/1080-store-in-database-for-mou-information-and-a-page-that-shows-it

This is the first of three PRs to add an MOU signing journey to the admin. This PR creates the db migration, model and interface to allow the engagement team to list the MOU's which have been signed.

The next PR will add the interface users use to view and agree to the MOU. The last PR will contain the HTML for the current MOU text.

We want to record that form creators have signed an MOU on behalf of
their organisation.

In early access, agreeing to an MOU will be managed by the engagement
team. This will change when we move to self-signup. The current MOU work
does not prevent form creators from making forms live, doesn't add a
step to the task list and the journey to agree to the MOU will be
initiated by the engagement team.

The requirements for MOU signature are:
- trial users and editors can sign an MOU
- more than one user belonging to an organisation can sign an MOU

Allowing trial users to sign an MOU adds a difficulty - trial users are
not associated with an organisation. To get round this, MouSignature
does not require organisation. Trial users can still sign an MOU and
orgainsation will be set to nil, when the user is associated with an
organisation at a later point, the MouSignature will be modified to add
the organisation.

It's possible to use Rails acronym inflections to use MOUSignature
instead of MouSignature instead. I decided it wasn't worth it though.

image


spec/models/mou_signature_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      user:,
+      organisation:,
+      agreed_at: Time.zone.now,
+    )
+  end
+
+  it "setting user to nil triggers a DB constraint" do
+    mou_signature.user = nil
+    expect { mou_signature.save(validate: false) }.to raise_error(ActiveRecord::StatementInvalid).with_message(/null value in column "user_id"/)
+    # expect { mou_signature.update_column(:user_id, nil) }.to raise_error(ActiveRecord::StatementInvalid).with_message(/null value in column "user_id"/)
```
DavidBiddle, line 17: Do we still need this commented out line?

app/models/user.rb
```diff
@@ -5,6 +5,7 @@ class User < ApplicationRecord
   class UserAuthenticationException < StandardError; end
 
   belongs_to :organisation, optional: true
+  has_many :mou_signatures
```
lfdebrux, line 8: Is it strictly necessary to have the `has_many` associations here in `user` and `organisation`? I'm not sure I can see a use case right now.

thomasiles: It's not necessary if users can't move between orgs. Currently they can if someone changes their organisation in the users screen.

I don't think we really have a use case where users have been around long enough to actually be in a position to sign an MOU in one department, move orgs and sign another one but its possible.

I went for this option here because it represents the db well and doesn't add much complexity. Is there a better way to structure the DB schema to not need it?

spec/models/mou_signature_spec.rb
thomasiles: We did not, good spot!

app/components/header_component/view.rb
```diff
@@ -2,14 +2,15 @@
 
 module HeaderComponent
   class View < ViewComponent::Base
-    attr_accessor :is_signed_in, :user_name, :user_profile_link, :signout_link, :list_of_users_path, :hosting_environment
+    attr_accessor :is_signed_in, :user_name, :user_profile_link, :signout_link, :list_of_users_path, :mou_path, :hosting_environment
 
-    def initialize(is_signed_in:, user_name:, user_profile_link:, signout_link:, list_of_users_path:, hosting_environment: HostingEnvironment)
+    def initialize(is_signed_in:, user_name:, user_profile_link:, signout_link:, list_of_users_path:, mou_path:, hosting_environment: HostingEnvironment)
```
aliuk2012, line 7: Is there anyway we could condense this?

if a method has a lot of params like this its usually a sign of a code smell.

spec/models/mou_signature_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      user:,
+      organisation:,
+      agreed_at: Time.zone.now,
+    )
+  end
+
+  it "setting user to nil triggers a DB constraint" do
+    mou_signature.user = nil
+    expect { mou_signature.save(validate: false) }.to raise_error(ActiveRecord::StatementInvalid).with_message(/null value in column "user_id"/)
+  end
```
aliuk2012, line 17: Could this not be a model validation to catch it before it gets to the DB?

app/models/mou_signature.rb
aliuk2012, line 1: I was expecting to see more validations here.
For example, can an agreed_at be in the future, could it be sometime in the past e.g. 10 years ago?

db/migrate/20231003140934_create_mou_signatures.rb
```diff
@@ -0,0 +1,31 @@
...
+                 comment: "User signatures of an MOU for an oragnisation" do |t|
+      t.references :user, null: false,
+                          foreign_key: true,
+                          comment: "User who signed MOU"
+
+      t.references :organisation, null: true,
+                                  foreign_key: true,
+                                  comment: "Organisation which user signed MOU on behalf of, or null"
+
+      t.datetime :agreed_at, null: false,
```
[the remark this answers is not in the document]
aliuk2012: 🤔  is it even needed? could you not just use created_at?

app/components/header_component/view.rb
thomasiles: yeah - I guess creating a HeaderComponentOptions object would be the answer?

spec/models/mou_signature_spec.rb
thomasiles: I've gone for DB constraints over model validations as they keep the DB in order and seem a bit more solid. I'm also reading https://sustainable-rails.com/ which recommends them over model validations :)

app/components/header_component/view.rb
aliuk2012: yeah potentially a service which returns these options would be better than trying to replicate and update all calls to the component everytime we add another param.

spec/models/mou_signature_spec.rb
aliuk2012: sure DB constraints are good but its better for Rails to handle those errors before it gets there. I would say its in addition rather than instead of DB. Also we have control over those errorr,
 if we migrate away from postgres the new db or Postgres changes the error message and returns a different error message meaning this test fails through no fault of ours.
 
 Could use  `belongs_to :user` in the model.

app/models/user.rb
aliuk2012: I feel Toms, implementation makes sense. It allows ActiveRecord to be aware of the relationship and optimize for it.
so if you wanted to pull out the last signed MOU you could use `user.mou_signatures.last` without having to do a separate query to search it up for the user.

db/migrate/20231003140934_create_mou_signatures.rb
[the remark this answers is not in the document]
thomasiles: I went for agreed_at as it felt part of the domain more than created_at and also covers the case where an MOU is was agreed in the past but just added to the DB. I could get rid and only have created_at though

app/policies/mou_signature_policy.rb
```diff
@@ -0,0 +1,12 @@
+class MouSignaturePolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
```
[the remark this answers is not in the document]
thomasiles: It's part of the pundit interface. I'll add an underscore and get rid of the reader to make it clearer that we are not using it https://github.com/varvet/pundit#headless-policies

db/migrate/20231003140934_create_mou_signatures.rb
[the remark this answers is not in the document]
aliuk2012: could setup `agreed_at` as an alias/method to reuse `created_at` .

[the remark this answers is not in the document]
jamie-o-wilkinson: That wouldn't achieve the same thing though would it? I was thinking of the case where an organisation agreed to an MOU before admin supported recording of signatures. We could then set `created_at` to be June 2023 for example.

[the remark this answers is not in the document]
lfdebrux: I vote for keeping `agreed_at`, as it's more obvious what it is about.

app/models/mou_signature.rb
aliuk2012: even if its being set in code, it might still be useful to make sure the code thats signing isn't passing something that we wouldn't accept

app/components/header_component/view.rb
thomasiles: I've refactored the header component so that links are passed in

db/migrate/20231003140934_create_mou_signatures.rb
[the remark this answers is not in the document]
thomasiles: I'm tempted to get rid of agreed_at so I don't have to write any validations or tests for it. Maybe adding an extra agreed_at was YAGNI and if we need to do anything fancy in the future we will just have to work things out then :)

spec/models/mou_signature_spec.rb
aliuk2012: Keep DB constraints but get rid of the tests and rely on belongs_to  in the model and have tests to make sure the model is invalid if a user is nil and you could check what error message is returned.

thomasiles: got rid of em 👍

app/service/navigation_items_service.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    navigation_items
+  end
+
+private
+
+  attr_reader :user
+
+  def user_provider
+    user&.provider&.to_sym
```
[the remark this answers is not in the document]
thomasiles: user.provider turned out to be nil the in the basic auth case, I've fixed that and removed them both

spec/models/mou_signature_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  it "has a valid factory" do
+    expect(mou_signature).to be_valid
+  end
+
+  it "is invalid without a user" do
+    error_message = I18n.t("errors.messages.required")
+
+    mou_signature.user = nil
+    expect(mou_signature).not_to be_valid
```
aliuk2012, line 14:
```suggestion
    expect(mou_signature).to be_invalid
```
I find it easier to read and it matches the spec description of "it is invalid without a user"

spec/models/mou_signature_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      organisation:,
+      agreed_at: Time.zone.now,
+    )
+  end
+
+  it "setting user to nil triggers a DB constraint" do
+    mou_signature.user = nil
+    expect { mou_signature.save(validate: false) }.to raise_error(ActiveRecord::StatementInvalid).with_message(/null value in column "user_id"/)
+    # expect { mou_signature.update_column(:user_id, nil) }.to raise_error(ActiveRecord::StatementInvalid).with_message(/null value in column "user_id"/)
+  end
```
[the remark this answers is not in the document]
thomasiles: agreed_at no longer exists, if we add it again later we can consider it :)

thomasiles:
> Thanks for addressing my comments and doing some nice refactoring on the components.

Thanks for taking the time to review it and offer suggestions!

spec/models/mou_signature_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  let(:user) { create(:user, organisation:) }
+  let(:mou_signature) do
+    described_class.create!(
+      user:,
+      organisation:,
+      agreed_at: Time.zone.now,
+    )
+  end
+
+  it "setting user to nil triggers a DB constraint" do
```
chao-xian: Nitpick: these specs are supposed to read something like:

`it "sets user to nil and triggers a DB constraint"` 

or

```
context "setting the user to nil" do
  it "triggers a DB constraint" do
    ...
  end
end
```

one to be aware of for future

spec/models/mou_signature_spec.rb
```diff
@@ -14,7 +14,6 @@
   it "setting user to nil triggers a DB constraint" do
     mou_signature.user = nil
     expect { mou_signature.save(validate: false) }.to raise_error(ActiveRecord::StatementInvalid).with_message(/null value in column "user_id"/)
-    # expect { mou_signature.update_column(:user_id, nil) }.to raise_error(ActiveRecord::StatementInvalid).with_message(/null value in column "user_id"/)
```
chao-xian: This commit should have been fixed up into a prior commit

PR #688, opened 2023-10-17
Suppress IE8 warning

DavidBiddle:

Trello card: none

This message is appearing because ie8 support is deprecated in govuk-frontend. We don't support Internet Explorer 8 anyway, so we can suppress this warning.

Test this by running `bin/setup` or `bin/vite build` and observing that the FE build no longer gives us the following warning: 
> The govuk-if-ie8 mixin is deprecated and will be removed in v5.0. To silence this warning, update $govuk-suppressed-warnings with key: "ie8"


thomasiles (approved):
nice

PR #690, opened 2023-10-17
Save a copy of session data to DraftQuestion

aliuk2012:

Trello card: 


app/forms/pages/question_form.rb
```diff
@@ -1,13 +1,24 @@
...
+  validates :draft_question, presence: true
   validates :hint_text, length: { maximum: 500 }
+
+  def submit
+    return false if invalid?
+
+    draft_question.question_text = question_text
+    draft_question.hint_text = hint_text
+    draft_question.is_optional = is_optional
+    draft_question.question_text = question_text
```
DavidBiddle, line 21: Could you do this with `merge` (or some other syntax) to make this more compact and less repetitive? I'm thinking maybe something like:
```suggestion
    draft_question.merge({ question_text:, hint_text:, is_optional:, question_text: })
```

spec/requests/pages/questions_controller_spec.rb
```diff
@@ -22,7 +22,7 @@
     login_as_editor_user
   end
 
-  describe "#new" do
+  describe "#create" do
```
DavidBiddle, line 25: Good spot on this being misnamed. Feels like there should be a test for `#new` in here as well, but that doesn't have to be part of this PR.

DavidBiddle (approved):
Raised a couple of comments, both optional. Really good refactor!

app/forms/pages/question_form.rb
aliuk2012: Thanks for the suggestion, I'll make the changes.....interesting that both of us didn't catch the code setting `question_text` twice

DavidBiddle: Oh yeah 😅

aliuk2012: @DavidBiddle I've pushed a commit which should make it easier to read without all the repetition

spec/requests/pages/questions_controller_spec.rb
aliuk2012: This is fairly large already. I've added a tech debt ticket here https://trello.com/c/l7OUqJEN/1223-tech-debt-add-missing-request-spec-for-questioncontroller-new-action

DavidBiddle (dismissed):

DavidBiddle (approved):

app/controllers/pages_controller.rb
```diff
@@ -52,4 +60,14 @@ def answer_types
   def convert_session_keys_to_symbols
     session[:page].deep_symbolize_keys! if session[:page].present?
   end
+
+  def setup_draft_question_for_existing_page
```
chao-xian, line 64: For a future iteration - can we introduce the factory pattern to address setting up our objects outside of our controllers?

app/forms/pages/address_settings_form.rb
```diff
@@ -20,7 +20,9 @@ def submit(session)
     }
 
     # Set answer_settings for the draft_question
-    draft_question.answer_settings = answer_settings.with_indifferent_access
+    draft_question
+      .assign_attributes({ answer_settings: answer_settings.with_indifferent_access })
```
chao-xian, line 24: Huh... I would have expected the linter to ask for:
```
draft_question.assign_attributes(
   { 
    answer_settings: answer_settings.with_indifferent_access 
   }
)
```

(not that it matters)

chao-xian:
Really nice work @aliuk2012 ! TYSM for putting in so much work on this.

PR #695, opened 2023-10-20
Update method name from update to add prefix

chao-xian:
This method only adds orgs to MouSignatures that don't have one. The name mislead me before.

### DIFF
--- a/app/models/mou_signature.rb
+++ b/app/models/mou_signature.rb
@@ -2,7 +2,7 @@ class MouSignature < ApplicationRecord
   belongs_to :user
   belongs_to :organisation, optional: true
 
-  def self.update_user_orgs(user)
+  def self.add_user_orgs(user)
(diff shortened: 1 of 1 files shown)


DavidBiddle (approved):
Seems reaasonable to me, but might benefit from a @alphagov/forms-feature-team-1 dev taking a look

chao-xian:
Thanks @DavidBiddle! Don't want to merge this before having your thoughts @thomasiles

aliuk2012:
🤔  a bit odd that you could rename the method and no tests fail.

aliuk2012:
@chao-xian @thomasiles this method name has changed since this PR but could I propose an alternative way we could have implemented this? 

af086136e3397f4e9cdb2356f8bcaac0ebca1acb

chao-xian:
The spirit of my PR which was to rename I think has been resolved elsewhere so I'll close this PR and let any relevant discussion continue in Al's refactoring PR.

PR #696, opened 2023-10-20
Expire sessions after 24 hours using gds-sso config

jamie-o-wilkinson:
For security reasons (See [OWASP recommendations](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#session-expiration)), we’d like to ensure forms-admin users log-in at least every 24 hours. To achieve this, we need to make sure that forms-admin re-authenticates with the current auth provider (in this case Auth0) after a given time.

This PR configures auth_valid_for for the gds-sso to require re-authentication after 24 hours. This acts as an absolute session timeout regardless of activity.

Trello card: https://trello.com/c/yi3v9Mgo


lfdebrux (approved):
Nice!

PR #697, opened 2023-10-23
Remove obsolete table called schema_info & session

aliuk2012:

Remove obsolete db tables from db/schema

@DavidBiddle could you check your local instance to see if you have `schema_info`, because it was added over 4 months ago but it could be that it has been removed since.

@thomasiles could you check your local instance and make sure it does not include `sessions` table. 

Trello card: 


lfdebrux (approved):

lfdebrux:
I agree with these changes, I dropped my database and ran all the migrations through, and the resultant schema file was missing these two tables.

PR #699, opened 2023-10-23
Specify auth0 connection

jamie-o-wilkinson:
The end-to-end tests currently login to the admin service through auth0, which sends an email with a code. The tests then check the email and find the code which is used to login.

This method works, but may be unreliable given the number of deploys we make in production.

This PR defaults forms-admin to using the passwordless "email" authentication flow, but also allows use of the username/password flow by end-to-end tests via specification of the `force_credentials_flow` query parameter.

Trello card: https://trello.com/c/no2hfOty

### DIFF
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -9,7 +9,19 @@ class AuthenticationController < ApplicationController
 
   def redirect_to_omniauth
     store_location(attempted_path) if request.get?
-    redirect_to "/auth/#{default_provider}"
+
+    params = {}
+    if default_provider == "auth0" && e2e_user?
+      # the value passed in connection must match the name given in the auth0 terraform
+      params[:connection] = "Username-Password-Authentication"
+    end
+
+    redirect_to ActionDispatch::Http::URL.path_for(path: "/auth/#{default_provider}", params:)
+  end
+
+  def e2e_user?
+    # this should be appended to the first request in the e2e tests to activate the username/password flow
+    request.params[:auth] == "e2e"
   end
 
   def callback_from_omniauth
--- a/config/initializers/authentication.rb
+++ b/config/initializers/authentication.rb
@@ -9,6 +9,7 @@ Rails.application.config.before_initialize do
     callback_path: "/auth/auth0/callback",
     authorize_params: {
       scope: "openid email",
+      connection: "email", # default to using the passwordless flow
     },
   )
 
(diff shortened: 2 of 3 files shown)


app/controllers/authentication_controller.rb
```diff
@@ -9,7 +9,17 @@ def self.call(env)
...
+      # the value passed in connection must match the name given in the auth0 terraform
+      redirect_to "/auth/#{default_provider}?connection=Username-Password-Authentication"
+    else
+      redirect_to "/auth/#{default_provider}"
+    end
+  end
+
+  def e2e_user?
+    # this is appended to the first request in the e2e tests to activate the username/password flow
+    request.params.key? "force_credentials_flow"
```
lfdebrux, line 22: I think maybe we should go with a more general parameter name/value combination, to give us more flexibility in future. Something like `?auth=e2e`?

```suggestion
    request.params[:auth] == "e2e"
```

app/controllers/authentication_controller.rb
```diff
@@ -9,7 +9,17 @@ def self.call(env)
 
   def redirect_to_omniauth
     store_location(attempted_path) if request.get?
-    redirect_to "/auth/#{default_provider}"
+    if default_provider == "auth0" && e2e_user?
+      # the value passed in connection must match the name given in the auth0 terraform
+      redirect_to "/auth/#{default_provider}?connection=Username-Password-Authentication"
+    else
+      redirect_to "/auth/#{default_provider}"
+    end
```
lfdebrux, line 17: `redirect_to` doesn't stop the execution of a function [[1](https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to)], so the way this is written makes it easier for us to accidentally do a double redirect in future.

What about building up a params variable and then constructing a URL before handing off to `redirect_to`?

```suggestion
    params = {}
    if default_provider == "auth0" && e2e_user?
      # the value passed in connection must match the name given in the auth0 terraform
      params[:connection] = Username-Password-Authentication"
    end
    
    redirect_to ActionDispatch::Http::URL.path_for(path: "/auth/#{default_provider}", params:)
```

app/controllers/authentication_controller.rb
jamie-o-wilkinson: I've made this change now, thanks for the suggestion!

app/controllers/authentication_controller.rb
jamie-o-wilkinson: I've made this change now, thanks for the suggestion!

lfdebrux (approved):
Works for me!

PR #702, opened 2023-10-24
Use default Rails logger for development

lfdebrux:

Our default logger, [lograge], outputs structured log messages that are optimised for machine readability, rather than human readability.

This commit switches the development environment to using the default Rails logging configuration. If a developer wants to see what the production logging looks like they can comment these lines out, or change the RAILS_ENV environment variable.

This also means we can be aware of what we are missing in our lograge configuration that we could be getting from Rails.

[lograge]: https://github.com/roidrage/lograge

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -67,4 +67,8 @@ Rails.application.configure do
 
   # Uncomment if you wish to allow Action Cable access from any origin.
   # config.action_cable.disable_request_forgery_protection = true
+
+  # Use default Rails logging
+  config.log_formatter = ::Logger::Formatter.new
+  config.lograge.enabled = false
 end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -67,8 +67,4 @@ Rails.application.configure do
 
   # Uncomment if you wish to allow Action Cable access from any origin.
   # config.action_cable.disable_request_forgery_protection = true
-
-  # Use default Rails logging
-  config.log_formatter = ::Logger::Formatter.new
-  config.lograge.enabled = false
 end
--- a/config/initializers/logging.rb
+++ b/config/initializers/logging.rb
@@ -12,7 +12,6 @@ unless Rails.env.development?
     if ENV["RAILS_LOG_TO_STDOUT"].present?
       config.logger = ActiveSupport::Logger.new($stdout)
       config.logger.formatter = config.log_formatter
-
     end
 
     # Lograge is used to format the standard HTTP request logging
@@ -27,7 +26,6 @@ unless Rails.env.development?
     config.lograge.custom_options = lambda do |event|
       {}.tap do |h|
         h[:host] = event.payload[:host]
-        h[:user_id] = event.payload[:user_id]
         h[:user_email] = event.payload[:user_email]
         h[:user_organisation_slug] = event.payload[:user_organisation_slug]
         h[:user_ip] = event.payload[:user_ip]
--- a/config/initializers/logging.rb
+++ b/config/initializers/logging.rb
@@ -1,0 +1,42 @@
+require "./app/lib/json_log_formatter"
+
+unless Rails.env.development?
+  Rails.application.configure do
+    ### LOGGING CONFIGURATION ###
+    config.log_level = :info
+
+    # Use JSON log formatter for better support in Splunk. To use conventional
+    # logging use the Logger::Formatter.new.
+    config.log_formatter = JsonLogFormatter.new
+
+    if ENV["RAILS_LOG_TO_STDOUT"].present?
+      config.logger = ActiveSupport::Logger.new($stdout)
+      config.logger.formatter = config.log_formatter
+
+    end
+
+    # Lograge is used to format the standard HTTP request logging
+    config.lograge.enabled = true
+    config.lograge.formatter = Lograge::Formatters::Json.new
+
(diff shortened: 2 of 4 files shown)


config/environments/development.rb
```diff
@@ -67,4 +67,8 @@
 
   # Uncomment if you wish to allow Action Cable access from any origin.
   # config.action_cable.disable_request_forgery_protection = true
+
+  # Use default Rails logging
+  config.log_formatter = ::Logger::Formatter.new
+  config.lograge.enabled = false
```
aliuk2012, line 73: I agree the change is needed, I'm just thinking that the default should always be Rails Logger rather than the exception. So instead of declaring the rails logger here, we should only declare and set Lograge in production as that is the exception

aliuk2012: But instead of all the config in application.rb could we not just move it to production.rb and then if you want to test it you can run rails in production mode to see what the logs will output?

lfdebrux: I meant test in specs, if we have the logging configuration in `application.rb` then we don't need to duplicate it between `environments/production.rb` and `environments/test.rb`.

danworth: Overriding the log formatter in the dev config was how I'd anticipated it being managed if the json output wasn't clear enough for local dev where the logs are viewed without json parsing power. I don't think having to run the apps in Production mode is viable because then you cannot run the all services together since things like auth become necessary AFAIK; and running them together was and will be necessary to check logs in different scenarios. I think there's also advantage to the default being the behaviour we have in production (where practical) because it makes replicating production behaviour simpler. I originally made the change because when I did the JSON work I seem to remember it being more onerous than necessary to get the logs as per prod locally. 

The change as Laurence has it seems simple to toggle local dev between json or std 👍  (I'm not actually working this morning because I'm still unwell)

lfdebrux: Here's an alternative solution; we move all the logging related code to a new initializer in `config/initializers/logging.rb`, and wrap everything in an `unless Rails.env.dev?` block...

```
# config/initializers/logging.rb

unless Rails.env.development?
  ### LOGGING CONFIGURATION ###
...
end
```

then we don't need to change `environments/development.rb` to get the Rails defaults, but we get the centralised configuration.

How does that sound as a compromise?

DavidBiddle: That seems like a pretty neat solution to me (but I don't mind the current implementation either)

aliuk2012: Yeah that sounds like a good alternative.  We can also add some unit tests for that initializer to make sure its doing what we expect it to do.

danworth: I preferred your original implementation using the `config/environments/*.rb` because to me it keeps environmental differences in limited known places and seems like that's what they're intended for. With the suggested alternative and similar types of switches based upon `Rails.env.development` throughout our codebase I think it'll become harder to see whats being controlled where and what the substantial differences are between environments. Although this is a simple and rather innocuous change to our logging it seems this discussion is more about the pattern and one which we might follow in future.

However, if the new suggestion is the least surprising and simple to understand and modify for the majority then I won't stand in the way 😄

chao-xian: I've decided we should go with the second option as to me it's clearer because we know it only applies to non-production like env. We should add the test as Al suggests. We already have switches based on env vars in some of the other Rails configs. Looking at each env config to look for how the logging is configured to me is more confusing that seeing a file at a level above named `logging.rb`. Each env file can still override that configuration I believe.

danworth: Sounds good and thanks for giving time to the conversation and arriving at a decision 👍 . I think personal preference comes down to which end of the stick you're looking at (and no doubt prior Rails experience). I can see this works for finding configuration differences of a particular component, in this case logging, by looking at its initializer. As someone who's been spending more time in Terraform (and less Rails) my focus was more on making it simpler to find the broader differences between development and the default (which ideally aligns with production) by just looking at `config/environments/development.rb`.

Onwards and thanks for making this improvement Laurence! 🚀

lfdebrux: Okay, I've pushed a commit that moves the logging code to a separate file, and adds tests. Please re-review all!

config/initializers/logging.rb
```diff
@@ -0,0 +1,40 @@
...
+        h[:host] = event.payload[:host]
+        h[:user_email] = event.payload[:user_email]
+        h[:user_organisation_slug] = event.payload[:user_organisation_slug]
+        h[:user_ip] = event.payload[:user_ip]
+        h[:request_id] = event.payload[:request_id]
+        h[:user_id] = event.payload[:user_id]
+        h[:form_id] = event.payload[:form_id] if event.payload[:form_id]
+        h[:page_id] = event.payload[:page_id] if event.payload[:page_id]
+        h[:exception] = event.payload[:exception] if event.payload[:exception]
+      end
```
aliuk2012, line 37:
```suggestion
      {}.tap do |h|
        %i[host user_email user_organisation_slug user_ip request_id user_id form_id page_id exception].each do |key|
          h[key] = event.payload[key] if event.payload[key]
        end
      end
```

I was wondering if this is a better way of writing this, we could even declare that list of keys in a const

lfdebrux: This change makes sense to me, I wonder if we could do it in a separate PR though? I personally think it's better to keep PRs focused on one thing.

aliuk2012 (approved):

PR #705, opened 2023-10-26
Correct message about URL staying the same

hannahkc:

Trello card:https://trello.com/c/fJiGg7G8/708-make-your-changes-live-page-is-inaccurate-on-a-draft-of-a-live-form

When you make the draft of a live form live, you are told: “The URL for the live form will remain the same.“

However, the URL may be different if the form name has changed. Links to the form will still work in this case, as they rely on the first part of the URL which never changes.

So we are changing this line to be more accurate and specific - to make sure people don’t worry unnecessarily about links to their form breaking.


aliuk2012 (approved):
Makes sense

PR #712, opened 2023-10-30
Add Pentest Partners temporarily

chao-xian:

Trello card: https://trello.com/c/VzgZIpIA/1129-allow-forms-to-be-submitted-to-pen-test-partners-in-staging-environment-for-ithc

Pentest Partners will need to be able to create forms and set an email address at their domain to receiver form submissions. This temporarily adds their email domain.

This has a dependency on the feature flag being set in forms-deploy...

### DIFF
--- a/app/forms/forms/submission_email_form.rb
+++ b/app/forms/forms/submission_email_form.rb
@@ -2,16 +2,10 @@ class Forms::SubmissionEmailForm < BaseForm
   attr_accessor :form, :temporary_submission_email, :email_code, :confirmation_code, :current_user, :notify_response_id
 
   EMAIL_REGEX = /.*@.*/
-  # TODO: remove after Friday 3 Nov 2023
-  GOVUK_EMAIL_REGEX = if FeatureService.enabled?(:pentest_partners_access_enabled)
-                        /(\.gov\.uk|@pentestpartners\.com)\z/i
-                      else
-                        /\.gov\.uk\z/i
-                      end
 
   validates :temporary_submission_email, presence: true
   validates :temporary_submission_email, format: { with: EMAIL_REGEX, message: :invalid_email }
-  validates :temporary_submission_email, format: { with: GOVUK_EMAIL_REGEX, message: :non_govuk_email }
+  validates :temporary_submission_email, govuk_email: true
 
   EMAIL_CODE_REGEX = /[0-9]{6}/
   validates :email_code, format: { with: EMAIL_CODE_REGEX, message: :invalid_email_code }, if: -> { email_code.present? }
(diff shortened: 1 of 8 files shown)


app/forms/forms/submission_email_form.rb
```diff
@@ -2,7 +2,7 @@ class Forms::SubmissionEmailForm < BaseForm
   attr_accessor :form, :temporary_submission_email, :email_code, :confirmation_code, :current_user, :notify_response_id
 
   EMAIL_REGEX = /.*@.*/
-  GOVUK_EMAIL_REGEX = /\.gov\.uk\z/i
+  GOVUK_EMAIL_REGEX = /(\.gov\.uk|pentestpartners\.com)\z/i # TODO: remove after Friday 3 Nov 2023
```
aliuk2012, line 5: Its a pity we couldn't use an env var to control access for both authentication and for this feature.

aliuk2012 (dismissed):

app/forms/forms/submission_email_form.rb
chao-xian: hmmm that's a good point... I'll rework this

DavidBiddle, line 5: Think I've spotted a potential issue here - someone could pass it with a sneaky domain purchase I think: `hacker@notpentestpartners.com`. This also applies to the existing validation, though the subdomain makes it slightly more convoluted (`hacker@mysubdomain.gov.uk.ihackwebsites.biz`).

chao-xian: 🤦

chao-xian: for the existing one the `\z` means it has to end with `.gov.uk` but that's a really good point about the `notpentesters.com` example.

chao-xian: We've done it a hacky way for now...

chao-xian: Have added a spec to guard against this.

config/settings/test.yml
```diff
@@ -1,3 +1,6 @@
+features:
+  pentest_partners_access_enabled: true
```
aliuk2012, line 2: The problem with setting this here is that https://github.com/alphagov/forms-admin/pull/712/files#diff-a66e72dce6f86faae2ba62aa5dd129e12ced85afd75419bf4668b71d0344236aR9 will never get exercised during the tests. I tested this locally by putting a breakpoint in the else block and not once did it pause the tests from running

DavidBiddle (approved):
Nice - like the custom validator approach!

chao-xian:
@aliuk2012 amazing! Love this approach too ❤️

PR #724, opened 2023-10-31
Add support link to header

DavidBiddle:

Trello card: https://trello.com/c/SD4Akxlt/1160-add-a-link-to-the-support-form-from-forms-admin

Adds a new header navigation item linking to the support page in forms-product-page. This relies on a new setting which will be configured in each environment as an environment variable. 

Also includes two small fixes:
- a fix for a warning we were seeing in the logs (see this [Slack thread](https://gds.slack.com/archives/C02QNDA5YBC/p1698418458463679) and the ['Add default value for classes parameter' commit message](https://github.com/alphagov/forms-admin/pull/724/commits/ddce147eeda726045b241a66006c9423923a1f7b) for more details). This was introduced in a recent refactor and doesn't have any negative impact on the service, but was causing noise in the logs (since the header component is loaded on every page).
- a fix for a bug in our automated accessibility tests, where we were testing pages we weren't intending to, which caused a failure in this PR. See the ['Make accessibility tests more targeted' commit message](https://github.com/alphagov/forms-admin/pull/724/commits/139378453a555ce275006eec658aea0a516b2c1b) for more details.

Note: Don't merge this until the corresponding forms-deploy PR has gone in (https://github.com/alphagov/forms-deploy/pull/388)

### Screenshot
The menu should now include a 'Support' link:
The GOV.UK Forms header with (from left to right) the crown logo, GOV.UK wordmark, local tag, and then menu items for 'MOUs', 'Users', 'Support', User's name ('A User' in this case), and 'Sign out'

### DIFF
--- a/app/service/navigation_items_service.rb
+++ b/app/service/navigation_items_service.rb
@@ -1,5 +1,9 @@
 class NavigationItemsService
-  NavigationItem = Struct.new(:text, :href, :active, :classes)
+  NavigationItem = Struct.new(:text, :href, :active, :classes) do
+    def initialize(text:, href:, active:, classes: [])
+      super(text, href, active, classes)
+    end
+  end
 
   include Rails.application.routes.url_helpers
 
--- a/app/service/navigation_items_service.rb
+++ b/app/service/navigation_items_service.rb
@@ -19,6 +19,7 @@ class NavigationItemsService
     navigation_items = []
     navigation_items << NavigationItem.new(text: I18n.t("header.mous"), href: mou_signatures_path, active: false) if should_show_mous_link?
     navigation_items << NavigationItem.new(text: I18n.t("header.users"), href: users_path, active: false) if should_show_user_profile_link?
+    navigation_items << NavigationItem.new(text: I18n.t("header.support"), href: Settings.forms_product_page.support_url, active: false) if Settings.forms_product_page.support_url.present?
     navigation_items << NavigationItem.new(text: user.name, href: user_profile_url, active: false) if user.name.present?
     navigation_items << NavigationItem.new(text: I18n.t("header.sign_out"), href: signout_url, active: false) if signout_url.present?
 
--- a/app/service/navigation_items_service.rb
+++ b/app/service/navigation_items_service.rb
@@ -20,19 +20,50 @@ class NavigationItemsService
   def navigation_items
     return [] if user.blank?
 
-    navigation_items = []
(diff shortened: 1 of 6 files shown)


app/service/navigation_items_service.rb
```diff
@@ -19,6 +23,7 @@ def navigation_items
     navigation_items = []
     navigation_items << NavigationItem.new(text: I18n.t("header.mous"), href: mou_signatures_path, active: false) if should_show_mous_link?
     navigation_items << NavigationItem.new(text: I18n.t("header.users"), href: users_path, active: false) if should_show_user_profile_link?
```
aliuk2012, line 25: This is starting to feel rather busy and hard to read. 

Could you maybe refactor this method so it sort of follows the style we used here https://github.com/alphagov/forms-runner/blob/main/app/services/form_submission_service.rb#L133-L137 by making each item a private method with a guard to return nil if it shouldn't display  the nav. 

🤔  It feels like we should be using pundit policy to control whether or not a user should see the various links.

app/service/navigation_items_service.rb
```diff
@@ -1,5 +1,9 @@
 class NavigationItemsService
-  NavigationItem = Struct.new(:text, :href, :active, :classes)
+  NavigationItem = Struct.new(:text, :href, :active, :classes) do
+    def initialize(text:, href:, active:, classes: [])
+      super(text, href, active, classes)
+    end
```
lfdebrux, line 5: Nice! Thanks for fixing those errors, it was really bothering me 😅

app/service/navigation_items_service.rb
```diff
@@ -16,19 +20,51 @@ def initialize(user:)
...
+    navigation_items = [
+      mou_navigation_item,
+      users_navigation_item,
+      support_navigation_item,
+      profile_navigation_item,
+      signout_navigation_item,
+    ]
 
-    navigation_items
+    navigation_items.compact
```
aliuk2012, line 31:
```suggestion
    navigation_items.compact_blank
```

app/service/navigation_items_service.rb
aliuk2012: Thanks this is loads better and easy to read. It is also easier to see what order each item would be in

DavidBiddle: Yeah, good suggestion - it's a defintie improvement!

aliuk2012 (approved):

PR #725, opened 2023-10-31
Replace Timecop with native Rails time helpers for specs

aliuk2012:
Theres no need for Timecop, we do not  use it in any of our other apps and we can rely on rails native time helpers to achieve the same result

Trello card: 


lfdebrux (approved):
Nice!

PR #726, opened 2023-11-01
Add failure page with correct retry link

lfdebrux:

Trello card: https://trello.com/c/eHzFLLjJ/1155-fix-auth0-failure-page

Currently we're using the failure page from gds-sso gem [[1]] if there is an issue when authenticating with OmniAuth. However, the redirect link in that failure page is hardcoded to send users to GOV.UK Signon, which isn't always what we want.

This commit copies the failure page into our own app so we can edit it, and makes the link dependant on the `strategy` parameter from OmniAuth (which should match the provider, at least in the scenarios we care about).

[1]: https://github.com/alphagov/gds-sso/blob/3aa846de2d71c1f90c120546ade3595e0883c60c/app/views/authentications/failure.html.erb

<details><summary><h4>Screenshots</h4></summary>

##### Before

Screenshot of https://admin.staging.forms.service.gov.uk/auth/failure?message=invalid_credentials&strategy=auth0, the link 'try again' points to /auth/gds

##### After

Screenshot of http://localhost:3000/auth/failure?message=invalid_credentials&strategy=auth0, the link 'try again' points to /auth/auth0

</details>

### DIFF
--- a/app/views/authentications/failure.html.erb
+++ b/app/views/authentications/failure.html.erb
@@ -1,0 +1,3 @@
+<h1>Error</h1>
+
+<p>There was a problem logging you into the application. Please <%= link_to "try again", "/auth/#{params[:strategy]}" %>.</p>
--- a/spec/integration/gds_sso_spec.rb
+++ b/spec/integration/gds_sso_spec.rb
@@ -5,6 +5,11 @@ RSpec.describe "usage of gds-sso gem" do
     allow(Settings).to receive(:auth_provider).and_return("gds_sso")
   end
 
+  after do
+    OmniAuth.config.mock_auth[:gds] = nil
+    OmniAuth.config.test_mode = false
+  end
+
   let(:omniauth_hash) do
     OmniAuth::AuthHash.new({
       provider: "gds",
@@ -27,11 +32,6 @@ RSpec.describe "usage of gds-sso gem" do
   describe "authentication" do
     before do
       OmniAuth.config.test_mode = true
-      OmniAuth.config.mock_auth[:gds] = nil
-    end
-
-    after do
-      OmniAuth.config.test_mode = false
     end
 
     it "redirects to OmniAuth when no user is logged in" do
@@ -89,6 +89,31 @@ RSpec.describe "usage of gds-sso gem" do
 
         expect(gds_sso.successful?).to be true
       end
+    end
+  end
+
+  describe "failure page" do
+    it "is shown if there is an authentication failure with external provider" do
+      OmniAuth.config.test_mode = true
+      OmniAuth.config.mock_auth[:gds] = :invalid_credentials
+
+      logout
+
+      get root_path
+
+      expect(response).to redirect_to "/auth/gds"
+      follow_redirect!
+
+      expect(response).to redirect_to "/auth/gds/callback"
+      follow_redirect!
+
+      expect(response).to redirect_to "/auth/failure?message=invalid_credentials&strategy=gds"
+    end
+
+    it "has a retry link" do
+      get "/auth/failure?message=invalid_credentials&strategy=gds"
+
+      expect(response.body).to include '<a href="/auth/gds">try again</a>'
     end
   end
 end
--- a/spec/integration/auth0_spec.rb
+++ b/spec/integration/auth0_spec.rb
@@ -1,22 +1,22 @@
 require "rails_helper"
 
 RSpec.describe "usage of omniauth-auth0 gem" do
+  before do
+    allow(Settings).to receive(:auth_provider).and_return("auth0")
+
+    OmniAuth.config.test_mode = true
+  end
+
+  after do
+    OmniAuth.config.mock_auth[:auth0] = nil
+    OmniAuth.config.test_mode = false
+  end
+
   let(:omniauth_hash) do
     Faker::Omniauth.auth0(
       uid: "123456",
       email: "test@example.com",
     )
-  end
-
-  before do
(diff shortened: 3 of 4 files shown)


app/views/authentications/failure.html.erb
```diff
@@ -0,0 +1,3 @@
+<h1>Error</h1>
+
+<p>There was a problem logging you into the application. Please <%= link_to "try again", "/auth/#{params[:strategy]}" %>.</p>
```
aliuk2012, line 3: Could this copy be moved to locales? 

Is the page styled correctly? Doesn't the heading level need css class to style it? I was just thinking that we should use `govuk_link_to` (https://govuk-components.netlify.app/helpers/link/) instead of `link_to`

lfdebrux: The page has no styling currently (either in main or in this branch), I thought for quickness I'd leave the page pretty much as is.

lfdebrux: Ideally we'd be using the proper error layout and translations and all that, but I think that might need to involve changes to the routes and controllers?

aliuk2012: Yeah that makes perfect sense. I just didn't have context of what the existing page looked like. 

It might be worth adding to the PR description that its unstyled or screenshot for future devs looking at this PR.

spec/integration/gds_sso_spec.rb
```diff
@@ -5,6 +5,10 @@
     allow(Settings).to receive(:auth_provider).and_return("gds_sso")
   end
 
+  after do
+    OmniAuth.config.test_mode = false
+  end
```
aliuk2012, line 10:
```suggestion
	let(:test_mode) { false }

  after do
    OmniAuth.config.test_mode = test_mode
  end
```

We can then set the test mode to true when we need it

spec/integration/gds_sso_spec.rb
```diff
@@ -91,4 +91,29 @@
...
+
+      expect(response).to redirect_to "/auth/failure?message=invalid_credentials&strategy=gds"
+    end
+
+    it "has a retry link" do
+      get "/auth/failure?message=invalid_credentials&strategy=gds"
+
+      expect(response.body).to include '<a href="/auth/gds">try again</a>'
+    end
+  end
```
aliuk2012, line 118:
```suggestion
  describe "when there is a failure during authentication" do
  	let(:test_mode) { true }
  	let(:login_path) { "/auth/gds" }
  	let(:login_callback_path) { "/auth/gds/callback" }
  	let(:invalid_creds_path) { "/auth/failure?message=invalid_credentials&strategy=gds" }
  	
    it "is shown if there is an authentication failure with external provider" do
      OmniAuth.config.mock_auth[:gds] = :invalid_credentials

      logout

      get root_path

      expect(response).to redirect_to login_path
      follow_redirect!

      expect(response).to redirect_to login_callback_path
      follow_redirect!

      expect(response).to redirect_to invalid_creds_path
    end

    it "has a retry link" do
      get invalid_creds_path

      expect(response.body).to include "<a href='#{login_path}'>try again</a>"
    end
  end
```

spec/integration/gds_sso_spec.rb
lfdebrux: We don't want it to be `true` when the tests in the scenario are done, because that could interfere with tests in other scenarios. Do you mean changing it for the `before` block?

spec/integration/gds_sso_spec.rb
lfdebrux: I'm not sure I see the advantage of this, if we're not going to be changing the values of the `let`s.

aliuk2012: The benefit is that if we ever change those values, all the tests will fail but you only need to update 1 single line rather than doing find/replace and having multiple line changes.

It might also just be me but I find it easier to understand what the link is and what its for. Kind of like using magic numbers in code but its essentially a random string thats looks like some url.

app/views/authentications/failure.html.erb
lfdebrux: I've added screenshots to the description

spec/integration/gds_sso_spec.rb
lfdebrux: I see. I think if we wanted to make these changes it would make sense to do it as a bigger refactor, that affects the whole file, rather than just this short example group.

lfdebrux: @aliuk2012 is there a way to assign to a `let` variable without having to create a whole `describe` group?

aliuk2012: @lfdebrux lets chat about this tomorrow when we pair on the trial user specs. I'm thinking why wouldn't you have a describe/context block to call out a specific state that something might be in.

lfdebrux:
@aliuk2012 I'm going to resolve your comments without addressing them, as I think we need to get this PR merged ASAP. I see your point about test structure, but I'm struggling to put it into practice, and I don't have much time to spare on it.

In any case, I'm now of the opinion that the main changes in this PR are sub-optimal, for reasons you've already mentioned; I think it would be better if we just let OmniAuth raise the failure as an exception, and then handle it through the usual `rescue_fron` mechanism. I've been working on a separate branch to do that, but have been having issues writing unit tests for it, and so again for reasons of time I think we should go with these changes for now and revisit later

aliuk2012 (approved):

PR #727, opened 2023-11-01
normalise fetching current_form and referring to it.

aliuk2012:

There were two controller methods being used to fetch the current form

- `fetch_form` method in the PageController which set `@form` and wasn't memorised. Also called using a before_action in the controller
- `current_form` method in FormController which set `@current_form` and was normalised

Now we should be able to use `current_form` method or instance variable wherever we would have used @form. There are some instances where `@form` is used but those are specifically in services or other libs not in controllers/view layers.

Trello card: 


app/controllers/forms_controller.rb
```diff
@@ -25,20 +24,17 @@ def mark_pages_section_completed
...
+      @forms = policy_scope(Form) || []
+      render "pages/index", locals: { current_form: }, status: :unprocessable_entity
     end
   end
 
 private
 
-  def current_form
-    @current_form ||= Form.find(params[:form_id])
-  end
```
aliuk2012, line 39: This was moved to application.rb so it could be used in any controller we would like. In this instance its being reused in PagesController and all the various Pages::XXXXXControllers

app/controllers/pages_controller.rb
```diff
@@ -26,22 +27,18 @@ def move_page
 private
 
   def check_user_has_permission
-    authorize @form, :can_view_form?
-  end
-
-  def fetch_form
-    @form = Form.find(params[:form_id])
```
aliuk2012, line 33: This wasn't needed because we already had `current_form` method

app/controllers/pages_controller.rb
```diff
@@ -1,12 +1,13 @@
 class PagesController < ApplicationController
-  before_action :fetch_form, :convert_session_keys_to_symbols
```
aliuk2012, line 2: `fetch_form` has been removed but it didn't need to be added to the before action it it was a proper memorised method like `current_form` which creates the instance variable the first time its called rather than always calling it even if it is not used.

app/controllers/application_controller.rb
```diff
@@ -110,4 +110,8 @@ def non_maintenance_bypass_ip_address?
     bypass_ip_list = bypass_ips.split(",").map { |ip| IPAddr.new ip.strip }
     bypass_ip_list.none? { |ip| ip.include?(user_ip(request.env.fetch("HTTP_X_FORWARDED_FOR", ""))) }
   end
+
+  def current_form
+    @current_form ||= Form.find(params[:form_id])
```
DavidBiddle, line 115: Will this return anything if it's called somewhere that doesn't have a `:form_id` param? Or will it throw an error?

aliuk2012: It will throw an error because the find method needs a Hash and if params doesn't have :form_id it would be nil. 

>raise ArgumentError, "expected attributes to be able to convert to Hash, got #{attributes.inspect}"

DavidBiddle: Cool, I think that's what I'd hope would happen (rather than it failing silently)

app/controllers/pages/address_settings_controller.rb
```diff
@@ -3,20 +3,20 @@ def new
     uk_address = session.dig(:page, :answer_settings, :input_type, :uk_address)
     international_address = session.dig(:page, :answer_settings, :input_type, :international_address)
     @address_settings_form = Pages::AddressSettingsForm.new(uk_address:, international_address:)
-    @address_settings_path = address_settings_create_path(@form)
-    @back_link_url = type_of_answer_new_path(@form)
-    render address_settings_view
```
DavidBiddle, line 8: Think we can get rid of the `address_settings_view` method if we're not using it any more

DavidBiddle (dismissed):
Looks good to me, nifty refactor - just made a suggestion about a method that I think can be removed now

app/controllers/pages/address_settings_controller.rb
aliuk2012: 💯 good spot 👏

DavidBiddle (approved):

PR #729, opened 2023-11-02
MOU signing by users

thomasiles:

Trello card: https://trello.com/c/q0Ud4qRd/1084-mou-digital-page-and-process-that-fires-data-to-the-database

Allow users to sign MOUs.

This PR uses placeholder for the text of the MOU as that will come in another PR.

### DIFF
--- a/app/controllers/mou_signatures_controller.rb
+++ b/app/controllers/mou_signatures_controller.rb
@@ -1,9 +1,48 @@
 class MouSignaturesController < ApplicationController
-  after_action :verify_authorized
+  after_action :verify_authorized, only: %i[index]
 
   def index
     authorize MouSignature, :can_manage_mous?
     @mou_signatures = MouSignature.all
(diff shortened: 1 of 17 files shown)


DavidBiddle (approved):
Looks good! 🎉

app/controllers/mou_signatures_controller.rb
```diff
@@ -1,9 +1,48 @@
 class MouSignaturesController < ApplicationController
-  after_action :verify_authorized
+  after_action :verify_authorized, only: %i[index]
```
aliuk2012, line 2: If none of the other actions need authorization then its ok to remove this before action.

thomasiles: as in just delete it?

thomasiles: I'm going to merge this as it is and tackle this in a follow-up PR if thats ok. I have another small issue I need to fix so I'll throw it in with that.

PR #731, opened 2023-11-02
Revert "Bump selenium-webdriver from 4.14.0 to 4.15.0"

DavidBiddle:
Reverts alphagov/forms-admin#728

The selenium-webdriver update broke our deployments, so reverting to unblock the pipeline.


aliuk2012 (approved):

DavidBiddle:
This wasn't actually the cause of the pipeline breaking, so this isn't necessary any more

PR #734, opened 2023-11-03
Fix build_title optional text check

DavidBiddle:

Trello card: None

The logic in this method was wrong - we never want selection questions displaying as optional, *even if* they've got 'None of the above' enabled. But the current logic will always show the '(optional)' text for selection questions, _whether or not_ they've got 'None of the above' enabled.


lfdebrux (approved):
Looks good to me

PR #736, opened 2023-11-06
Remove detailed_guidance feature flag

SamJamCul:

Trello card: https://trello.com/c/syLfXSVT/1158-remove-feature-flags-for-detailed-guidance

Removes the feature flag `detailed_guidance_enabled` and refactors tests to reflect the feature being permanently enabled. 

### DIFF
--- a/spec/views/pages/edit.html.erb_spec.rb
+++ b/spec/views/pages/edit.html.erb_spec.rb
@@ -5,7 +5,7 @@ describe "pages/edit.html.erb" do
 
   before do
     # Initialize models
-    page = Page.new(id: 1, question_text:, form_id: 1, answer_type: "email", answer_settings: nil)
+    page = Page.new(id: 1, question_text:, form_id: 1, answer_type: "email", answer_settings: nil, page_heading: nil)
     question_form = Pages::QuestionForm.new(page_id: 1, question_text:, form_id: 1, answer_type: "email", answer_settings: nil)
     form = Form.new(id: 1, name: "Form 1", form_id: 1, pages: [page])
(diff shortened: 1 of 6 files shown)


spec/views/pages/edit.html.erb_spec.rb
```diff
@@ -5,7 +5,7 @@
 
   before do
     # Initialize models
-    page = Page.new(id: 1, question_text:, form_id: 1, answer_type: "email", answer_settings: nil)
+    page = Page.new(id: 1, question_text:, form_id: 1, answer_type: "email", answer_settings: nil, page_heading: nil)
```
SamJamCul, line 8: I'm just taking a closer look at what exactly made this test start failing here, but the test passes with this change.

DavidBiddle: What was the failure? Wondering if it's related to this change on Friday: https://github.com/alphagov/forms-runner/pull/482/files

DavidBiddle: Oh wait, that's runner - ignore me!

SamJamCul: Ah I think I've understood it now, I read the error properly. It looks like underneath the feature flag chunk in `app/views/pages/_form.html.erb` there's a section that's checking the `page_heading`. That test was running with the flag off so never ran the `page_heading` check, but now it's on the page every time it needs to get stubbed out.

SamJamCul: Just needed some rubber 🦆!

aliuk2012 (approved):

PR #738, opened 2023-11-06
Update to govuk frontend v5

DavidBiddle:

Trello card: https://trello.com/c/VeOkfpZv/1250-update-forms-runner-to-govuk-frontend-v5

Updates the govuk-frontend npm package and the govuk_components and govuk_design_system_formbuilder gems to v5.0.0.

Also includes:
- several changes mandated or suggested by the upgrade guide - see individual commit messages for details. 
- a version of the fix for the visible environment tag we did in the forms-runner (https://github.com/alphagov/forms-runner/pull/541) 

### Screenshots
Visually things are almost the same, except tags are now sentence-case rather than uppercase:

The name of a form ('A live form with metrics') next to two tags - a pink 'Draft' tag and a blue 'Live' tag.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,8 +49,8 @@ gem "bootsnap", require: false
 gem "vite_rails"
 
 # For GOV.UK branding
-gem "govuk-components", "~> 4.1.2"
-gem "govuk_design_system_formbuilder", "~> 4.1.1"
+gem "govuk-components", "~> 5.0.0"
+gem "govuk_design_system_formbuilder", "~> 5.0.0"
 
 # The autocomplete component is not currently published as a gem, if changing
 # the hash, also change in package.json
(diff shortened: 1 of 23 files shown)


app/components/markdown_editor_component/view.html.erb
```diff
@@ -17,7 +17,7 @@
 
       <%= f.govuk_submit preview_button_translation, secondary: true,  name: "route_to", value: "preview", class: "app-markdown-editor__preview-button","data-ajax-markdown-trigger": true %>
 
-      <details class="govuk-details" data-module="govuk-details">
+      <details class="govuk-details">
```
aliuk2012, line 20: @DavidBiddle could we not use https://govuk-components.netlify.app/components/details/ so we dont have these govuk component hardcoded anywhere?

DavidBiddle: Good shout - dealt with in a separate PR because I'm not sure how long this one will be open https://github.com/alphagov/forms-admin/pull/744

aliuk2012: Might be worth rebasing this PR now so we shouldn't need to worry about this tweak going forward.

aliuk2012 (approved):

PR #740, opened 2023-11-06
Show notification banner when role is upgraded to editor

jamie-o-wilkinson:
This commit adds a banner which is shown to users when they first go to the index after their role is upgraded to editor. This uses the PaperTrail version history to establish if the role has changed to editor, and will create a new version the first time the role changes to editor, to prevent the banner being shown subsequently. I'm not sure this is the best way to achieve this, but it does avoid having to persist anything else (or anything feature specific) to determine if the the banner should be shown. It does also sound like this is a [valid use case for PaperTrail.
](https://github.com/paper-trail-gem/paper_trail#saving-a-new-version-manually)

<details><summary>Screenshot</summary>

image

</details> 

Trello card: https://trello.com/c/IbfXhuYq

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -76,6 +76,15 @@ class User < ApplicationRecord
     mou_signatures.find_by(organisation:)
   end
 
+  def role_changed_to_editor?
+    role_changed_to_editor = editor? && !versions.empty? && versions.last.reify.role != "editor"
+    if role_changed_to_editor
+      self.paper_trail_event = "Role upgrade reported"
+      paper_trail.save_with_version
+    end
+    role_changed_to_editor
+  end
+
 private
 
   def requires_name?
--- a/app/components/role_upgrade_component/view.html.erb
+++ b/app/components/role_upgrade_component/view.html.erb
@@ -1,0 +1,4 @@
+<%= govuk_notification_banner(title_text: t("role_upgrade.title"), success: true) do |banner| %>
+  <% banner.with_heading(text: t("role_upgrade.heading")) %>
+  <%= t("role_upgrade.content_html") %>
+<% end %>
--- a/app/components/role_upgrade_component/view.rb
+++ b/app/components/role_upgrade_component/view.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+module RoleUpgradeComponent
+  class View < ViewComponent::Base
+  end
+end
--- a/app/views/forms/index.html.erb
+++ b/app/views/forms/index.html.erb
@@ -2,7 +2,11 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render TrialRoleWarningComponent::View.new(link_url: user_upgrade_request_path) if @current_user.trial? %>
+    <% if @current_user.trial? %>
+      <%= render TrialRoleWarningComponent::View.new(link_url: user_upgrade_request_path) %>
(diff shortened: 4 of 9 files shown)


app/models/user.rb
```diff
@@ -76,6 +76,14 @@ def current_organisation_mou_signature
     mou_signatures.find_by(organisation:)
   end
 
+  def role_changed_to_editor?
+    role_changed_to_editor = editor? && !versions.empty? && versions.last.reify.role != "editor"
```
jamie-o-wilkinson, line 80: This uses the PaperTrail version history to determine if the role has been changed to editor since the last version was saved. It then saves a version with no changes so that the next time it is checked `role_changed_to_editor` is false. 

This seems a bit hacky, but the alternatives either didn't work the way we'd want (e.g. flash messages would appear on each login, session storage would lead to users missing the message if their session expired before they logged in after upgrade), or involved persisting something specific for this feature that describes state change rather than state (e.g. `role_changed` rather than `role`) which seems worse to me. 

I'd welcome any suggestions!

hannahkc:
In the screenshot, it looks like the spacing between the title and body in the banner is not quite right. I think it should be a bit of a wider gap - like the trail account banner. (Sorry if this is already in hand!)

app/models/user.rb
lfdebrux: Using PaperTrail doesn't sound like too bad an option to me, although I'd be interested to know what @aliuk2012 thinks. 

The only thing I would suggest is using a custom event, so it's maybe more clear in the database what's happened?

https://github.com/paper-trail-gem/paper_trail#the-versionsevent-column

jamie-o-wilkinson: Good suggestion! I'd like that, but on investigation it looks like `paper_trail.save_with_version` always saves with an update event: https://www.rubydoc.info/gems/paper_trail/11.0.0/PaperTrail%2FRecordTrail:save_with_version

lfdebrux: Hmm, looking at the source code it appears to still be overridable?

`save_with_version` does (indirectly) create a `Events::Update` instance [[1](https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L152), [2](https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L101), [3](https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/record_trail.rb#L218)], but the actual value of the `event` column is still customisable:

```
          event: @record.paper_trail_event || "update",
```

https://github.com/paper-trail-gem/paper_trail/blob/47dbc2271c7055fde6f0fe7d3c8bf44f62cbef63/lib/paper_trail/events/update.rb#L29

jamie-o-wilkinson: Hmm interesting. I tried it with no luck. I must've done something silly, I'll look again.

jamie-o-wilkinson:
> In the screenshot, it looks like the spacing between the title and body in the banner is not quite right. I think it should be a bit of a wider gap - like the trail account banner. (Sorry if this is already in hand!)

@hannahkc Good spot, I've resolved this now.

app/models/user.rb
jamie-o-wilkinson: Got it working, thanks for the suggestion!

PR #743, opened 2023-11-07
consistently set default answer settings

aliuk2012:

Trello card: 


app/forms/pages/type_of_answer_form.rb
```diff
@@ -4,15 +4,32 @@ class Pages::TypeOfAnswerForm < BaseForm
...
   validates :answer_type, presence: true, inclusion: { in: Page::ANSWER_TYPES }
 
-  def submit(session)
+  def submit
     return false if invalid?
+    return true unless answer_type_changed?
 
     draft_question
-      .assign_attributes({ answer_type:, answer_settings: nil })
+      .assign_attributes({ answer_type:, answer_settings: default_answer_settings_for_answer_type })
```
aliuk2012, line 12: This is a huge issue we currently have. if a user changes their answer_type or in some cases only clicked "continue" we would completely remove all existing answer settings.

This feels like a better approach because we know what default settings each answer type should have. This isn't perfect but I'll explain why during Thursday's mod session.

app/forms/pages/type_of_answer_form.rb
```diff
@@ -4,15 +4,32 @@ class Pages::TypeOfAnswerForm < BaseForm
   validates :draft_question, presence: true
   validates :answer_type, presence: true, inclusion: { in: Page::ANSWER_TYPES }
 
-  def submit(session)
+  def submit
     return false if invalid?
+    return true unless answer_type_changed?
```
aliuk2012, line 9: We did have a similar guard for "Edit question" journey but it wasn't clear what would happen if someone managed to call this method directly instead of through a controlled route.

spec/forms/pages/type_of_answer_form_spec.rb
```diff
@@ -38,29 +37,37 @@
...
     it "returns false if the form is invalid" do
       allow(type_of_answer_form).to receive(:invalid?).and_return(true)
-      expect(type_of_answer_form.submit(session_mock)).to be_falsey
-    end
-
-    it "sets a session key called 'page' as a hash with the answer type in it" do
-      type_of_answer_form.answer_type = "email"
-      type_of_answer_form.submit(session_mock)
-      expect(session_mock[:page]).to include(answer_type: "email")
+      expect(type_of_answer_form.submit).to be_falsey
```
DavidBiddle, line 42: I know it's in the original code here, but wondering if you know why we're using `be_falsey` here instead of `be false`?

DavidBiddle: Not essential to this PR but if `be_falsey` isn't needed then we could potentially make the expectation more specific

aliuk2012: I'm going to leave this as is for the moment, but we do need to review our specs more generally and refactor those too.

PR #659, opened 2023-10-03
Create new QuestionForm object

aliuk2012:
### What problem does this pull request solve?

Create new QuestionForm object to replace using Page ActiveResource model directly. This means that the "Edit question" page just becomes another step in the add/edit question wizard. 

Updated the sequence diagrams https://github.com/alphagov/forms/pull/104

#### Current flow

```mermaid
erDiagram
    FORM{
        integer ID pk "this wouldn't necessarily be in sync with what is in forms_api"
    }
    session{
    
    }
    PAGES{
        integer ID pk
        integer forms_id fk
        integer user_id fk
        integer page_id "if page already exists"
        string question_text 
        string hint_text
        boolean is_optional
        string page_heading
        string guidance_markdown
        jsonb answer_settings
        datetime created_at
        datetime updated_at
    }
    PAGES_TYPE_OF_ANSWER_FORM{
        string answer_type
    
    }
    PAGES_ADDRESS_SETTINGS_FORM{
        boolean uk_address
        boolean international_address 
        string notes "input type becomes a hash instead of a string for this answer type eg. { input_type: { uk_address:, international_address: } }"
    }
    
    PAGES_DATE_SETTINGS_FORM{
        string input_type "can be 'date_of_birth' or 'other_date'"
        string notes "{ input_type:}"
    }

    PAGES_NAME_SETTINGS_FORM{
        string input_type "can be %w[full_name first_and_last_name first_middle_and_last_name]"
        boolean title_needed
        string notes "{ input_type:, title_needed: }"
    }

    PAGES_QUESTION_TEXT_FORM{
        string question_text "this is the same value that will be used for the final Page class"
        string notes "does not update answer_settings"
    }

    PAGES_SELECTIONS_SETTINGS_FORM{
        json seection_options "this is a list of hashes [{name: 'Option 1'}, {name: 'Option 2'}...]"
        boolean only_one_option "forms-runner uses this to control whether radio buttons or checkboxes are shown"
        boolean include_none_of_the_above "sets 'is_optional' and allows users to skip question"
        string notes "updates answer_settings and is_optional {is_option: include_none_of_the_above, answer_settings: { only_one_option:, selection_options: }}"
    }
    PAGES_TEXT_SETTINGS_FORM{
        string input_type "single_line | long_text]"
        string hint_text
        string notes "{ input_type: }"
    }

    PAGES_GUIDANCE_FORM{
        string page_heading
        string guidance_markdown
[...]

PR #755, opened 2023-11-09
Rename clear_question_session_data to clear_draft_questions_data

aliuk2012:
This work follow on from:
* https://github.com/alphagov/forms-admin/pull/743
* https://github.com/alphagov/forms-admin/pull/690
* https://github.com/alphagov/forms-admin/pull/669
* https://github.com/alphagov/forms-admin/pull/658
* https://github.com/alphagov/forms-admin/pull/659

https://github.com/alphagov/forms-admin/pull/743 changes meant that we can now stop recording settings information to session and start relying on only using DraftQuestion.

Trello card: https://trello.com/c/RGsOE1bl/1074-switch-all-the-page-form-objects-over-to-use-draftquestion-instead-of-session


app/controllers/pages/questions_controller.rb
```diff
@@ -6,34 +6,28 @@ def new
     is_optional = draft_question.is_optional == "true"
     page_heading = draft_question.page_heading
     guidance_markdown = draft_question.guidance_markdown
-    @question_form = Pages::QuestionForm.new(form_id: current_form.id, answer_type:, question_text:, answer_settings:, is_optional:, draft_question:)
+    @question_form = Pages::QuestionForm.new(answer_type:, question_text:, answer_settings:, is_optional:, draft_question:)
```
lfdebrux, line 9: Can we use `page_params_for_form_object` here?

app/controllers/pages/questions_controller.rb
```diff
@@ -6,34 +6,28 @@ def new
...
     end
   end
 
   def edit
-    @question_form = Pages::QuestionForm.new(form_id: current_form.id,
-                                             answer_type: draft_question.answer_type,
+    @question_form = Pages::QuestionForm.new(answer_type: draft_question.answer_type,
                                              question_text: page.question_text,
                                              hint_text: draft_question.hint_text,
                                              is_optional: draft_question.is_optional,
```
lfdebrux, line 33: Can we use `page_params_for_form_object` here?

lfdebrux (commented):
These changes make sense to me.

Commit 8384d13 could probably do with being split into two, and I think the refactoring you're doing could be extended further, but otherwise this all looks good.

app/controllers/pages/questions_controller.rb
aliuk2012: We cant use it because it includes params passed in from the form.  I do see what you mean that we could refactor this further so there is less duplication with passing in all these values to the forms object.

PR #765, opened 2023-11-15
1184-prevent-forms-runner-and-forms-admin-cloudwatch-requests-when-in-unsuitable-environments

SamJamCul:

Trello card: https://trello.com/c/eGLh3RTp/1184-prevent-forms-runner-and-forms-admin-cloudwatch-requests-when-in-unsuitable-environments

Now that https://github.com/alphagov/forms-deploy/pull/435 is in, we can make use of the environment variable that it's set to turn off metrics in dev and user research. 

This PR makes the CloudWatchService check the environment variable `cloudwatch_metrics_enabled` before it continues making CloudWatch requests, and breaks early if they're disabled.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -77,6 +77,7 @@ class Form < ActiveResource::Base
 
   def metrics_data
     return nil unless FeatureService.enabled?(:metrics_for_form_creators_enabled)
+    return nil unless has_live_version
 
     # If the form went live today, there won't be any metrics to show
     today = Time.zone.today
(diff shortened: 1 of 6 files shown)


app/models/form.rb
```diff
@@ -95,6 +96,8 @@ def metrics_data
 
     Sentry.capture_exception(e)
     nil
+  rescue CloudWatchService::MetricsDisabledError
+    nil
```
DavidBiddle, line 100: Nice solution!

DavidBiddle (approved):
Looks good to me! 🎉

PR #767, opened 2023-11-16
Remove cddo auth provider

thomasiles:
### Remove the CDDO SSO login code and tests

Trello card: https://trello.com/c/j6BVMUIg/1204-remove-cddo-sign-in-code-from-forms-admin


lfdebrux (approved):
LGTM

thomasiles:
> Do we want to remove this line as well?
> 
> https://github.com/alphagov/forms-admin/blob/83b8604025c3b3242e6fcb807cdce833b9db8f75/spec/controller/application_controller_spec.rb#L40

good find :) Yes we do, I'll merge this and do that in another PR.

PR #769, opened 2023-11-16
WIP: Spike remove gds sso

thomasiles:
This is a quick spike to look at what the minimum amount of needed to remove the gds sso gem from our application.

It's not complete and not very tested. It works fine but:
- not all tests have been removed
- nothing has been refactored
- there are some odd bits from the gem which I've carried over but probably need questioning
- I haven't removed any columns from the User table


app/models/user.rb
```diff
@@ -85,6 +84,18 @@ def role_changed_to_editor?
     role_changed_to_editor
   end
 
+  def clear_remotely_signed_out!
+    # rubocop:disable Rails/SkipsModelValidations
+    update_attribute(:remotely_signed_out, false)
```
thomasiles, line 89: not sure why update_attribute is used here - maybe to make sure it happens despite the state the user is in?

app/service/navigation_items_service.rb
```diff
@@ -70,22 +70,11 @@ def user_provider
   end
 
   def signout_url
-    if user_provider == :gds
-      gds_sign_out_path
-    elsif %i[auth0 cddo_sso mock_gds_sso].include? user_provider
+    if %i[auth0 mock_gds_sso].include? user_provider
```
thomasiles, line 73: this can be updated

config/initializers/authentication.rb
```diff
@@ -1,8 +1,42 @@
...
+  user.clear_remotely_signed_out!
+end
+
+Warden::Manager.serialize_into_session do |user|
+  if user.respond_to?(:uid) && user.uid
+    [user.uid, Time.zone.now.utc.iso8601]
+  end
+end
+
+Warden::Manager.serialize_from_session do |(uid, auth_timestamp)|
```
thomasiles, line 17: we could add more tests around these parts lifted from gds_sso

spec/requests/authentication_controller_spec.rb
```diff
@@ -122,17 +116,17 @@ def authenticate!
...
-      allow(controller_spy).to receive(:callback_from_omniauth).and_call_original
+    #   allow(controller_spy).to receive(:callback_from_omniauth).and_call_original
 
-      get "/auth/gds/callback"
+    #   get "/auth/gds/callback"
 
-      expect(controller_spy).to have_received :callback_from_omniauth
-    end
+    #   expect(controller_spy).to have_received :callback_from_omniauth
+    # end
```
thomasiles, line 129: This is a gds_sso specific test which I'm not sure we need

thomasiles, line 129: This is a gds_sso specific test which I'm not sure we need

app/models/user.rb
lfdebrux: I don't think we need the `remotely_signed_out!` stuff? The existing stuff is reliant on the GOV.UK Signon servers calling an API exposed by gds_sso gem.

theseanything:
@thomasiles do we still need this or can this be closed?

PR #772, opened 2023-11-16
Fixup edit question view to use draft_question instead of page model

aliuk2012:

This work should fix the issues raised and reverted in the below PR:
* https://github.com/alphagov/forms-admin/pull/763
* https://github.com/alphagov/forms-admin/pull/764
* https://github.com/alphagov/forms-admin/pull/768

This is still left over work from our refactor from session  to activerecord and using DraftQuestion rather than Page.

This doesn't cover all page_object usage in the view but to try limit the scope of this change, I've focused on replacing the simple instances first. The next PR will focus on the answer setting summary component.

Trello card: https://trello.com/c/HeXLnq1m/1201-form-creators-cannot-view-all-details-for-a-particular-page-when-they-edit-the-question


spec/requests/pages/questions_controller_spec.rb
```diff
@@ -119,7 +119,7 @@
           mock.get "/api/v1/forms/2/pages", req_headers, form_pages_response, 200
           mock.get "/api/v1/forms/2/pages/1", req_headers, page_response, 200
         end
-
+        draft_question
```
lfdebrux, line 122: Is this line here just for debugging? Or does it do something?

aliuk2012: Ah, this isn't for debugging. The way lets work in spec is that they are only initialised and set when they are first called. If you look at the code now where does it call `draft_question` before this point and we need a draft_question setup before making the request. If you were to remove this line the tests should break.

I could have used `let!(:draft_question)` rather than `let(:draft_question) which would initialised it when the code was first read but its considered back practice to use let! because you lose the benefit of normal lets only being initialised if they need to and its waste to create the draft question if some specs dont need it.

lfdebrux: Okay, can we put in a comment explaining what's going on?

aliuk2012: done and squashed.

PR #775, opened 2023-11-16
Refactor PageSettingsSummaryComponent to not need so many params

aliuk2012:
PageSettingsSummaryComponent required a lot of params when it was being called. This on its own is a bit of a code smell. Along with all these params we were also setting them up in the relevent new/edit views and passing them along to the `forms` partial before passing them on to the view component. This is a lot of processing for the views but also we are defining those values in the view instead of controller action. It feels like a lot of unnecessary noise when we could get the component to workout which link it needs to display from the data we already provide it. 

These changes will help us when we come to implement https://trello.com/c/Z46xTPB5/1085-spike-add-check-your-question-page because we dont have to lift/shift all these params to the new page and we dont have to work out if the user is creating/editing questions.

This refactor is still not 100% because ideally working out the what rows/data/links etc should be done in a service and once that is done theres no real need for this component at all because we can use generic govuk_components summary list component.

Trello card: https://trello.com/c/emgcDnRB/1210-refactor-pagesettingssummarycomponent-to-not-need-so-many-params

### DIFF
--- a/app/components/page_settings_summary_component/view.rb
+++ b/app/components/page_settings_summary_component/view.rb
@@ -2,15 +2,20 @@
 
 module PageSettingsSummaryComponent
   class View < ViewComponent::Base
-    def initialize(draft_question, change_answer_type_path: "", change_selections_settings_path: "", change_text_settings_path: "", change_date_settings_path: "", change_address_settings_path: "", change_name_settings_path: "")
+    include Rails.application.routes.url_helpers
+    def initialize(draft_question, change_answer_type_path: "", change_text_settings_path: "", change_date_settings_path: "", change_address_settings_path: "", change_name_settings_path: "")
       super
       @draft_question = draft_question
       @change_answer_type_path = change_answer_type_path
-      @change_selections_settings_path = change_selections_settings_path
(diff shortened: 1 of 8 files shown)


app/components/page_settings_summary_component/view.rb
```diff
@@ -35,5 +40,67 @@ def answer_settings
...
       answer_settings[:selection_options].map { |option| option[:name] }.join(", ")
     end
+
+    def change_answer_type_path
+      if is_new_question?
+        type_of_answer_new_path(form_id: @draft_question.form_id)
+      else
+        type_of_answer_edit_path(form_id: @draft_question.form_id, page_id: @draft_question.page_id)
+      end
+    end
```
DavidBiddle, line 50: The similarity of all these path methods makes me think there's probably a way to abstract them into something reusable

lfdebrux (approved):
I've tested this locally, couldn't find any bugs! 🎉

app/components/page_settings_summary_component/view.rb
DavidBiddle: Should say this is an optional suggestion unless you can think of a neat way to do it!

PR #781, opened 2023-11-20
Fix create/update question journey when user should be expecting to see validation error messages

aliuk2012:

When a user was trying to create/update a question if there was a validation error message then the page would fail to load and they would see a 500 error page. I added two new request specs and fixed the issue for both actions

Trello card: https://trello.com/c/THcM9yHL/1212-users-see-500-error-page-when-they-attempt-to-save-invalid-new-question-without-question-text


thomasiles (approved):
nice!

PR #783, opened 2023-11-21
Pages::QuestionForm does not require form or page id

aliuk2012:
This work follow on from:
* https://github.com/alphagov/forms-admin/pull/743
* https://github.com/alphagov/forms-admin/pull/690
* https://github.com/alphagov/forms-admin/pull/669
* https://github.com/alphagov/forms-admin/pull/658
* https://github.com/alphagov/forms-admin/pull/659

Theres no need for the form object to know about form id or page id. Its primary forcus is on display a question text, hint text and is optional field. Behind the scenes it still needs answer type because in the view answer type is used as part of the locales key lookup to customise the hint text for the hint text field.

Trello card: https://trello.com/c/RGsOE1bl/1074-switch-all-the-page-form-objects-over-to-use-draftquestion-instead-of-session

### DIFF
--- a/app/controllers/pages/questions_controller.rb
+++ b/app/controllers/pages/questions_controller.rb
@@ -1,25 +1,20 @@
 class Pages::QuestionsController < PagesController
   def new
-    @question_form = Pages::QuestionForm.new(form_id: current_form.id,
-                                             answer_type: draft_question.answer_type,
+    @question_form = Pages::QuestionForm.new(answer_type: draft_question.answer_type,
                                              question_text: draft_question.question_text,
                                              answer_settings: draft_question.answer_settings,
                                              is_optional: draft_question.is_optional,
                                              draft_question:)
 
+    # TODO: Remove this once we have a check your question view. The new view should also pull data directly from draft_question instead of through page model
     @page = Page.new(form_id: current_form.id)
     render :new, locals: { current_form:, draft_question: }
   end
 
   def create
-    @question_form = Pages::QuestionForm.new(page_params.merge(answer_settings: draft_question.answer_settings,
-                                                               page_heading: draft_question.page_heading,
-                                                               guidance_markdown: draft_question.guidance_markdown,
-                                                               draft_question:))
-    @page = Page.new(page_params.merge(answer_settings: draft_question.answer_settings,
-                                       page_heading: draft_question.page_heading,
-                                       guidance_markdown: draft_question.guidance_markdown,
-                                       answer_type: draft_question.answer_type))
+    @question_form = Pages::QuestionForm.new(page_params_for_form_object)
+
+    @page = Page.new(page_params_for_forms_api)
 
     # TODO: Move Page creation to be part of the form submit method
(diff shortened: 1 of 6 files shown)


app/controllers/pages/questions_controller.rb
```diff
@@ -1,25 +1,20 @@
...
   def create
-    @question_form = Pages::QuestionForm.new(page_params.merge(answer_settings: draft_question.answer_settings,
-                                                               page_heading: draft_question.page_heading,
-                                                               guidance_markdown: draft_question.guidance_markdown,
-                                                               draft_question:))
-    @page = Page.new(page_params.merge(answer_settings: draft_question.answer_settings,
-                                       page_heading: draft_question.page_heading,
-                                       guidance_markdown: draft_question.guidance_markdown,
-                                       answer_type: draft_question.answer_type))
+    @question_form = Pages::QuestionForm.new(page_params.merge(page_params_for_form_object))
```
lfdebrux, line 15: I don't think we need to do `page_params.merge` here? It looks like `page_params_for_form_object` already does it.

```suggestion
    @question_form = Pages::QuestionForm.new(page_params_for_form_object)
```

app/controllers/pages/questions_controller.rb
```diff
@@ -41,16 +35,9 @@ def edit
...
-    page.guidance_markdown = draft_question.guidance_markdown
-    page.answer_type = draft_question.answer_type
-    page.answer_settings = draft_question.answer_settings
+    page.load(page_params_for_forms_api)
 
-    @question_form = Pages::QuestionForm.new(page_params.merge(answer_settings: draft_question.answer_settings,
-                                                               page_heading: draft_question.page_heading,
-                                                               guidance_markdown: draft_question.guidance_markdown,
-                                                               draft_question:))
+    @question_form = Pages::QuestionForm.new(page_params_for_form_object)
```
lfdebrux, line 40: How come this instance didn't use `answer_type` before? 🤔 Does it make a difference if it has it set now?

lfdebrux (dismissed):
Works for me, I have some small questions but nothing that should make a difference I think.

app/controllers/pages/questions_controller.rb
aliuk2012: Good catch, thanks I'll push up another commit to fix this

app/controllers/pages/questions_controller.rb
aliuk2012: Great question, I had a look into it and it doesn't make a difference.  I think I'm going to leave it as is for the moment so we dont risk breaking it now. When we come to adding "Check your question" page all of this will become a lot clearer as to what is needed for the form fields and what is needed for the page object to be created/updated.

app/controllers/pages/questions_controller.rb
aliuk2012: I decided to squash the commit to save you having to approve it again once the two commits were squashed.

lfdebrux (approved):

PR #786, opened 2023-11-22
BUG: Users not seeing pre-filled in forms when they edit an existing questions answer settings

aliuk2012:

**ISSUE:**  When a user tried to edit an existing questions answer setting, the settings page was not populating the form fields with the last saved values.

**CAUSE**: We weren't consistently switching the hash keys to symbols using either `with_indifferent_access` or `deep_symbolize_keys`. 

**FIX**:  see commit message for 3acec0922a61868bbd62dbd2159c38840bcace7b

### Going forward...
You should always use symbols when trying to access any values within answer_settings.

Trello card: https://trello.com/c/KnK13cLX/1213-when-a-user-tries-to-change-an-existing-answer-setting-the-page-is-not-populating-the-form-fields-with-their-previous-saved-valu

### DIFF
--- a/spec/models/draft_question_spec.rb
+++ b/spec/models/draft_question_spec.rb
@@ -149,4 +149,39 @@ RSpec.describe DraftQuestion, type: :model do
       end
     end
   end
+
+  describe "answer_settings" do
+    let(:draft_question) { create :draft_question, answer_settings: }
+    let(:answer_settings) do
+      {
+        "hello" => "I have a string as a key",
+        "nested_attributes" => {
+          "name" => "Joe Bloggs",
+        },
+      }
+    end
+
+    it "returns a hash with all keys as symbols" do
+      expect(draft_question.answer_settings).to eq(hello: "I have a string as a key",
+                                                   nested_attributes: {
+                                                     name: "Joe Bloggs",
+                                                   })
+    end
+
+    context "when answer_settings is empty JSON" do
+      let(:answer_settings) { {} }
+
+      it "returns empty hash" do
+        expect(draft_question.answer_settings).to be_empty
+      end
+    end
+
+    context "when answer_settings is nil" do
+      let(:answer_settings) { nil }
+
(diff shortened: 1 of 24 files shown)


spec/models/draft_question_spec.rb
```diff
@@ -149,4 +149,39 @@
...
+      it "returns empty hash" do
+        expect(draft_question.answer_settings).to be_empty
+      end
+    end
+
+    context "when answer_settings is nil" do
+      let(:answer_settings) { nil }
+
+      it "returns empty hash" do
+        expect(draft_question.answer_settings).to be_nil
```
thomasiles, line 183: this `it` block and the one above it both have the text "returns empty hash". This one uses `be_nil,` not `be_empty`. Should the text be different?

A question outside of this PR: does `answer_settings` need a difference between an empty hash and nil?

Some people seem to use active record serializers to do the conversion [like here](https://github.com/rails/rails/issues/25894#issuecomment-777516944). They often replace nil with `{}` for convenience. I don't know if that would help us?

thomasiles (approved):
Impressive tests, fix and writeup!🎖️

spec/models/draft_question_spec.rb
aliuk2012: @thomasiles Your question about empty hash/nil, I did think about it initially and decided to just return whatever it was set to......Maybe we should return an empty hash. Its a case of changing this one line to return empty hash instead of `raw_setting`.

https://github.com/alphagov/forms-admin/pull/786/files#diff-baf1f2aaaa9c2156d3a608cfd7e96f8f7b85deed52347177f551d1c24f154dc2R14

aliuk2012: @thomasiles I've pushed a new commit which should default `answer_settings` to empty hash and also resolves the copy issue you spotted in the specs

thomasiles (approved):

PR #787, opened 2023-11-22
Change the OmniAuth redirect into a POST rather than a GET

thomasiles:
# Change the OmniAuth redirect into a POST rather than a GET

[Trello card](https://trello.com/c/kFUuF91A/1182-fix-security-vulnerability-in-how-we-redirect-users-to-sign-in-page)

This PR changes how users login to our system.

We don't keep a list of user names and passwords within our service to log users in. Instead, we rely on external services to manage users and check users are who they claim to be. When an unauthenticated request is made to our application, we redirect the user to the external service to login and then be redirected back to us.

Each service has a slightly expectations about how users are redirected and what information they need. To manage the differences, we use we use a library called OmniAuth. 

OmniAuth works by creating new routes like `/auth/auth0` and `/auth/gds_sso`.
These routes are the start of the login journey for that service. In our application code, all we need to do to make a user follow that journey is redirect the user to one of the /auth/* routes.

This process is shown in the [authentication flow diagram](https://github.com/alphagov/forms/blob/main/diagrams/sequence-diagrams/authentication.md)
as steps 1-5.

There is a potential security issue with this approach. Because the OmniAuth routes start the auth journey based on a GET request, it's very easy for a user to be sent through them without realising. The user's browser follows redirects automatically. When they try to access our service, they are redirected to the login page without taking any further action to confirm they want to login.

If a user is already logged into an auth provider, they may bounce through the authentication without ever pressing a button.

This [CVE-2015-9284](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284) against OmniAuth explains how it can turn into a vulnerability.

If a user can associate multiple accounts with a service, perhaps a username and password and a google account. It can be possible for a an attacker to trick a victim to associate a google account the attacker owns with the victims account. It's a bit obscure - the victim would have to be logged into both accounts first.

By itself, with our current authentication setup, this isn't a security issue. Forms authentication only allows one identity to be linked to a user account at a time and Auth0 always asks users to login. There is no concept of using other accounts to login. Even when we were migrating away from GDS Signon, a user could only be associated with one login
method. We also only use our Auth0 tenant for our single application. It's not possible to have an account with another service and not have a forms account.

## Why do we need to change if CVE-2015-9284 does not cause us any issues?

Redirecting a user through to an authentication method directly still causes us a few issues:
- Auth0 logs that we are sending [too many requests to the login
  page](https://trello.com/c/OcsYrxXD/1270-spike-how-do-we-stop-bots-from-trying-to-access-sign-in-pages-for-forms-admin). This might be because we redirect any traffic - bots, services like slack which show embedded previews etc.
- We still have the signon omniauth route in production. It will be removed when we clean up the code but user's can be redirected to [https://admin.dev.forms.service.gov.uk/auth/gds]. Because the settings are no longer configured in our service, users are not able to login. It's still possible to trick users to make GET requests to these routes and there may be security issues lurking there.
- If we change our Auth0 config in the future, such as allowing users to link their google accounts, we may open ourselves up to the original OmniAuth CVE.

## What change are we making to secure the service?
We are switching the /auth/* routes to only respond to POST requests. This will stop browsers following redirects straight to the authentication method without a form submit. This is the fix recommended by OmniAuth. They have not made this the default to keep compatibility with existing code and created a new GEM to handle the logic.

This means that our current auth flow must change. Users can no longer be redirected straight to Auth0.
Instead they are redirected to /login, where they must submit a form to make a POST request. The form is actually a single submit button. When the login form is submitted, using a POST request, the user is redirected straight to Auth0. 

This solves the security issue but it does add friction to our login flow.
To fix this, we are using some client-side javascript to automatically press the Login button. To the user, it should feel similar to the redirect, with a slightly awkward page load in between.

This is a compromise.

If it's annoying or causes issues we will remove the javascript and users will need to press the button themselves.

It also removes some of the user intent - it's still possible to send users directly to and authentication provider by giving them a link.

### Do we need to change the e2e tests
No. Because of the javascript which clicks the button for the user, the flow hasn't changed. The admin code has been changed to add the required connection string to the post request form when the `auth=e2e` string is present.

### Does that work across browsers
Yes. It seems to work across all the browsers I used to test it in browserstack.

### Does it work for screen readers etc.
Yes. Screen readers don't read the login page as the redirect happens pretty quickly. It's possible they might in some situations but I think it would ok as reading the login screen makes sense in the context.

### What happens if JS is not enabled or fails?
The login button is not clicked and user must do it themselves.

### Can we provide a direct link to login from our product pages?
Not without relaxing our security constraints on the login page. The form uses a CSRF token to ensure the request came from our site. The product pages could submit a form across domains but would have no way of generating a CSRF token.

We could come up with another way of authenticating the POST request but I think it would be better to change the design to something more traditional.

### How do other sites handle this?
It's more common to have the product pages and service on the same site. That way a login button can be pressed by the user to send them to auth.

### What does the login screen look like?
localhost_3000_sign-in

### DIFF
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -5,18 +5,18 @@ class AuthenticationController < ApplicationController
   layout false
 
   def self.call(env)
-    action(:redirect_to_login).call(env)
+    action(:redirect_to_sign_in).call(env)
   end
 
-  def redirect_to_login
+  def redirect_to_sign_in
     store_location(attempted_path) if request.get?
 
-    redirect_to(login_url(request.query_parameters))
+    redirect_to(sign_in_url(request.query_parameters))
   end
 
-  def login
+  def sign_in
     @is_e2e_user = e2e_user?
-    render "authentications/login", layout: "application"
+    render "authentications/sign_in", layout: "application"
   end
 
   def callback_from_omniauth
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -8,6 +8,10 @@ class AuthenticationController < ApplicationController
     action(:redirect_to_omniauth).call(env)
   end
 
+  def login
+    render "authentications/login", layout: "application"
+  end
+
   def redirect_to_omniauth
     store_location(attempted_path) if request.get?
 
@@ -17,7 +21,7 @@ class AuthenticationController < ApplicationController
       params[:connection] = "Username-Password-Authentication"
     end
 
-    redirect_to ActionDispatch::Http::URL.path_for(path: "/auth/#{default_provider}", params:)
+    redirect_to(login_url, params:)
   end
 
   def e2e_user?
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -1,6 +1,6 @@
 class AuthenticationController < ApplicationController
   skip_before_action :authenticate_and_check_access
-  protect_from_forgery with: :null_session, only: %i[redirect_to_omniauth]
+  protect_from_forgery with: :null_session, only: %i[redirect_to_sign_in]
 
   layout false
 
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -1,5 +1,6 @@
 class AuthenticationController < ApplicationController
   skip_before_action :authenticate_and_check_access
+  protect_from_forgery with: :null_session, only: %i[redirect_to_omniauth]
 
   layout false
 
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -14,23 +14,20 @@ class AuthenticationController < ApplicationController
     redirect_to(sign_in_url(request.query_parameters))
   end
 
-  def sign_in
-    @is_e2e_user = e2e_user?
-    render "authentications/sign_in", layout: "application"
-  end
-
   def callback_from_omniauth
     authenticate_user!
     redirect_to stored_location || "/"
   end
 
+  def sign_in
+    @is_e2e_user = e2e_user?
+    render "authentications/sign_in", layout: "application"
+  end
+
   def sign_up
     store_location root_path
-    if default_provider == "auth0"
-      redirect_to "/auth/auth0?screen_hint=signup"
-    else
-      redirect_to "/auth/#{default_provider}"
-    end
+    @is_e2e_user = e2e_user?
+    render "authentications/sign_up", layout: "application"
   end
 
   def sign_out
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -5,28 +5,18 @@ class AuthenticationController < ApplicationController
   layout false
 
   def self.call(env)
-    action(:redirect_to_omniauth).call(env)
+    action(:redirect_to_login).call(env)
+  end
+
+  def redirect_to_login
+    store_location(attempted_path) if request.get?
+
+    redirect_to(login_url(request.query_parameters))
   end
 
   def login
+    @is_e2e_user = e2e_user?
     render "authentications/login", layout: "application"
-  end
-
-  def redirect_to_omniauth
-    store_location(attempted_path) if request.get?
-
-    params = {}
-    if default_provider == "auth0" && e2e_user?
-      # the value passed in connection must match the name given in the auth0 terraform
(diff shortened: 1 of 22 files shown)


app/views/authentications/login.html.erb
```diff
@@ -0,0 +1,14 @@
...
+<%= login_button(is_e2e_user: @is_e2e_user) %>
+
+<% content_for :body_end do %>
+  <script>
+    (function() {
+      var loginButton = document.querySelector('[data-module="login-button"]');
+      if (loginButton && typeof loginButton.click === 'function') {
+        loginButton.click();
+      }
+    })();
```
thomasiles, line 12: The alternative to inlining this code is to either put it in the application bundle or add a new entrypoint to be loaded separately.

I could be convinced with either option. I inlined it so it will load quickly and stuck it at the bottom of the body to not block rendering.

I'm not sure whats best. I haven't provided tests for this code because it's so simple and is tested by the feature tests and e2e tests.

config/settings/development.yml
```diff
@@ -9,4 +9,6 @@ forms_api:
   auth_key: changeme
 
 # Use mock authentication
+# in development use 'mock_gds_sso' to always be logged in or 'developer' to use a
+# mock flow through omniauth
 auth_provider: mock_gds_sso
```
thomasiles, line 14: It's very hard to test this change without matching production more, hence adding the developer provider.

lfdebrux:
Thank you for the very detailed write up!

One thing I'm curious about, if someone hosted an iframe which pointed to the `/auth/auth0` route of our site, would the JavaScript execute? Could we tell browsers to not execute the JavaScript?

config/locales/en.yml
```diff
@@ -395,6 +395,7 @@ en:
       Any changes you make to a live form will be updated in the form
       immediately.
     notification_title: Important
+  login_button: Login
```
lfdebrux, line 398: The GOV.UK style guide recommends using the phrase "sign in" to refer to logging in [[1](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#sign-in-or-log-in)]; this matches the language used on most services in DSP [[2](https://github.com/alphagov/govuk-design-system-backlog/issues/95#issuecomment-434643428)].

We should change this content, and also _ideally_ change the route and method names to be `sign-in` or `sign_in`.

lfdebrux (dismissed):

config/locales/en.yml
thomasiles: that makes sense, I'll change it.

It also reminds me that I need to change the sign-up link to use post too 🙄

app/controllers/authentication_controller.rb
```diff
@@ -1,41 +1,33 @@
 class AuthenticationController < ApplicationController
   skip_before_action :authenticate_and_check_access
+  protect_from_forgery with: :null_session, only: %i[redirect_to_omniauth]
```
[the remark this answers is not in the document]
lfdebrux: I'm curious, why do we need `protect_from_forgery` now, and how is it related to the developer auth provider?

app/helpers/application_helper.rb
```diff
@@ -79,4 +79,25 @@ def user_access_options(access_options = %w[true false])
...
+    govuk_button_to t("sign_in_button"), omniauth_authorize_path, params: sign_in_params(is_e2e_user:), data: { module: "sign-in-button" }
+  end
+
+  def sign_up_button(is_e2e_user:)
+    govuk_button_to t("sign_up_button"), omniauth_authorize_path, params: sign_in_params(is_e2e_user:, login_type: :sign_up), data: { module: "sign-in-button" }
+  end
+
+  def omniauth_authorize_path
+    "/auth/#{Settings.auth_provider}/"
+  end
```
lfdebrux, line 93: This won't work for Signon (see definition of `default_provider` method in AuthenticationController), but I'm guessing you knew that already.

lfdebrux (approved):

app/views/authentications/login.html.erb
lfdebrux: I think what you've done here looks good to me

config/settings/development.yml
lfdebrux: Nice!

app/helpers/application_helper.rb
thomasiles: I didn't! I guess it won't matter if we are removing signon in the near future anyway?

lfdebrux: Yeah I'd reckon so.

PR #790, opened 2023-11-23
Refactor metrics for live form check

SamJamCul:

Trello card: https://trello.com/c/eGLh3RTp/1184-prevent-forms-runner-and-forms-admin-cloudwatch-requests-when-in-unsuitable-environments

This is the last PR in the series which sets up an environment variable that controls CloudWatch logging. We are able to decide whether we send retrieve things from CloudWatch or not in each environment, in order to prevent unnecessary requests.


DavidBiddle (approved):
Looks good to me 🎉

PR #792, opened 2023-11-24
Add markdown editor to what happens next field

DavidBiddle:

Trello card: https://trello.com/c/AbPj2LRr/1191-allow-markdown-on-confirmation-page-what-happens-next-content
Also fixes: https://trello.com/c/fAC4f0CM/503-hmtl-appears-in-the-what-happens-next-content-on-a-live-forms-details-page

This PR adds the markdown editor component to the what happens next page. The markdown editor will be shown when the what_happens_next_markdown field is anything other than `nil`. This means that once we've done the migration in the API this field should start appearing immediately.

As part of this PR we:
- add the component to the what happens next page
- render the what happens next HTML and markdown on the live form page

To do in other PRs:
- convert the HTML content in the API into markdown and store it in this new field
- cleanup: delete the old field and all the logic related to it which we no longer need

### Testing locally
If you want to see the markdown editor on your local machine, set up (using the Rails API or the seed file) a form with `what_happens_next_markdown` equal to `""` (or any markdown string). 

#### What happens next field before conversion step in API
A plaintext field with the label 'Enter some information to tell people what will happen next', follwed by a green 'Save and continue' button. The field has a HTML link in it

#### What happens next field after conversion step in API
A markdown editor for a field with the label 'Enter some information to tell people what will happen next', follwed by a green 'Save and continue' button. The markdown editor has toolbar buttons for links, bullets and numbered lists, as well as tabs for writing and previewing content, and a 'Formatting help' link. The field has a markdown link and a bulleted list in it

#### What happens next markdown field (with unsupported markdown)
The same markdown field as above. This time the user has added a level 2 heading, and the field has a red inline error with the text "Error: The information about what happens next can only contain formatting for links, bulleted lists (*), or numbered lists (1.)"

#### What happens next preview on live form page
The what happens next text shown in context on the live form page between the questions and submission email. The markdown from the precious steps has been rendered to HTML inside a 2px grey border.

### DIFF
--- a/app/controllers/forms/what_happens_next_controller.rb
+++ b/app/controllers/forms/what_happens_next_controller.rb
@@ -1,14 +1,17 @@
 module Forms
   class WhatHappensNextController < ApplicationController
     after_action :verify_authorized
+
     def new
       authorize current_form, :can_view_form?
       @what_happens_next_form = WhatHappensNextForm.new(form: current_form).assign_form_values
+      @preview_html = preview_html(@what_happens_next_form)
     end
 
     def create
       authorize current_form, :can_view_form?
       @what_happens_next_form = WhatHappensNextForm.new(**what_happens_next_form_params)
+      @preview_html = preview_html(@what_happens_next_form)
 
       if @what_happens_next_form.submit
         redirect_to form_path(@what_happens_next_form.form), success: t("banner.success.form.what_happens_next_saved")
@@ -17,10 +20,24 @@ module Forms
       end
     end
 
+    def render_preview
+      authorize current_form, :can_view_form?
+      @what_happens_next_form = WhatHappensNextForm.new(what_happens_next_markdown: params[:markdown])
+      @what_happens_next_form.validate
+
+      render json: { preview_html: preview_html(@what_happens_next_form), errors: @what_happens_next_form.errors[:what_happens_next_markdown] }.to_json
+    end
+
   private
 
     def what_happens_next_form_params
-      params.require(:forms_what_happens_next_form).permit(:what_happens_next_text).merge(form: current_form)
+      params.require(:forms_what_happens_next_form).permit(:what_happens_next_text, :what_happens_next_markdown).merge(form: current_form)
+    end
+
+    def preview_html(what_happens_next_form)
+      return t("guidance.no_guidance_added_html") if what_happens_next_form.what_happens_next_markdown.blank?
+
+      GovukFormsMarkdown.render(what_happens_next_form.what_happens_next_text)
     end
   end
 end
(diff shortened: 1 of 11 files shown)


spec/requests/forms/what_happens_next_controller_spec.rb
```diff
@@ -104,5 +108,107 @@
...
+          expect(form).to have_been_read
+        end
+
+        it "renders the template" do
+          expect(response).to have_rendered("forms/what_happens_next/new")
+        end
+
+        it "returns 200" do
+          expect(response).to have_http_status(:ok)
+        end
```
aliuk2012, line 146: 🤔  should this be 422 because the data is invalid?

spec/requests/forms/what_happens_next_controller_spec.rb
```diff
@@ -104,5 +108,107 @@
...
+
+    context "when markdown is blank" do
+      let(:markdown) { "" }
+
+      it "returns a JSON object containing the converted HTML" do
+        expect(response.body).to eq({ preview_html: I18n.t("guidance.no_guidance_added_html"), errors: [] }.to_json)
+      end
+
+      it "returns 200" do
+        expect(response).to have_http_status(:ok)
```
aliuk2012, line 198: Should this be 422?

DavidBiddle: No - blank markdown is valid in this case (blank text was valid before)

app/controllers/forms/what_happens_next_controller.rb
```diff
@@ -1,14 +1,17 @@
 module Forms
   class WhatHappensNextController < ApplicationController
+    before_action :check_user_has_permission
```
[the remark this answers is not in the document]
DavidBiddle: Doesn't look like it is needed - will remove!

app/controllers/forms/what_happens_next_controller.rb
```diff
@@ -1,26 +1,49 @@
 module Forms
   class WhatHappensNextController < ApplicationController
-    after_action :verify_authorized
```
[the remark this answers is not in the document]
DavidBiddle: Yep, that's it. Thank you!

app/controllers/forms/what_happens_next_controller.rb
```diff
@@ -1,26 +1,52 @@
...
+        if @what_happens_next_form.submit
+          redirect_to form_path(@what_happens_next_form.form), success: t("banner.success.form.what_happens_next_saved")
+        else
+          render :new, status: :unprocessable_entity
+        end
       end
     end
 
+    def render_preview
+      authorize current_form, :can_view_form?
```
aliuk2012, line 33: We need to add a request spec here or do we test this somewhere else? Eg. what happens if the user does not have permission to view this form. What do they see on their page and what actually happens?

DavidBiddle: I don't think we're really doing this anywhere else and to me it would make sense for any tests around the authorisation behaviour to be in the application controller's tests or some separate authentication tests, since that's where the logic is, rather than add tests for this in every controller action. I think for this bit it should be enough to test that the controller is calling `Pundit.authorize` in each action, which I've now added.

aliuk2012 (approved):

PR #798, opened 2023-11-29
Convert delete page button to a link

aliuk2012:

Theres no real reason for us to submit a form to redirect to delete question page. This is part of refactoring that we will be doing to the "Edit Question" page call to actions to make it easier for form creators. There should be no visual change to the user and these changes helps simplify the controllers private method which works out which submit button was pressed.

image

Also something worth noting that as part of these changes it highlighted that we didn't actually have any request specs for delete from "Edit question" page

Trello card: https://trello.com/c/4fXdQIRk/1230-simplify-calls-to-action-ctas-at-the-bottom-of-the-edit-question-page

### DIFF
--- a/spec/views/pages/_forms.html.erb_spec.rb
+++ b/spec/views/pages/_forms.html.erb_spec.rb
@@ -56,7 +56,7 @@ describe "pages/_form.html.erb", type: :view do
     let(:is_new_page) { false }
 
     it "contains a link to add guidance" do
-      expect(rendered).to have_link(text: I18n.t("guidance.add_guidance"), href: guidance_edit_path(form_id: form.id, page_id: page.id))
+      expect(rendered).to have_link(text: I18n.t("guidance.add_guidance"), href: guidance_edit_path(form_id: 1, page_id: 2))
     end
 
     it "has no hidden field for the answer type" do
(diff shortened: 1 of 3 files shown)


spec/views/pages/_forms.html.erb_spec.rb
```diff
@@ -56,15 +56,15 @@
...
+      expect(rendered).to have_link(text: I18n.t("guidance.add_guidance"), href: guidance_edit_path(form_id: 1, page_id: 2))
     end
 
     it "has no hidden field for the answer type" do
       expect(rendered).not_to have_field("question_form[answer_type]", type: :hidden)
     end
 
     it "has a delete button" do
-      expect(rendered).to have_button("delete")
+      expect(rendered).to have_link(text: I18n.t('pages.delete_question'), href: delete_page_path(form_id: form.id, page_id: page.id))
```
DavidBiddle, line 67: This'll fix the linting:

```suggestion
      expect(rendered).to have_link(text: I18n.t("pages.delete_question"), href: delete_page_path(form_id: form.id, page_id: page.id))
```

DavidBiddle (dismissed):
Looks good, just a small linting error that needs fixing

PR #801, opened 2023-11-30
Create a "start new question" route

aliuk2012:

Currently, we hard code a specific page in the add/edit question journey. 

Issues with current implementation which this PR addresses:

- It will give us problems in the future when the add/edit journey is changed and there is a new page that the user should start from. Lots of different places to update and potential to have missed a redirect/link.

- It also doesn't allow us to setup the draft question for a new question (eg. if I were adding or editing a question and decided to "start" again by going to type of answer page, the draft_question would still have all the data for my previous attempt at adding/editing instead of starting fresh. This issue was discovered while working on https://trello.com/c/BV55CFc9/1229-update-save-question-notification-banner-with-calls-to-action-ctas and trying to "Add a new question" from existing "Edit question" page.

Trello card: https://trello.com/c/BV55CFc9/1229-update-save-question-notification-banner-with-calls-to-action-ctas


DavidBiddle:
The code you've written looks good and it works locally, but I think there's another link to the type-of-answer route that needs changing. 

When the user has a draft form with no questions, and they click 'Add and edit your questions', they're taken directly to the new page wizard, starting at the type of answer question. The logic for this is in the form task list service: https://github.com/alphagov/forms-admin/blob/main/app/service/form_task_list_service.rb#L45

aliuk2012:
@DavidBiddle  good catch.....this is exactly why we need a "start new question" path and proof that we should have done it sooner.

PR #803, opened 2023-11-30
Refactor guidance markdown code for reusability

DavidBiddle:

Trello card: https://trello.com/c/AbPj2LRr/1191-allow-markdown-on-confirmation-page-what-happens-next-content

Makes some refactors to the guidance and markdown code to make reusing the component easier. Originally part of #792 

Contains:
- Updating the markdown parameter name used by the AJAX preview to make it more generic
- Making allowing markdown headings configurable, so we can resuse code between guidance and WHN markdown
- Add a markdown validator that can be reused between models

### DIFF
--- a/app/frontend/javascript/ajax-markdown-preview/index.js
+++ b/app/frontend/javascript/ajax-markdown-preview/index.js
@@ -39,7 +39,7 @@ const triggerAjaxMarkdownPreview = async () => {
         redirect: 'follow',
         referrerPolicy: 'same-origin',
         body: JSON.stringify({
-          guidance_markdown: store.source.value,
+          markdown: store.source.value,
           authenticity_token: store.authenticityToken
         })
       })
--- a/app/validators/markdown_validator.rb
+++ b/app/validators/markdown_validator.rb
@@ -1,0 +1,19 @@
+require "govuk_forms_markdown"
+
+class MarkdownValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return true if value.blank?
+
+    markdown_validation = GovukFormsMarkdown.validate(value, allow_headings: options.fetch(:allow_headings))
+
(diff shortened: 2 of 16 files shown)


app/frontend/javascript/ajax-markdown-preview/index.js
```diff
@@ -39,7 +39,7 @@ const triggerAjaxMarkdownPreview = async () => {
         redirect: 'follow',
         referrerPolicy: 'same-origin',
         body: JSON.stringify({
-          guidance_markdown: store.source.value,
+          markdown: store.source.value,
```
aliuk2012, line 42: Should this have broken some js test?

app/validators/markdown_validator.rb
aliuk2012, line 1: I think we should add some unit test for this validator so we arent so reliant on testing everything again in the models

spec/validator/markdown_validator_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      expect(markdown_model).to be_invalid
+      expect(markdown_model.errors.map(&:type)).to include :unsupported_markdown_syntax
+    end
+
+    it "is invalid if guidance markdown contains level 3 headings" do
+      markdown_model.markdown = "### Heading level 3"
+      expect(markdown_model).to be_invalid
+      expect(markdown_model.errors.map(&:type)).to include :unsupported_markdown_syntax
+    end
+  end
```
aliuk2012, line 55: a shared_example might be good here so we could reuse it for the what happens next

aliuk2012 (commented):
A couple of comments but otherwise most of it looks good to go

app/frontend/javascript/ajax-markdown-preview/index.js
DavidBiddle: Good spot - have added one

PR #804, opened 2023-11-30
Deny access to users from restricting organisations

jamie-o-wilkinson:
Some publishing and content teams are concerned about access to trial accounts. This PR adds a denylist list of email domains, and restricts access to users with matching email domains on user creation. This won't affect existing users.

Users that have their access denied by this mechanism will be shown a customised access denied error page, with content to indicate the user should contact their publishing team to arrange access, rather than GOV.UK Forms.
 
Trello card: https://trello.com/c/L6fNhbAM/1222-deny-access-to-new-forms-admin-users-with-specific-email-domains

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -4,6 +4,12 @@ class User < ApplicationRecord
 
   class UserAuthenticationException < StandardError; end
 
+  EMAIL_DOMAIN_DENYLIST = [
+    "dwp.gov.uk",
+    "engineering.dwp.gov.uk",
+    "hse.gov.uk",
+  ].freeze
+
   belongs_to :organisation, optional: true
   has_many :mou_signatures
   has_many :draft_questions
@@ -20,6 +26,10 @@ class User < ApplicationRecord
   validates :role, presence: true
   validates :organisation_id, presence: true, if: :requires_organisation?
   validates :has_access, inclusion: [true, false]
+
+  before_create do
+    self.has_access = false if organisation_restricted_access?
+  end
 
   def self.find_for_gds_oauth(auth_hash)
     find_for_auth(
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -4,6 +4,20 @@ class User < ApplicationRecord
 
   class UserAuthenticationException < StandardError; end
 
+  EMAIL_DOMAIN_DENYLIST = [
+    "dwp.gov.uk",
```
lfdebrux, line 8: It would be nice to be able to test this locally/in dev, but without having to use an email address with a domain in this list.

If we added `digital.cabinet-office.gov.uk` to this list when `Rails.env.development? || Settings.forms_env.to_sym == :development`, we could verify the behaviour by trying to sign in with a `+` email address.

lfdebrux (approved):
Looks good to me, I have a couple of suggestions but nothing blocking!

app/models/user.rb
jamie-o-wilkinson: Hmm interesting suggestion. Wouldn't this make it harder for new team members to use dev, as their actual email address would be blocked by default? Or am I missing something that means our normal email addresses wouldn't be blocked?

lfdebrux: @jamie-o-wilkinson yes it would make it harder for new team members to use dev... I couldn't think of another way of doing it though 😅 The developer strategy sounds nice though, I wonder if it's worth pulling that out into a separate PR @thomasiles?

lfdebrux (approved):

app/models/user.rb
jamie-o-wilkinson: I'm going to merge this PR and then look at this afterwards!

PR #807, opened 2023-12-04
Allow contact details and submission emails to have non-gov.uk domains

lfdebrux:

Trello card: https://trello.com/c/mvZcpnbn/ 

There are a number of government organisations which have domain names that do not end with .gov.uk. We now have users on our service who work for those organisations, as we have added an allow list of domains to the sign up configuration (see alphagov/forms-deploy#440).

We want those users to be able to create forms on our service, including getting form submissions by email. These users should be able to use their government email addresses for the form contact details and submission address.

Rather than adding a duplicate of the allow list to this app, this commit changes the `AllowedEmailDomainValidator` to allow any email address that has the same domain as the user. This may be an imperfect solution, in that it limits users more than a strict allow list would, but is easier to maintain until we have a way to share the list better.

This PR also does a bunch of refactoring of the related code; if people prefer I can split that out to a different PR first.


thomasiles (dismissed):
nice tidying

app/controllers/forms/contact_details_controller.rb
```diff
@@ -13,14 +13,14 @@ def create
...
-        render :new
+        render :new, status: :unprocessable_entity
       end
     end
 
   private
 
     def contact_details_form_params
-      params.require(:forms_contact_details_form).permit(:email, :phone, :link_href, :link_text, contact_details_supplied: []).merge(form: current_form)
+      params.require(:forms_contact_details_form).permit(:email, :phone, :link_href, :link_text, contact_details_supplied: []).merge(form: current_form).merge(current_user:)
```
aliuk2012, line 23:
```suggestion
      params.require(:forms_contact_details_form).permit(:email, :phone, :link_href, :link_text, contact_details_supplied: []).merge(form: current_form, current_user:)
```

spec/validator/allowed_email_domain_validator_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  attr_accessor :current_user, :email
+
+  validates :email, allowed_email_domain: true
+end
+
+RSpec.describe AllowedEmailDomainValidator do
+  let(:model) { Validatable.new }
+
+  it "validates email with .gov.uk" do
+    model.email = "test.gov.uk"
```
aliuk2012, line 21: this doesn't look like a valid email address. is this correct?

app/validators/allowed_email_domain_validator.rb
```diff
@@ -0,0 +1,16 @@
+class AllowedEmailDomainValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    if value.present?
+      return if value =~ /\.gov\.uk\z/i
+
+      # TODO: we might not want to check agiainst current user domain
```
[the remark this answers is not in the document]
lfdebrux: Woops, thanks!

spec/validator/allowed_email_domain_validator_spec.rb
lfdebrux: Good catch, I'll fix that

app/validators/allowed_email_domain_validator.rb
```diff
@@ -0,0 +1,16 @@
+class AllowedEmailDomainValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    if value.present?
+      return if value =~ /\.gov\.uk\z/i
+
+      # TODO: we might not want to check against current user domain
+      # when we have a proper allow list?
+      if record.respond_to?(:current_user) && record.current_user.present?
+        user_domain = record.current_user.email.then { |email| email[email.rindex("@")..] }
+        return if user_domain && value.end_with?(user_domain)
```
DavidBiddle, line 10: Will this be vulnerable to malicious domains that end with the user's domain? i.e. if the user is `david@agovernmentwebsite.co.uk` could the submission email be`hacker@notagovernmentwebsite.co.uk`?

lfdebrux: I don't _think_ so, I tried to check that with the test at https://github.com/alphagov/forms-admin/blob/9f93faa39294ee76f9c6bd123891a61ea7c2324f/spec/validator/allowed_email_domain_validator_spec.rb#L49-L52

lfdebrux: I've tweaked the code here slightly to make it clearer that the string should include the `@` symbol

DavidBiddle: Sorry, didn't spot that test! That change is helpful though so thanks for updating

app/validators/allowed_email_domain_validator.rb
```diff
@@ -0,0 +1,17 @@
...
+  def validate_each(record, attribute, value)
+    if value.present?
+      return if value =~ /\.gov\.uk\z/i
+
+      # TODO: we might not want to check against current user domain
+      # when we have a proper allow list?
+      if record.respond_to?(:current_user) && record.current_user.present?
+        user_domain_with_at = record.current_user.email
+          .then { |email| email[email.rindex("@")..] }
+        return if user_domain && value.end_with?(user_domain_with_at)
```
DavidBiddle, line 11: This reference to `user_domain` wasn't updated and I think that's what's causing your test failure:
```suggestion
        return if user_domain_with_at && value.end_with?(user_domain_with_at)
```

lfdebrux: Thank you! Fixed now.

DavidBiddle (approved):
Really good work! 🎉

PR #810, opened 2023-12-05
Rename form objects module

lfdebrux:

We use the form object pattern in our apps, and use the traditional name for that module, `forms`, however in our service that name is already used in a lot of places.

This commit renames the form objects module to `FormObjects`, to disambiguate. Following Rails naming conventions, the following renames are done:

- `app/forms` becomes `app/form_objects`
- `spec/forms` becomes `spec/form_objects`
- `spec/factories/forms` becomes `spec/factories/form_objects`

Because Rails autoloads most modules, this doesn't actually change much other code, and shouldn't break anything.


DavidBiddle (approved):
Good change and I don't see any issues locally 🎉

DavidBiddle (approved):

PR #812, opened 2023-12-05
Add super_admin cases to forms policy specs

thomasiles:
### Improve the form policy specs before making changes

This is preparing the way to allow super_admins to view and access all forms.

There is no test showing that super_admin users can only access forms owned by their organisation.

This commit adds super_admin to the existing tests for editors and cleans up the tests removing the active resource network call mocks.

I've chosen to remove the mocks as the details of how active resource fetches the objects doesn't seem relevant to testing the policy.

Trello card: https://trello.com/c/iZ4Itleq/1226-without-autocomplete-update-forms-list-page-so-that-super-admins-can-see-edit-all-forms-across-all-departments

### DIFF
--- a/spec/policies/form_policy_spec.rb
+++ b/spec/policies/form_policy_spec.rb
@@ -26,20 +26,24 @@ describe FormPolicy do
   end
 
   describe "#can_view_form?" do
-    context "with a form editor" do
-      it { is_expected.to permit_actions(%i[can_view_form]) }
+    (User.roles.keys - %w[trial]).each do |role|
+      context "with a form #{role}" do
+        let(:user) { build :user, role:, organisation: }
 
-      context "but from another organisation" do
-        let(:organisation) { build :organisation, id: 2, slug: "non-gds" }
+        it { is_expected.to permit_actions(%i[can_view_form]) }
 
-        it { is_expected.to forbid_actions(%i[can_view_form]) }
-      end
+        context "but from another organisation" do
+          let(:organisation) { build :organisation, id: 2, slug: "non-gds" }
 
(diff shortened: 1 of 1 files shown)


spec/policies/form_policy_spec.rb
```diff
@@ -26,20 +26,24 @@
   end
 
   describe "#can_view_form?" do
-    context "with a form editor" do
-      it { is_expected.to permit_actions(%i[can_view_form]) }
+    (User.roles.keys - %w[trial]).each do |role|
```
thomasiles, line 29: Is it worth adding a class constant to the User model to do this for us?

Something like:
```ruby
APPROVED_ROLES = [:editor, :super_admin]
```

spec/policies/form_policy_spec.rb
```diff
@@ -146,105 +150,54 @@
...
-        let(:user) { build :user, :with_trial_role }
+          before do
+            allow(scope).to receive(:where).with(organisation_id: user.organisation.id)
+          end
 
-        it "does not an error" do
-          expect { policy }.not_to raise_error
+          it "returns only their organisation records" do
+            policy_scope.resolve
+            expect(scope).to have_received(:where).with(organisation_id: user.organisation.id)
```
thomasiles, line 179: Testing against the methods called on a double of Form feels less realistic than than using the HTTP mock.

I think chopping off the scope of the tests at this point is better though.

The details of how the `Form` class fetches objects shouldn't be part of these tests. We were mocking the return values to those calls anyway. They weren't testing anything more solid as we were supplying the answers.

PR #819, opened 2023-12-07
Add client-side data scrubbing to Sentry configuration

lfdebrux:

Trello card: https://trello.com/c/EW20O2Zx 

This PR adds code to the Sentry initializer that filters out anything that looks like an email address before it is sent to the Sentry.io servers.

I've tested this locally by using the DSN from the our `forms-debugging-localhost-instances` project. If you have access to our Sentry installation you can see examples of events with scrubbed data at https://govuk-forms.sentry.io/issues/4706430125/ and https://govuk-forms.sentry.io/issues/4706430136/. Note that you may need to look at the latest event for those issues to see the client-side filtering in action.

Although I have made sure that the replacement text used to mask out sensitive data client-side is different to the one used for server-side scrubbing, the server-side scrubbing is also very diligent, and will filter out values if the key contains the term "email" even when the value doesn't look like an email address. This does mean that in production usage we might have occasions where we're not sure if the client-side filtering got to the email address first or not; we might want to have a think about that further.

This PR also includes automated tests for the filter; as well as testing the logic of the filter itself we test it's integration with Sentry. Writing these tests was pretty hard-going, note that we had to add a bit of test specific logic to the Sentry configuration to make these tests work, as well as be very careful about how we reach into the Sentry code to test the behaviour we're interested in.

The filter is pretty thorough, it uses a regex to find anything that looks like a valid email address anywhere in the Sentry event object. The regex comes from https://www.regular-expressions.info/email.html, and covers 99% of valid email addresses. There is probably a bit of unnecessary cycles being spent here, however as this code should only be invoked when there is an exception, I think that's acceptable. Also, in production Sentry runs in a background thread so the CPU time being used shouldn't affect threads serving users unless the server is already close to capacity.

### DIFF
--- a/app/lib/email_parameter_filter_proc.rb
+++ b/app/lib/email_parameter_filter_proc.rb
@@ -1,0 +1,12 @@
+class EmailParameterFilterProc
+  # email regexp from regular-expressions.info/email.html
+  EMAIL_REGEXP = /\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}\b/i
+
+  def self.new(mask: ActiveSupport::ParameterFilter::FILTERED)
+    @mask = mask
+
+    proc do |_key, value|
+      value.is_a?(String) ? value.gsub!(EMAIL_REGEXP, @mask) : value
+    end
+  end
+end
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,3 +1,7 @@
+require "active_support/parameter_filter"
+
+require "./app/lib/email_parameter_filter_proc"
+
 if Settings.sentry.dsn.present?
   Sentry.init do |config|
     config.dsn = Settings.sentry.dsn
@@ -5,6 +9,17 @@ if Settings.sentry.dsn.present?
     config.debug = true
     config.enable_tracing = false
     config.environment = Settings.sentry.environment
+
+    # use synchronous/blocking code for integration tests
+    config.background_worker_threads = 0 if Rails.env.test?
+
+    filter = ActiveSupport::ParameterFilter.new(
+      [EmailParameterFilterProc.new(mask: Settings.sentry.filter_mask)],
+      mask: Settings.sentry.filter_mask,
+    )
+    config.before_send = lambda do |event, _hint|
+      filter.filter(event.to_hash)
+    end
   end
 end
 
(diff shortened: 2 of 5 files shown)


app/lib/email_parameter_filter_proc.rb
```diff
@@ -0,0 +1,12 @@
+class EmailParameterFilterProc
+  # email regexp from regular-expressions.info/email.html
+  EMAIL_REGEXP = /\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}\b/i
```
DavidBiddle, line 3: Optional, but you could use `URI::MailTo::EMAIL_REGEXP` from Ruby's standard library, which is equivalent to the more longwinded:
`/\A[a-zA-Z0-9.!\#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*\z/`

This would mean not having to maintain this RegEx, plus would align with how we're validating form fillers' email addresses in forms-runner.
```suggestion
  EMAIL_REGEXP = URI::MailTo::EMAIL_REGEXP
```

lfdebrux: Oooh, I was wondering if there was something in the standard library for that! Your google-fu is obviously better than mine 😂 I couldn't find this documented anywhere so didn't realise this exists.

lfdebrux: I'll make that change.

lfdebrux: Ah, the URI::MailTo regexp has anchors that expects the whole string to be an email address 🤔

app/views/live/show_form.html.erb
```diff
@@ -13,6 +13,8 @@
 
     <h2 class="govuk-heading-l">Your form</h2>
 
+    <p><%= form.has_live_version %></p>
```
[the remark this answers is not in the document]
lfdebrux: Gone now!

app/lib/email_parameter_filter_proc.rb
DavidBiddle: Ah, OK. Nothing wrong with the way you wewre doing it originally 😁

DavidBiddle (approved):
Code looks good (well done on makiong this testable!). Can't easily test the change on my machine as I don't have Sentry set up locally, but I know you've tested on dev. I've run the tests locally and they pass, and it's not broken anything else as far as I can tell.

PR #830, opened 2023-12-12
Add Rake task to create custom organisation in database

lfdebrux:

Currently when we want to add an organisation to the forms-admin database we either have to rely on it being in the GOV.UK organisations API, or use the AWS data API and an SQL command to add it [[1](https://github.com/alphagov/forms-team/wiki/How-to-add-an-organisation-to-GOV.UK-Forms)].

This PR adds a Rake task we can run using the forms-cli, which should be safer and easier.

I've tested this on my local machine.

### DIFF
--- a/lib/tasks/organisations.rake
+++ b/lib/tasks/organisations.rake
@@ -10,4 +10,20 @@ namespace :organisations do
       OrganisationsFetcher.new.call
     end
   end
+
+  desc "Add organisation that is not in GOV.UK organisation database"
+  task :create, %i[name] => :environment do |_, args|
+    usage_message = "usage: rails organisations:create[<name>]".freeze
+    abort usage_message if args[:name].blank?
+
+    organisation = Organisation.find_by(name: args[:name])
+    abort "Organisation already exists: #{organisation.inspect}" if organisation
+
+    organisation = Organisation.create!(
+      slug: args[:name].parameterize,
+      name: args[:name],
+    )
+
+    puts "Created #{organisation.inspect}"
+  end
 end
--- a/spec/lib/tasks/organisations.rake_spec.rb
+++ b/spec/lib/tasks/organisations.rake_spec.rb
@@ -1,0 +1,55 @@
+require "rake"
+
+require "rails_helper"
+
+RSpec.describe "organisations.rake" do
+  before do
+    # Rake.application.options.trace = true
+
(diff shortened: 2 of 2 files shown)


lib/tasks/organisations.rake
```diff
@@ -10,4 +10,17 @@ namespace :organisations do
       OrganisationsFetcher.new.call
     end
   end
+
+  desc "Add organisation that is not in GOV.UK organisation database"
+  task :create, %i[name] => :environment do |_, args|
+    usage_message = 'usage: rails organisations:create["<name>"]'.freeze
```
DavidBiddle, line 16: I might be misunderstanding this usage message - running `rails organisations:create["Department of Seasonal Events"]` returns `zsh: no matches found: [Department of Seasonal Events]` for me

DavidBiddle: Ah yes, it works in bash for me too. And it works in zsh when you escape the square brackets:

```zsh
bundle exec rails organisations:create\["Department of Seasonal Events"\]
```

lfdebrux:
> It might benefit from a test - in forms-api we have a test for the un-living a form rake task (https://github.com/alphagov/forms-api/blob/main/spec/lib/tasks/forms_admin_spec.rb) - this might be worth doing here.

Ooh I was wondering if it was possible to write a test, this is very helpful, thank you!

> You were probably already planning to do this but just in case - when this is merged please could you update the guidance on the ['How-to-add-an-organisation-to-GOV.UK-Forms' wiki page](https://github.com/alphagov/forms-team/wiki/How-to-add-an-organisation-to-GOV.UK-Forms) to use this task?

Yep will do :D

lib/tasks/organisations.rake
lfdebrux: That's good to know. I hope it might also work with `"bundle exec rails organisations:create[Department of Seasonal Events]"`.

Thanks for testing it!

DavidBiddle: Just tested `bundle exec rails "organisations:create[Department of Seasonal Events]"` and can confirm it works 👍

lfdebrux:
@DavidBiddle I've added tests and also tweaked the task slightly, if you have time to re-review that would be appreciated 🙏

DavidBiddle (approved):
Good tests! Task runs locally, code looks good 👍

PR #831, opened 2023-12-12
Change super admin scope

thomasiles:
### Allow Super admins to view forms accross organisations

Trello card https://trello.com/c/iZ4Itleq/1226-without-autocomplete-update-forms-list-page-so-that-super-admins-can-see-edit-all-forms-across-all-departments

image

This PR gives super admins the permissions to list, view and modify all forms.

It also changes the super admins home page to let them easily list forms from other organisations. There currently isn't a way for them to list forms owned by trial users which are not owned by an organisation.

A few points to note:
- Super admins see an extra column, Created by, which shows which user first created the form. If that isn't available, "Unknown" is shown. Editors and trial users don't see the column.
- The list of organisations only shows organisations with a least one user to cut down on the size of the list.

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -1,8 +1,15 @@
 class FormsController < ApplicationController
   after_action :verify_authorized, except: :index
   after_action :verify_policy_scoped, only: :index
+
   def index
-    @forms = policy_scope(Form) || []
+    if @current_user.super_admin?
+      @search_form = Forms::SearchForm.new({ organisation_id: @current_user.organisation_id }.merge(search_params))
+
+      @forms = policy_scope(Form).where(organisation_id: @search_form.organisation_id) || []
+    else
+      @forms = policy_scope(Form) || []
+    end
   end
 
   def show
@@ -36,5 +43,9 @@ private
 
   def mark_complete_form_params
     params.require(:forms_mark_complete_form).permit(:mark_complete).merge(form: current_form)
+  end
+
+  def search_params
+    params.permit(:organisation_id)
   end
 end
--- a/app/form_objects/forms/search_form.rb
+++ b/app/form_objects/forms/search_form.rb
@@ -1,0 +1,7 @@
+class Forms::SearchForm < BaseForm
+  attr_accessor :organisation_id
+
+  def organisation
+    Organisation.find_by(id: organisation_id)
+  end
+end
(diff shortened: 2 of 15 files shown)


hannahkc:
Hi hi! The line above the organisation selection field should read: 

**Change which organisation’s forms you’re viewing**

Also I note that there's an additional heading for the table (**<Org name> forms**) so we have the same heading twice. Presumably this is for accessibility reasons? I think we probably want to keep these two titles unique from each other though - also for accessibility? 

Just thinking we could change the first one to: **You’re viewing <Org name> forms**
That would tie in nicer with the line of text beneath it as well as making it unique from the table heading.

What you think? cc. @hazalarpalikli

thomasiles:
thanks - I'll update the hint text.

You are right, I kept the table caption (the second heading) as it describes the table. The first one is the really the label for the select list so can't be used for the overall list of forms. Is something more question texty better?

hazalarpalikli:
Yes good shout @hannahkc , it would make more sense to change the first title 'you are reviewing' especially for the screen readers. i'll make sure to update the designs as well with this

hazalarpalikli:
Group 1245

Just did a quick mock up. I second what Tom says maybe the first title needs to be a question?

hannahkc:
Oh sorry - I put some angle brackets in my previous message that made the text disappear. Sorry!

To clarify, I think the content should be (using GDS as the org example): 

You're viewing Government Digital Service forms

Change which organisation’s forms you’re viewing

Government Digital Service forms

hazalarpalikli:
Yes it makes sense to me now :) So it needs to look like this
Group 1245

hannahkc:
**You're** NOT **you are** please.

hazalarpalikli:
Group 1245
Opps thank you! @hannahkc  Deleted the previous comment to avoid confusion :) this is how it needs to look @thomasiles

DavidBiddle (approved):
Code looks good and it's very slick on my local machine 🎉

app/controllers/forms_controller.rb
```diff
@@ -1,8 +1,15 @@
 class FormsController < ApplicationController
   after_action :verify_authorized, except: :index
   after_action :verify_policy_scoped, only: :index
+
   def index
-    @forms = policy_scope(Form) || []
+    if @current_user.super_admin?
+      @search_form = Forms::SearchForm.new({ organisation_id: @current_user.organisation_id }.merge(search_params))
```
lfdebrux, line 7: I'm not sure I understand why we've created a form class here. Why is this the best way to get the organisation ID out?

PR #836, opened 2023-12-15
Fix slow performance on forms#index

thomasiles:
When display the list of organisations, it only shows the organisations which have at least one user. This cuts the list down from every organisation in the government to just the ones we care about when filtering.

The current scope produces the following query:

```sql
SELECT DISTINCT "organisations".* FROM "organisations" INNER JOIN "users" ON "users"."organisation_id" = "organisations"."id" ORDER BY "organisations"."slug" ASC
```

Changing the scope to

```ruby
scope :with_users, -> { where(User.where('organisation_id = organisations.id').arel.exists) }
```

Produces the following SQL:

```sql
SELECT "organisations".* FROM "organisations" WHERE EXISTS (SELECT "users".* FROM "users" WHERE (organisation_id = organisations.id)) ORDER BY "organisations"."slug" ASC
```

I think that should be much quicker but it’s hard to test locally as I don’t have many users. Running explain analyse on both gives these results:

```
Old way:
Unique  (cost=44.30..48.33 rows=230 width=120) (actual time=0.056..0.063 rows=3 loops=1)
  ->  Sort  (cost=44.30..44.88 rows=230 width=120) (actual time=0.056..0.057 rows=9 loops=1)
        Sort Key: organisations.slug, organisations.id, organisations.govuk_content_id, organisations.name, organisations.created_at, organisations.updated_at
        Sort Method: quicksort  Memory: 26kB
        ->  Hash Join  (cost=22.38..35.28 rows=230 width=120) (actual time=0.017..0.026 rows=9 loops=1)
              Hash Cond: (users.organisation_id = organisations.id)
              ->  Seq Scan on users  (cost=0.00..12.30 rows=230 width=8) (actual time=0.006..0.008 rows=21 loops=1)
              ->  Hash  (cost=15.50..15.50 rows=550 width=120) (actual time=0.005..0.006 rows=4 loops=1)
                    Buckets: 1024  Batches: 1  Memory Usage: 9kB
                    ->  Seq Scan on organisations  (cost=0.00..15.50 rows=550 width=120) (actual time=0.002..0.003 rows=4 loops=1)
Planning Time: 0.175 ms
Execution Time: 0.111 ms

New way:
Sort  (cost=45.90..46.47 rows=230 width=120) (actual time=0.052..0.054 rows=3 loops=1)
  Sort Key: organisations.slug
  Sort Method: quicksort  Memory: 25kB
  ->  Hash Join  (cost=17.38..36.88 rows=230 width=120) (actual time=0.041..0.045 rows=3 loops=1)
        Hash Cond: (organisations.id = users.organisation_id)
        ->  Seq Scan on organisations  (cost=0.00..15.50 rows=550 width=120) (actual time=0.010..0.010 rows=4 loops=1)
        ->  Hash  (cost=14.88..14.88 rows=200 width=8) (actual time=0.027..0.027 rows=3 loops=1)
              Buckets: 1024  Batches: 1  Memory Usage: 9kB
              ->  HashAggregate  (cost=12.88..14.88 rows=200 width=8) (actual time=0.023..0.025 rows=4 loops=1)
                    Group Key: users.organisation_id
                    Batches: 1  Memory Usage: 40kB
                    ->  Seq Scan on users  (cost=0.00..12.30 rows=230 width=8) (actual time=0.007..0.012 rows=21 loops=1)
Planning Time: 0.141 ms
Execution Time: 0.084 ms
```

Trello card: 

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -6,5 +6,5 @@ class Organisation < ApplicationRecord
 
   has_many :mou_signatures
 
-  scope :with_users, -> { joins(:users).distinct }
+  scope :with_users, -> { where(User.where("organisation_id = organisations.id").arel.exists) }
 end


chao-xian (approved):
Looks good

PR #845, opened 2023-12-19
Add feature flag and a new mailer

aliuk2012:

This PR lays the ground work for the feature which will notify a form processing team that a live forms submission email address has changed and not to expect further form submissions. 

Visiting `http://localhost:3000/rails/mailers` you will see a "preview" of the mailer and an example of whats personalisation is sent to GOV.UK notify

Another PR will follow which will integrate the feature flag and the mailer into the service.

### DIFF
--- a/app/mailers/submission_email_mailer.rb
+++ b/app/mailers/submission_email_mailer.rb
@@ -13,4 +13,16 @@ class SubmissionEmailMailer < GovukNotifyRails::Mailer
 
     mail(to: new_submission_email)
   end
+
+  def notify_submission_email_has_changed(live_email:, form_name:, current_user:)
+    set_template(Settings.govuk_notify.live_submission_email_of_no_further_form_submissions)
+
+    set_personalisation(
+      form_creator_name: current_user.name,
+      form_creator_email: current_user.email,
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -151,7 +151,7 @@ en:
 
       %{link} if you think this is incorrect.
     body_html_org_restricted: Your GOV.UK publishing team is managing the creation of forms for your department. Please contact them to discuss your form needs.
-    link_text: Contact the GOV⁠.⁠UK Forms team
+    link_text: Contact the GOV.UK Forms team
```
aliuk2012, line 154: This is interesting, this is what it looked like in my IDE. 
image

lfdebrux: These are word joiner characters, used to prevent line breaks between 'GOV', '.', and 'UK' [[1](https://en.wikipedia.org/wiki/Word_joiner)].

aliuk2012: Removed the commit now.

thomasiles (approved):
looks good 👍 

I guess this isn't a journey which needs testing using in the e2e tests?

PR #847, opened 2023-12-19
Introduce MakeFormLive service

aliuk2012:
### Relates to 
- https://github.com/alphagov/forms-admin/pull/845

Instead of Form Object having responsibility for making a form live and any other events that occured as a result of those changes, it would be much better for a service to do it instead.

This PR introduces a service which (for the time being),
- make a form live
- calculate what page title should be returned


thomasiles (dismissed):
looks nice 🏅

thomasiles (approved):

PR #849, opened 2023-12-20
update MakeFormLiveService to notify live form submission if the email has changed

aliuk2012:

This PR updates the existing `MakeFormLiveService` to send an email to the live forms submission address, as a means of informing the form processing team that the live form now has a new submission email address and no further emails with be sent to that email address.

This is behind a feature flag which is disabled at the moment. I tested this in dev with a forms-deploy branch that enabled the feature and it worked without any issues.

Trello card: https://trello.com/c/XletoC1y/1235-notify-a-confirmed-submission-email-that-it-will-no-longer-be-receiving-form-submissions-and-update-confirmation-page-content

### DIFF
--- a/app/controllers/forms/make_live_controller.rb
+++ b/app/controllers/forms/make_live_controller.rb
@@ -14,7 +14,7 @@ module Forms
 
       return redirect_to form_path(@make_live_form.form) unless user_wants_to_make_form_live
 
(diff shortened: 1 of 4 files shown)


SamJamCul (approved):
Looking good!

PR #850, opened 2023-12-21
Add org select for superadmins

thomasiles:

Trello card: https://trello.com/c/iZ4Itleq/1226-without-autocomplete-update-forms-list-page-so-that-super-admins-can-see-edit-all-forms-across-all-departments

Add a select list of organisations with users for super_admins to choose from on forms#index.

image

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -5,4 +5,6 @@ class Organisation < ApplicationRecord
   has_many :users
 
   has_many :mou_signatures
+
+  scope :with_users, -> { where(User.where("organisation_id = organisations.id").arel.exists) }
 end
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -6,5 +6,5 @@ class Organisation < ApplicationRecord
 
   has_many :mou_signatures
 
-  scope :with_users, -> { where(User.where("organisation_id = organisations.id").arel.exists) }
+  scope :with_users, -> { where(User.where("organisation_id = organisations.id").arel.exists).order(:name) }
 end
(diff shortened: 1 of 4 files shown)


app/models/organisation.rb
```diff
@@ -5,4 +5,6 @@ class Organisation < ApplicationRecord
   has_many :users
 
   has_many :mou_signatures
+
+  scope :with_users, -> { where(User.where("organisation_id = organisations.id").arel.exists) }
```
aliuk2012, line 9: Does orgs have a default sort? I'm wondering if it would be worth adding a sort here to make sure its in some sort of order

spec/models/organisation_spec.rb
```diff
@@ -27,4 +27,23 @@
...
+        organisation2 = FactoryBot.create(:organisation, slug: "org_2")
+
+        FactoryBot.create(:user, organisation: organisation1)
+        FactoryBot.create(:user, organisation: organisation1)
+        FactoryBot.create(:user, organisation: organisation2)
+
+        organisations_with_users = described_class.with_users
+
+        expect(organisations_with_users).to include(organisation1)
+        expect(organisations_with_users).to include(organisation2)
```
aliuk2012, line 44: If you add a sort to the scope, it might be worth testing the sort here by having a single expect that matches the orgs in a list.

app/models/organisation.rb
aliuk2012: 🤔  does it matter if the user is an editor or trial account?

aliuk2012 (approved):

PR #860, opened 2023-12-28
Inform form creators that we will email form processing teams when submission email address has changed

aliuk2012:

<details>
<summary>Screenshot: For first draft or if the submission email has not changed between draft  live version</summary>

image

</details>

<details>
<summary>Screenshot: When submission email has changed between draft  live version</summary>

image

</details>

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -94,6 +94,10 @@ class ApplicationController < ActionController::Base
     @current_form ||= Form.find(params[:form_id])
   end
 
+  def current_live_form
+    @current_live_form ||= Form.find_live(params[:form_id])
+  end
+
 private
 
   def authenticate_and_check_access
--- a/app/controllers/forms/live_controller.rb
+++ b/app/controllers/forms/live_controller.rb
@@ -9,10 +9,4 @@ class Forms::LiveController < ApplicationController
     authorize current_form, :can_view_form?
     render template: "live/show_pages", locals: { form: current_live_form }
   end
-
-private
-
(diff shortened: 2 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -94,6 +94,10 @@ def current_form
     @current_form ||= Form.find(params[:form_id])
   end
 
+  def current_live_form
+    @current_live_form ||= Form.find_live(params[:form_id])
```
DavidBiddle, line 98: What does this return if there's no live form?

app/controllers/application_controller.rb
```diff
@@ -94,6 +94,10 @@ def current_form
     @current_form ||= Form.find(params[:form_id])
   end
 
+  def current_live_form
```
DavidBiddle, line 97: Should this have a test now that it's a public method?

DavidBiddle (commented):
Code looks good and works for me locally. I do have a couple of queries about the `current_live_form` method

app/controllers/application_controller.rb
aliuk2012: It would raise a 404 like `current_form`

app/controllers/application_controller.rb
aliuk2012: I don't think so, We don't test `current_form` and when it's called we stub out the request in the test it relates to so it's being tested indirectly. It's a case of testing behaviour rather than implementation.

DavidBiddle (approved):

PR #870, opened 2024-01-03
Bump the version of the dfe-autocomplete gem

DavidBiddle:

Trello card: 

This commit bumps the dfe-autocomplete version in the Gemfile to match the version we're using in our package.json. In this case I don't think we expect any behaviour changes as a result of this, but the versions should be in sync.


thomasiles (approved):
works for me 👍

PR #877, opened 2024-01-08
Configure autocomplete

thomasiles:
### Add autocomplete for super_admins looking at other org's forms

Trello card: https://trello.com/c/2MPQysX1/1233-add-autocomplete-functionality-to-forms-list-input-text-component-for-orgs

Super admins currently can choose to see other organisations forms using a select box on the homepage. This commit enhances the select box to an autocomplete box.

image

We already use the accessible autocomplete component to choose organisation in the users#edit page.
The autocomplete box on the homepage is also chooses organisation but behaves differently.

It is more like a select box than a text box. There is no way to clear it and submit an empty value, which the users page requires. It also shows all values like a select box to allow the user to list the organisations which are available to choose.

These differences in configuration mean we need a way to change how the autcomplete component is initialised. The component isn't very flexible and [can only be initialised once per page](https://github.com/DFE-Digital/dfe-autocomplete/blob/main/src/wrapper.js).

There are a few ways round this:
- put in a PR for the library to export `setupAccessibleAutoComplete` itself so we can provide custom options per autocomplete
- use different javascript entrypoints on pages which use the component with configuration in
- use a custom script to set the options in views

As we only use one auotcomplete component per page, I went with the last option as it seems the least intrusive. I think the PR approach may be better in the future if we use a lot of autocompletes.

We add the javascript to the bottom of the body and mark it with defer. As modules are loaded using defer, this should ensure our inline script is loaded after the application bundle.

### DIFF
--- a/app/frontend/entrypoints/application.js
+++ b/app/frontend/entrypoints/application.js
@@ -39,15 +39,4 @@ document
 
 initAll()
 
-// We set rawAttribute to true to pass the value of the text input to the
-// controller when the autocomplete is being used, so we can detect when the
-// user has cleared the field. We set source to false to overide the DfE
-// component's handling of sort order and synonyms, as described in the code
-// found in
-// https://github.com/DFE-Digital/dfe-autocomplete/blob/36d80e6b5bba67c92cd9ec6982a4e536d1889aed/src/dfe-autocomplete.js#L54C1-L54C1
-// this means the autocomplete will get options straight from the select
-// element, removing the null option, which isn't needed for autocomplete.
-dfeAutocomplete({
-  rawAttribute: true,
-  source: false
-})
+window.dfeAutocomplete = dfeAutocomplete
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -100,4 +100,22 @@ module ApplicationHelper
       end
     end
   end
+
+  def init_autocomplete_script(show_all_values: false, raw_attribute: false, source: false)
+    content_for(:body_end) do
+      javascript_tag defer: true do
+        "
+      document.addEventListener('DOMContentLoaded', function(event) {
+        if(window.dfeAutocomplete !== undefined && typeof window.dfeAutocomplete === 'function') {
+          dfeAutocomplete({
+            showAllValues: #{show_all_values},
+            rawAttribute: #{raw_attribute},
+            source: #{source}
+          })
+        }
+      });
+        ".html_safe
+      end
+    end
+  end
 end
--- a/spec/features/form/super_admin_view_forms_spec.rb
+++ b/spec/features/form/super_admin_view_forms_spec.rb
@@ -5,7 +5,6 @@ feature "View forms", type: :feature do
   let(:other_org) { create :organisation, id: 2, slug: "Other org" }
   let(:other_org_user) { create :user, organisation: other_org }
   let(:other_org_forms) { [build(:form, id: 2, organisation_id: other_org.id, name: "Other org form")] }
-  # let(:other_org_forms) { [build (:form, id: 1, name: "Apply for a juggling license", organisation_id: other_org.id)] }
 
   let(:req_headers) do
     {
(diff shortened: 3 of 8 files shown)


app/frontend/entrypoints/application.js
```diff
@@ -39,15 +39,4 @@ document
...
-// component's handling of sort order and synonyms, as described in the code
-// found in
-// https://github.com/DFE-Digital/dfe-autocomplete/blob/36d80e6b5bba67c92cd9ec6982a4e536d1889aed/src/dfe-autocomplete.js#L54C1-L54C1
-// this means the autocomplete will get options straight from the select
-// element, removing the null option, which isn't needed for autocomplete.
-dfeAutocomplete({
-  rawAttribute: true,
-  source: false
-})
+window.dfeAutocomplete = dfeAutocomplete
```
thomasiles, line 42: put the dfeAutocomplete function into the window object. This allows us to call it with different options from inline scripts in views.

This breaks encapsulation but I don't think there is a better way.

app/helpers/application_helper.rb
```diff
@@ -100,4 +100,22 @@ def sign_in_params(is_e2e_user:, login_type: :sign_in)
...
+            showAllValues: #{show_all_values},
+            rawAttribute: #{raw_attribute},
+            source: #{source}
+          })
+        }
+      });
+        ".html_safe
+      end
+    end
+  end
```
thomasiles, line 120: This is the code views can use to initialise autocomplete components with their own options from views.

We use defer on the script tag to queue it behind our application bundle and use DOMContentLoaded to ensure it doesn't run early. I don't think it ever would given the script is at the end of the body and marked with defer.

DavidBiddle: Neat!

spec/features/form/super_admin_view_forms_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+feature "View forms", type: :feature do
+  let(:org_forms) { [build(:form, id: 1, name: "Org form")] }
+  let(:other_org) { create :organisation, id: 2, slug: "Other org" }
+  let(:other_org_user) { create :user, organisation: other_org }
+  let(:other_org_forms) { [build(:form, id: 2, organisation_id: other_org.id, name: "Other org form")] }
+  # let(:other_org_forms) { [build (:form, id: 1, name: "Apply for a juggling license", organisation_id: other_org.id)] }
```
DavidBiddle, line 8: Is this commented out line still needed?

DavidBiddle (dismissed):
Thanks for explaining the reasons for this approach so clearly! I think it would still be worth creating an issue in the dfe-autocomplete repo asking them to export `setupAccessibleAutoComplete`, so that we can bring this JS back into our bundle at some point, but this approach is good in the mean time / if that isn't something they want to prioritise.

Code looks good and all runs OK locally 🎉 I've raised one minor comment about a commented out line which might be removable but otherwise looks great

app/frontend/entrypoints/application.js
lfdebrux: Is it worth raising a feature request at https://github.com/DFE-Digital/dfe-autocomplete/ asking for the ability to configure the component with `data-` attributes or something similar?

lfdebrux: Or just raising the problem of `setupAccessibleAutocomplete` not being exported as an issue, like you said in the PR comment and @DavidBiddle said 😅

lfdebrux (commented):
LGTM!

app/frontend/entrypoints/application.js
thomasiles: yeah - I think data- attributes are the way to go but probably a bit much to ask for :) I'll stick a PR in as you both suggest

spec/features/form/super_admin_view_forms_spec.rb
thomasiles: no! good spot I'll get rid

thomasiles: it's gone 👍

DavidBiddle (approved):

PR #886, opened 2024-01-11
Modify Development Database Config

danworth:
These changes allow for forms-admin to connect to postgres running locally without needing to specify any additional environment variables.

Running the apps locally.


DavidBiddle (approved):
Change looks good and code seems to work for me locally 🎉

PR #894, opened 2024-01-18
Add 'Created by' column for super admins in form list

aliuk2012:

This update adds a new column 'Created by' to the form list which is visible only to super admin users. They will now be able to see who created each form. Tests are also updated to ensure proper functionality.

As Form is an ActiveResource and Users are an ActiveRecord we cannot rely on relationships to put this information together. For efficiency reasons, we pull out a list of unique creator ids from the list of forms and use that to query ActiveRecord once to get a list of user id and name which is used to lookup from.

Trello card: 

### DIFF
--- a/app/service/form_list_service.rb
+++ b/app/service/form_list_service.rb
@@ -1,4 +1,3 @@
-# require "helpers/govuk_rails_compatible_link_helper"
 class FormListService
   include Rails.application.routes.url_helpers
   include ActionView::Helpers::UrlHelper
@@ -16,6 +15,12 @@ class FormListService
     @forms = forms
     @current_user = current_user
     @search_form = search_form
+    if current_user.super_admin?
+      @list_of_creator_id = forms.map(&:creator_id).uniq
+      @list_of_creators = User.where(id: @list_of_creator_id)
+                               .select(:id, :name)
+                               .map { |user| { id: user.id, name: user.name } }
+    end
   end
 
   def data
@@ -35,13 +40,18 @@ private
   end
 
   def head
-    [I18n.t("home.form_name_heading"),
-     { text: I18n.t("home.form_status_heading"), numeric: true }]
+    [
+      I18n.t("home.form_name_heading"),
+      (current_user.super_admin? ? { text: I18n.t("home.created_by") } : nil),
+      { text: I18n.t("home.form_status_heading"), numeric: true },
+    ].compact
   end
 
   def rows
     forms.map do |form|
-      [{ text: form_name_link(form) }, { text: form_status_tags(form), numeric: true }]
+      [{ text: form_name_link(form) },
+       (current_user.super_admin? ? { text: find_creator_name(form) } : nil),
+       { text: form_status_tags(form), numeric: true }].compact
     end
   end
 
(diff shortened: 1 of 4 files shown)


spec/service/form_list_service_spec.rb
```diff
@@ -61,6 +73,28 @@
           ])
         end
       end
+
+      context "when user is super admin" do
```
jamie-o-wilkinson, line 77: Should we have a test to check the new column is not available for other roles?

aliuk2012: Which other roles? 
By default it should have two as tested here https://github.com/alphagov/forms-admin/pull/894/files#diff-c5ae5556f8c48f48fff536926172d25d520905e4ac7de3bb3681586b3ecaa482R64

this "super admin" restriction is only a short term solution we will remove it once we have tested this in prod to see if there are any performance impact

jamie-o-wilkinson: I was thinking of trial and editor, but of course neither of those can do anything with users anyway (`can_manage_user?` in https://github.com/alphagov/forms-admin/blob/main/app/policies/user_policy.rb#L25), so it probably doesn't matter...

aliuk2012: Actually, when we come to remove the `super_admin` only restriction it might be worth hiding the extra column for `trial_users` because in theory the forms would be created by trial user so a bit pointless having lookups to return the trial users name

jamie-o-wilkinson: My thought was if we're testing that column is visible for super-admins, should we be testing that it is not visible for trial and editor users? (sorry I got confused between the form and users list 😅 )

aliuk2012: We are already testing a "default" number of columns here https://github.com/alphagov/forms-admin/pull/894/files#diff-c5ae5556f8c48f48fff536926172d25d520905e4ac7de3bb3681586b3ecaa482R60-R75

My thought is that when we come to removing the super-admin conditions in the service, these tests would fail and would have to be updating to reflect the new default behaviour. In  this PR, we are only testing the exception of a user being a super-admin all other types should only have 2 columns which is the default.

jamie-o-wilkinson: Gotcha, sorry I hadn't parsed that test properly. In that case, LGTM!

jamie-o-wilkinson (approved):

app/service/form_list_service.rb
```diff
@@ -16,6 +15,12 @@ def initialize(forms:, current_user:, search_form: nil)
     @forms = forms
     @current_user = current_user
     @search_form = search_form
+    if current_user.super_admin?
+      @list_of_creator_id = forms.map(&:creator_id).uniq
+      @list_of_creators = User.where(id: @list_of_creator_id)
+                               .select(:id, :name)
+                               .map { |user| { id: user.id, name: user.name } }
+    end
```
thomasiles, line 23: I think this creates queries like this when there is a form which doesn't have a creator set (which I think might still be possible?)

```sql
SELECT "users"."id", "users"."name" FROM "users" WHERE ("users"."id" IN ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20, $21, $22) OR "users"."id" IS NULL)
```

I don't think the `OR "users"."id" IS NULL)` matters because we should. never have a user without an id?

if it does we could add `.compact` to line 19.

Should list_of creators be a __list__ of creators? Would a hash be better? e.g. 

```ruby
User.where(id: @list_of_creator_id).select(:id, :name).pluck(:id, :name).to_h
```
This would simplify find_creator_name a bit.

This is a tiny optimisation though and might be over doing it :)

thomasiles (approved):
looks good to me, nice way to capture the logic of getting creators names 🏅

PR #895, opened 2024-01-18
Switch how we build users by using their respective factory instead of inline editing role attribute or traits

aliuk2012:

This only refactors editors and super admin. I'll do another PR for trial users which has raised some errors in existing tests where the factory that was being generated was in an invalid state and it wasn't picked up because we were either changing only the role or using the trial_user trait.

Trello card: 


jamie-o-wilkinson (approved):
LGTM

PR #901, opened 2024-01-22
Force form states to be displayed inline instead of wrapping

aliuk2012:
Previously, states would wrap and overlap each other because they had no padding. This wrap wraps a flexbox contain around it and forces the two states to always be inline with 5px padding between them.

### Screenshots - Before

image

### Screenshots - After

image

Trello card:  https://trello.com/c/ePWiTjCg/1260-show-which-user-made-each-form


DavidBiddle (approved):
Looks good to me and works locally :tada:

PR #903, opened 2024-01-23
Add AWS application load balancer trace ID to logs

lfdebrux:

We want to be able to correlate our Elastic Load Balancer logs with app logs. This is particularly useful when there is a `460` error because the client has closed the connection for taking too long.

This PR adds the `X-Amzn-Trace-Id` header [[1]] to our request logging.

[1]: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html


danworth (approved):
Brilliant and I like the way you've found to test this logging payload in the request spec (I think this could be extended for the other aspects of the payload.

PR #904, opened 2024-01-23
allow all editors/super admins to see who created a form

aliuk2012:
Originally, only super admins could view "created by" column in the list of forms. This was because we wanted to make sure the application could handle it. If we do deploy this and things go wrong we can easily revert this commit and we can review the situation.

The reason why only editor/super admins can see this column is that trial users should not be able to see forms created by other users (even in the same organisation) and it seemed wasteful to have a column in the report which is populated with their name for each form they create.

### Screenshots - Trial user does not see "created by" column

image

### Screenshot - Editor/Super admin see "created by" column

image

Trello card: 

### DIFF
--- a/app/service/form_list_service.rb
+++ b/app/service/form_list_service.rb
@@ -15,7 +15,7 @@ class FormListService
     @forms = forms
     @current_user = current_user
     @search_form = search_form
-    if current_user.super_admin?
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
nice 👍

PR #905, opened 2024-01-23
Refactor Dockerfile

lfdebrux:

This PR makes a few tweaks to the Dockerfile to make it easier to use with Docker Compose and the `update_app_versions` script in forms-deploy.

If we like these changes I'll make the same to the other app Dockerfiles.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,9 @@ USER ruby
 
 COPY --chown=ruby:ruby Gemfile* ./
 
-RUN bundle config set --local without development:test \
-  && bundle config set --local jobs "$(nproc)"
+ARG BUNDLE_WITHOUT=development:test
+RUN [ -z "$BUNDLE_WITHOUT" ] || bundle config set --local without "$BUNDLE_WITHOUT"
+RUN bundle config set --local jobs "$(nproc)"
 
 RUN bundle install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN bundle install
 COPY --chown=ruby:ruby package.json package-lock.json ./
 RUN npm ci --ignore-scripts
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,11 @@
 FROM ruby:3.2.2-alpine3.18@sha256:198e97ccb12cd0297c274d10e504138f412f90bed50c36ebde0a466ab89cf526 AS build
+ARG NODEJS_VERSION=18
```
lfdebrux, line 2: @AP-Hunt I couldn't decide whether to pull out the other version numbers as well; it occurred to me we could do something like:

```suggestion
ARG ALPINE_VERSION=3.18
ARG NODEJS_VERSION=18
ARG RUBY_VERSION=3.2.2

ARG DOCKER_IMAGE_DIGEST=sha256:198e97ccb12cd0297c274d10e504138f412f90bed50c36ebde0a466ab89cf526

FROM ruby:${RUBY_VERSION}-alpine${ALPINE_VERSION}@${DOCKER_IMAGE_DIGEST} AS build
```

It would make the `update_app_versions` script easier to write, but the Dockerfile harder to read. What do you think?

DavidBiddle: This still feels quite readable to me as a non-expert - and I think these version numbers are the bits that change the most, so it's nice to be able to see them at a glance.

AP-Hunt: I agree with David. The expression on line 7 is more complicated, but splitting the versions into their own args makes for overall simpler maintenance.

lfdebrux: Thanks both, I've redone this commit, re-review would be appreciated 🙏

DavidBiddle (approved):
Looks godd to me and builds locally 💯

PR #915, opened 2024-01-31
Use setup phase to change Auth0 client for E2E tests

jamie-o-wilkinson:
This PR changes forms-admin’s OmniAuth middleware config to allow request-time modification of the auth strategy using the [setup phase](https://github.com/omniauth/omniauth/wiki/Setup-Phase). This lets us examine the request params to determine if it is from the E2E tests, and modify things like the client ID and client secret to use the E2E Auth0 application created in https://github.com/alphagov/forms-deploy/pull/539

Trello card: https://trello.com/c/QYU1ZGjU


thomasiles (dismissed):
looks good - I like the tests.

thomasiles (approved):

PR #922, opened 2024-02-05
Ask user's name and org on their first sign in

thomasiles:
# This is a spike to look at asking new users for their name and organisation

Trello card: https://trello.com/c/hlFB6SvQ/1333-spike-how-should-we-get-users-organisation-and-name-when-they-create-an-account

image

This spike doesn't include tests as I think it's better to finalise the details before writing those.

The card attached and discussion around looks at different options to finding a users name and organisation - two details we need but don't have when they access the system.

This spike looks at asking the user for both when they first sign in. It was decided that we would only ask new users at first, not existing users.

The feature does this by redirecting users to a page asking for name and org when they first login. They can't view forms or access any other page until they have provided this information.

## Content
I've not made any effort with the content - we need:
- page title
- nice question text
- errors for when the user doesn't provide a name or organisation
- when the user submits an organisation which isn't in the list
- the URL is currently `/user_identification`

## Tech details
We currently don't have a way to tell that a sign in is the user's first. Auth0 doesn't provide this information (I don't think) and we don't record it when we create the `User` in the database.

To record this a `sign_in_count` value is added to the `User` model and incremented each time a user logs in. This was stolen from the [devise trackable module](https://github.com/heartcombo/devise/blob/main/lib/devise/models/trackable.rb). A `before_action` is used to check this value on every page load (apart from authentications themselves) to redirect the user to the form. We should probably make sure this doesn't trigger on error pages and other pages where it shouldn't too.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   before_action :check_maintenance_mode_is_enabled
   before_action :authenticate_and_check_access
   before_action :set_paper_trail_whodunnit
+  before_action :run_first_sign_in_flow
 
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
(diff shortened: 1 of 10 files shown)


thomasiles:
This has been replaced by other PRs which implement this feature.

PR #928, opened 2024-02-06
Add very basic groups

thomasiles:
# Add scaffold for groups

Trello card: https://trello.com/c/jAnyz2EM/1364-add-scaffolding-code-for-groups-slice-1

This is a draft pull request for adding the basic concept of groups. The ticket is technical. The design has been thrown together - anything can be changed.

Currently:
- the pages are under the URL `/groups`. There are no links from the main pages.
- groups are not linked to Organisations or Users
- there are no protections stopping any forms user listing/viewing/deleting/editing any groups. We'll add that in later slices. We could limit showing groups to super_admins for now to deploy the feature but I don't think it matters much as at the moment there are no consequences to groups and this commit is already quite big.

I had thought that if we wanted to integrate this code we should be careful about settings page titles and all text to the locale files.

I think we should hold off on this now though. The designs will change the pages we need and the content we need on those pages. I think holding off on those changes is a better approach to get this feature done.

## Index
/groups
This lists all the groups. Clicking the group name takes the user to the show group page. Clicking 'Change' takes them to the edit screen
image

## New group
/groups/new
Name is the only field currently and is required.
image
image

## Show group
/groups/:group_id

This page shows the details of the group but isn't editable. I guess when we add forms and users to groups it might list both?

The delete button destroys the group and doesn't currently have a confirmation step
image

/groups/:group_id/edit
Allows the user to edit the group and change the only current field, name.
image


app/models/group.rb
```diff
@@ -1,3 +1,14 @@
...
+  before_create :set_external_id
+
+  def to_param
+    external_id
+  end
+
+private
+
+  def set_external_id
+    self.external_id = SecureRandom.urlsafe_base64(10)
```
thomasiles, line 12: Whats our best option for generating external ids and do we want pick a standard for other objects which will need it as well, such as Form?

- We want something that's sparse enough that they can't be enumerated by guessing URLs
- They way it's implemented here means we generated a random id, if there is already a Group with that id in the db, a unique error will be thrown when we try and save it but we don't try and solve this. If we choose a random value with enough bytes, we don't have to worry about this happening. It tends to be a higher number than you think due to the [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem)
- This is all internal to the admin, we don't allow clients to generate their own id's at any point so we don't need to overthink coordination
- We don't want URLs to be too long and unwieldy 

`urlsafe_base64` defaults to 16 bytes which is a nice number, from alphabet a-z A-Z 0-9 + / =
[SecureRandom](https://apidock.com/ruby/SecureRandom).[urlsafe_base64](https://apidock.com/ruby/SecureRandom/urlsafe_base64/class) = "b4GOKm4pOYU_-BOXcrUGDg"

Base 58 leaves out like base 64 but leaves out: + / = 0 O I
Defaults to 16 __characters__ long
[SecureRandom](https://apidock.com/rails/SecureRandom).[base58](https://apidock.com/rails/SecureRandom/base58/class) # => "4kUgL2pdQMSCQtjE"

There is a handy calculator here for comparing sizes: https://zelark.github.io/nano-id-cc/

db/migrate/20240208170628_add_external_id_to_group.rb
```diff
@@ -0,0 +1,12 @@
+class AddExternalIdToGroup < ActiveRecord::Migration[7.1]
+  def change
+    # Rubcop will complain that we are making a column not null without a default value
+    # which wouldn't work if we already had data in the table. We are going to ignore this
+    # because we don't have any data in the table yet.
+
+    # rubocop:disable Rails/NotNullColumn
```
thomasiles, line 7: Rubocop is slightly over protective here so we disable it. 

Rubocop is worrying about the consequences of adding a not null constraint to an existing column where not all rows have a value. This would be awkward because the migration would fail.

We don't have to worry because this is a new table even though the migrations are split here. There won't be any groups without this value and we don't ever want it not to be set. There is no obvious default value as we have taken the decision to generate the external_id in ruby.

app/models/group.rb
```diff
@@ -1,3 +1,14 @@
 class Group < ApplicationRecord
   validates :name, presence: true
+  before_create :set_external_id
```
thomasiles, line 3: This commit takes the decision that the random string which will be used to reference the object from outside the system will be generated in the application code rather than the DB.

It's possible to use postgres functions (and maybe the pg_crypto add on) to generate default values when items are inserted into the DB. Pushing the logic into the DB means that other users of the DB such as rails console or direct access through SQL and AWS would ensure sensible id's are always generated.

The disadvantage is that it's slightly less obvious to application developers how this is happening as we don't often use DB for functions. The same is true for testing and I think it would be harder to make changes in the future.

Instead we generate the values in ruby in the application and enforce the unique constraint at the database level. This makes it clear how the value is generated, gives us flexibility in the future to do other things but stops duplicates entering the db.

app/models/group.rb
thomasiles: This was discussed in tech weekly and with @SamJamCul . We decided that an 8 char long base58 string would be a good place to start and review later on if needed.

db/schema.rb
```diff
@@ -65,6 +73,19 @@
     t.index ["slug"], name: "index_organisations_on_slug", unique: true
   end
 
+  create_table "schema_info", id: false, force: :cascade do |t|
```
jamie-o-wilkinson, line 76: Are these session related changes needed?

db/schema.rb
```diff
@@ -81,6 +102,7 @@
     t.bigint "organisation_id"
     t.boolean "has_access", default: true
     t.string "provider"
+    t.integer "sign_in_count", default: 0, null: false
```
jamie-o-wilkinson, line 105: Looks like this might be a hanger on from the name/organisation requirement?

jamie-o-wilkinson (commented):
A couple of questions but otherwise looks good, thank you! 🎉

db/schema.rb
thomasiles: no! I'll get rid of these and the ones below, they are trapped in there from my local db, good spot!

jamie-o-wilkinson (approved):

PR #929, opened 2024-02-06
Add favicon for use in AUTH0

thomasiles:
The Auth0 login page doesn't have a favicon.

This commit copies the govuk favicon from the node_modules assets dir into the public folder so it has a static URL which auth0 can use.

It's called `public/auth-favicon.ico` rather than `favion.ico` to avoid being used in other contexts than Auth0.

We set the admin favicon through HTML. I think it's probably a good idea to duplicate `favicon.ico` for browsers and applications which check it rather than using the value set in the HTML.

This commit is only about getting an icon in Auth0 so doesn't touch that.

**This is now using the new crown icon.**

```bash
cp node_modules/govuk-frontend/dist/govuk/assets/images/favicon.ico public/ auth-favicon.ico
```

There is a [PR to update the auth0 terraform](https://github.com/alphagov/forms-deploy/pull/568), which should be applied after this is deployed. 

Trello card: https://trello.com/c/FlmddpGh/1356-add-favicon-to-auth0


DavidBiddle (approved):
Nice! 🎉

lfdebrux (approved):

lfdebrux (changes requested):
This should be changed to the Tudor crown now

thomasiles:
It's now got the latest crown, has been rebased to main as has the changes to the terraform.

lfdebrux (approved):

PR #940, opened 2024-02-14
Add payment links task

SamJamCul:
https://trello.com/c/VEe72eLB/1357-add-backend-code-to-support-saving-payment-links-for-a-form

This PR relies on changes to forms-api here https://github.com/alphagov/forms-api/pull/443

Work here is mainly behind the `payment_links` feature flag.

Adds an optional section to the task list, in which a payment link can be set. 

For a payment link to be valid, it needs to be a url which matches this regex `/\Ahttps:\/\/www\.gov\.uk\/payments\/.+\/.+/` - in other words, it goes https://www.gov.uk/payments/something/something

Payment links are not currently used in the runner, so adding them will not affect the previews for forms.

### DIFF
--- a/app/form_objects/forms/payment_link_form.rb
+++ b/app/form_objects/forms/payment_link_form.rb
@@ -1,0 +1,24 @@
+require "uri"
+
+class Forms::PaymentLinkForm < BaseForm
+  attr_accessor :form, :payment_url
+
+  validates :payment_url, url: true, if: -> { payment_url.present? }
+  validate :is_pay_url?, if: -> { payment_url.present? }
+
+  def submit
+    return false if invalid?
+
+    form.payment_url = payment_url
+    form.save!
+  end
+
+  def assign_form_values
+    self.payment_url = form.payment_url
+    self
+  end
+
+  def is_pay_url?
+    errors.add :payment_url, :url unless payment_url.starts_with?("https://www.gov.uk/payments/")
+  end
+end
--- a/app/service/form_task_list_service.rb
+++ b/app/service/form_task_list_service.rb
@@ -18,10 +18,10 @@ class FormTaskListService
 
   def all_sections
     sections = [
-      section_1,
-      section_3,
-      section_4,
-      section_5,
+      create_form_section,
+      email_address_section,
+      privacy_and_contact_details_section,
+      make_form_live_section,
     ]
 
     sections.insert(1, section_2) if FeatureService.enabled?(:payment_links)
@@ -35,24 +35,24 @@ private
     @form.has_live_version ? "edit" : "create"
   end
 
-  def section_1
+  def create_form_section
     {
-      title: I18n.t("forms.task_list_#{create_or_edit}.section_1.title"),
-      rows: section_1_tasks,
+      title: I18n.t("forms.task_list_#{create_or_edit}.create_form_section.title"),
+      rows: create_form_section_tasks,
     }
   end
 
-  def section_1_tasks
+  def create_form_section_tasks
     question_path = if @form.pages.any?
                       form_pages_path(@form.id)
                     else
                       start_new_question_path(@form.id)
                     end
     [
-      { task_name: I18n.t("forms.task_list_#{create_or_edit}.section_1.name"), path: change_form_name_path(@form.id), status: @task_statuses[:name_status] },
-      { task_name: I18n.t("forms.task_list_#{create_or_edit}.section_1.questions"), path: question_path, status: @task_statuses[:pages_status] },
-      { task_name: I18n.t("forms.task_list_#{create_or_edit}.section_1.declaration"), path: declaration_path(@form.id), status: @task_statuses[:declaration_status] },
-      { task_name: I18n.t("forms.task_list_#{create_or_edit}.section_1.what_happens_next"), path: what_happens_next_path(@form.id), status: @task_statuses[:what_happens_next_status] },
+      { task_name: I18n.t("forms.task_list_#{create_or_edit}.create_form_section.name"), path: change_form_name_path(@form.id), status: @task_statuses[:name_status] },
+      { task_name: I18n.t("forms.task_list_#{create_or_edit}.create_form_section.questions"), path: question_path, status: @task_statuses[:pages_status] },
+      { task_name: I18n.t("forms.task_list_#{create_or_edit}.create_form_section.declaration"), path: declaration_path(@form.id), status: @task_statuses[:declaration_status] },
+      { task_name: I18n.t("forms.task_list_#{create_or_edit}.create_form_section.what_happens_next"), path: what_happens_next_path(@form.id), status: @task_statuses[:what_happens_next_status] },
     ]
   end
 
@@ -69,16 +69,16 @@ private
     ]
   end
 
-  def section_3
+  def email_address_section
     section = {
-      title: I18n.t("forms.task_list_#{create_or_edit}.section_3.title"),
+      title: I18n.t("forms.task_list_#{create_or_edit}.email_address_section.title"),
     }
 
     if Pundit.policy(@current_user, @form).can_change_form_submission_email?
-      section[:rows] = section_3_tasks
+      section[:rows] = email_address_section_tasks
     else
       section[:body_text] = I18n.t(
-        "forms.task_list_create.section_3.if_not_permitted.body_text",
+        "forms.task_list_create.email_address_section.if_not_permitted.body_text",
         submission_email: @form.submission_email,
       )
     end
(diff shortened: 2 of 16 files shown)


app/service/form_task_list_service.rb
```diff
@@ -17,12 +17,16 @@ def initialize(form:, current_user:)
...
 
   def all_sections
-    [
+    sections = [
       section_1,
-      section_2,
       section_3,
       section_4,
+      section_5,
     ]
```
lfdebrux, line 25: I'm wondering if maybe we consider renaming the section methods to be based on the task rather than a number 😅 Having to do all this renaming when doing a small reorder seems like a massive pain!

config/locales/en.yml
```diff
@@ -645,6 +653,38 @@ en:
...
+      body_html: |
+        <p>Once someone's submitted their form they'll see a confirmation page showing: </p>
+
+        <ul class="govuk-list govuk-list--bullet">
+          <li>a blue banner saying “You still need to pay”</li>
+          <li>their GOV.UK Forms reference number</li>
+          <li>a green ‘Continue to pay’ button - this will take them to GOV.UK Pay to make their payment</li>
+        </ul>
+
+        <p>The reference number will also be included in a confirmation email,  someone chooses to receive this - along with any 'what happens next' information you've added.</p>
```
lfdebrux, line 672: I don't understand this sentence, and it doesn't seem to match anything in the Mural board linked from the ticket?

app/form_objects/forms/payment_link_form.rb
```diff
@@ -0,0 +1,21 @@
+require "uri"
+
+class Forms::PaymentLinkForm < BaseForm
+  attr_accessor :form, :payment_url
+
+  PAY_URL_REGEX = /\Ahttps:\/\/www\.gov\.uk\/payments\/.+\/.+/
```
lfdebrux, line 6: The `.` in this regex will match `/`, so this regex will match `https://www.gov.uk/payments/A/B/C/D/...`. Is that okay?

```
irb(main):003> Forms::PaymentLinkForm::PAY_URL_REGEX.match("https://www.gov.uk/payments/something/something/something/something")
=> #<MatchData "https://www.gov.uk/payments/something/something/something/something">
irb(main):004> Forms::PaymentLinkForm::PAY_URL_REGEX.match("https://www.gov.uk/payments/something")
```

app/views/forms/payment_link/new.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      <span class="govuk-caption-l"><%= @payment_link_form.form.name %></span>
+      <h1 class="govuk-heading-l">
+        <%= t('payment_link_form.heading') %>
+      </h1>
+
+      <%= t('payment_link_form.body_html') %>
+
+      <h2 class="govuk-heading-m">
+        <%= t('payment_link_form.how_this_will_work.heading') %>
+      </h2>
```
lfdebrux, line 21: What's the reason for not putting the `h2` headers in the translation file?

app/service/form_task_list_service.rb
SamJamCul: Good point, David and I did discuss this and we weren't sure if it would be better to instead save it for a refactor into using the GOV.UK task list component. If we are planning on shifting over to that, then that'd be the best time to name things more appropriately

config/locales/en.yml
SamJamCul: The content from the Mural is currently getting updated based on UR findings, so it's going to be a bit disconnected. I can change it to match what's there currently though

app/views/forms/payment_link/new.html.erb
SamJamCul: Ooh not sure, @DavidBiddle any thoughts?

config/locales/en.yml
DavidBiddle: I think it's missing an 'if' (almost certainly copy-paste error on my part), which might be why it's hard to understand:

```suggestion
        <p>The reference number will also be included in a confirmation email, if someone chooses to receive this - along with any 'what happens next' information you've added.</p>
```

app/views/forms/payment_link/new.html.erb
DavidBiddle: Maybe just personal preference but I generally try to keep as much of the HTML structure as is reasonable in the view file, and have the translation file contains smaller blocks of content. I guess my thinking is mainly that this should help make the locale files easier for non-devs to update.

app/service/form_task_list_service.rb
lfdebrux: Okay, makes sense!

config/locales/en.yml
lfdebrux: Ah yep that's the one!

app/views/forms/payment_link/new.html.erb
lfdebrux: Interesting, that makes sense. Thanks for explaining!

app/form_objects/forms/payment_link_form.rb
DavidBiddle: We might benefit from using the [uri module](https://docs.ruby-lang.org/en/2.1.0/URI.html) here, so that we can validate bits of the URL individually rather than write one monster regex

chao-xian: yeah let's not use regex - Ruby's String.starts_with() AND use URI.parse() https://ruby-doc.org/stdlib-2.5.1/libdoc/uri/rdoc/URI/RFC2396_Parser.html

spec/form_objects/forms/payment_link_form_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        payment_link_form = described_class.new(form:, payment_url: "")
+
+        expect(payment_link_form).to be_valid
+      end
+
+      it "is valid if a PAY url is given" do
+        payment_link_form = described_class.new(form:, payment_url: "https://www.gov.uk/payments/test-org/test-service")
+
+        expect(payment_link_form).to be_valid
+      end
```
chao-xian: We need to include examples that are _not_ valid otherwise we don't know if our validation rule is correct.

spec/form_objects/forms/payment_link_form_spec.rb
```diff
@@ -2,59 +2,30 @@
...
-      end
+    let(:form) do
+      build(:form, :live)
     end
 
-    context "when form is not live" do
-      let(:form) do
-        build(:form)
-      end
+    it "validates the URL" do
```
chao-xian, line 9: This spec declaration is confusing. The example used _isn't_ valid, so it isn't being validated... Maybe something like
``` 
context "is a Pay link"
  it "validates"
...
context "not a Pay link"
  it "fails validation"
```

app/service/form_task_list_service.rb
```diff
@@ -18,10 +18,10 @@ def initialize(form:, current_user:)
 
   def all_sections
     sections = [
-      section_1,
-      section_3,
-      section_4,
-      section_5,
+      create_form_section,
```
chao-xian, line 21: Nice!

chao-xian (changes requested):

app/form_objects/forms/payment_link_form.rb
SamJamCul: Changes made, @lfdebrux if they do manage to paste something in that doesn't match the link that pay gives them but is still a valid url, it'll be hard for us to spot it. Hopefully their testing will reveal that it's not quite right

spec/form_objects/forms/payment_link_form_spec.rb
SamJamCul: fixed, context specified for the tests

app/form_objects/forms/payment_link_form.rb
chao-xian: I also asked Pay if they had a good way of validating and they said what we're doing is legit

lfdebrux: That sounds good enough, nice one @SamJamCul!

DavidBiddle (dismissed):
Looks really good, nice one Sam! 🎉 works nicely on my machine too.

Have raised one tiny comment for something I did 😬

config/locales/en.yml
```diff
@@ -647,6 +655,38 @@ en:
...
+        <p>Before you can take payments using a payment link you’ll need to:</p>
+
+        <ul class="govuk-list govuk-list--bullet">
+          <li>set up a GOV.UK Pay account if you don't have one</li>
+          <li>talk to your local finance team to set up a payment service provider (PSP)</li>
+        </ul>
+
+        <p class="govuk-inset-text">It may be up to several months before you’re ready to take payments. This depends on the PSP arrangements for your department.</p>
+
+        <p><a href="https://www.payments.service.gov.uk/govuk-payment-pages/ target="_blank" rel="noreferrer noopener">Find out more about GOV.UK Pay payment links (opens in a new tab)</a></p>
```
DavidBiddle, line 687: Sorry 😬 
```suggestion
        <p><a href="https://www.payments.service.gov.uk/govuk-payment-pages/" target="_blank" rel="noreferrer noopener">Find out more about GOV.UK Pay payment links (opens in a new tab)</a></p>
```

DavidBiddle (approved):

app/form_objects/forms/payment_link_form.rb
```diff
@@ -3,9 +3,8 @@
 class Forms::PaymentLinkForm < BaseForm
   attr_accessor :form, :payment_url
 
-  PAY_URL_REGEX = /\Ahttps:\/\/www\.gov\.uk\/payments\/.+\/.+/
-
-  validates :payment_url, url: true, format: { with: PAY_URL_REGEX, message: :url }, if: -> { payment_url.present? }
+  validates :payment_url, url: true, if: -> { payment_url.present? }
+  validate :is_pay_url?, if: -> { payment_url.present? }
```
chao-xian, line 7: should it be `validates`? also think this change should be in the earlier commit

DavidBiddle: I believe `validate` is correct here - I think:

- `validate` is for [custom methods](https://guides.rubyonrails.org/active_record_validations.html#custom-methods) (like `is_pay_url?` here), 
- `validates` is for the builting [validation helpers](https://guides.rubyonrails.org/active_record_validations.html#validation-helpers) (like `url: true` on the line above)
- `validates_with` is for [custom validator classes](https://guides.rubyonrails.org/active_record_validations.html#custom-validators)

SamJamCul: I've moved where this is comitted, but I'll leave the `validate` as it's a custom validation

chao-xian (approved):

PR #943, opened 2024-02-15
Add initial HTML content for payment link page

DavidBiddle:

Trello card: 


DavidBiddle:
This is now covered in #940

PR #948, opened 2024-02-16
WIP: Email domains

thomasiles:

Trello card: https://trello.com/c/hlFB6SvQ/1333-spike-how-should-we-get-users-organisation-and-name-when-they-create-an-account

This PR sketches a rough flow for automatically assigning a user's organisation based on their email address.

A new array column is added to organisation to hold an array of email domains owned by that organisation.

When a user first signs in, the domain of their email address is extracted and used to lookup an organisation. If one a domain is found (only single matches are supported) the organisation is assigned to the user. If no organisation is found using the domain, they are presented with a list of organisations to choose from.

The email domains for organisations were converted to JSON and then imported through the Rails command line. They were matched by organisation name directly. Further work would be required to clean up the data and make more efforts at matching our current organisations.

Using this technique, 114 of the 1191 Organisations imported from the API were matched and given domains.

The hacky code used to import the list looked like the following:

```ruby
file = File.open "domains.json"
json_domains = JSON.load file
json_domains.each { |r| o=Organisation.find_by(name: r['name']); o && o.domains = r['domains']; o && o.save }
```

Where `domains.json` is an array of records like the following:
```ruby
{"name"=>"Accountant in Bankruptcy", "type"=>"central", "domains"=>["aib.gov.uk"]}
```

This PR is for illustrations purposes only and should never be merged/deployed.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   before_action :check_maintenance_mode_is_enabled
   before_action :authenticate_and_check_access
   before_action :set_paper_trail_whodunnit
+  before_action :run_first_sign_in_flow
 
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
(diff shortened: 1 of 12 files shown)


thomasiles:
This was a spike and was never going to be merged so I'm closing it.

PR #950, opened 2024-02-19
Move group factory to models folder

lfdebrux:

We conventially put our factory code for models in the folder `spec/factories/models`; let's put the factory for the `Group` model in there.


danworth (dismissed):
All for following convention 👍

DavidBiddle (dismissed):
Seems like a very sensible change!

PR #953, opened 2024-02-20
Add association between groups and organisations

lfdebrux:

This PR adds an association between groups and organisations with the following policies:

- Groups belong to an organisation (mandatory)
- All users can create groups
- New groups belong to the same organisation as the user that created them
- Normal users can view and edit all groups in their organisation
- Super admin users can view and edit all groups

Note that the changes in this PR don't do anything with users who don't have an organisation set (trial users). As we're still figuring out the details on groups and trial users, I left this as a future problem.

This PR also does some refactoring that I thought was useful for implementing the above changes in a cleaner way, if people prefer though I'm happy to split those refactors out into separate PRs.

You can test the changes in this PR locally by running forms-admin and visiting the page `/groups`. I've also added some test groups to the database seed to aid in getting a local test setup up and running quickly.


app/models/group.rb
```diff
@@ -1,4 +1,6 @@
 class Group < ApplicationRecord
+  belongs_to :organisation
```
thomasiles, line 2: Is it worth adding the reverse `has_many :groups` to `Organisation` now or wait until we need it?

app/policies/application_policy.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class ApplicationPolicy
```
thomasiles, line 3: nice!

thomasiles (approved):
Looks really good - the refactors are nice 🏅

app/models/group.rb
lfdebrux: I can't immediately think of a use for the reverse relationship, so I vote leave it out?

PR #955, opened 2024-02-20
Update Gemfile to use file reference for Ruby version

chao-xian:

This means we only have 1 file to update (and means I can update this [script](https://github.com/alphagov/forms-deploy/blob/main/support/update_app_versions.sh#L70) to not bother with updating the Gemfile)


lfdebrux (approved):
Nice improvement!

PR #958, opened 2024-02-21
Fix `User#role_changed_to_editor?`

lfdebrux:

When a user has been created with an editor role the `role_changed_to_editor?` method raised the exception:

```
undefined method `role' for nil:NilClass
```

This was happening because the reification of the last version was `nil`.

I found this when playing around in local development with the `developer` auth provider and signing in as one of the users from the database seed. Currently in production users won't be created with the editor role, but that might change, so I fixed it. Plus it's nice not to have errors in development :sweat_smile:.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -94,7 +94,12 @@ class User < ApplicationRecord
   end
 
   def role_changed_to_editor?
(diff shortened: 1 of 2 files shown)


DavidBiddle (approved):
Good spot and nice fix 🎉

PR #971, opened 2024-02-23
Simplify FormListService

lfdebrux:

Refactor the logic in FormListService to move some of the complexity up into the caller.

This will make it easier to extend FormListService to support other scenarios where we want to list forms, such as when we add groups.


thomasiles (approved):
nice 👍

PR #975, opened 2024-02-26
Add user group association

thomasiles:
# Users can belong to Groups

This is part of the user management work adding the concept of groups to connect users, organisations and forms.
 
- Users can belong to zero or more groups

This PR adds a new model and DB table, `Membership`. This table joins 'User' to `Group`.

There are some business rules we want to enforce with this relationship:
- Users can belong to zero or more groups
- Users can only belong to a single group once
- Users can only belong to a Group if they are in the same Organisation as the Group

The first two are handled by unique constraints in the ruby models and at in the database, through the schema design and indexes.

The third cannot easily be enforced at the database level using our current schema. It's enforced only through the validation on the Membership class.

## Users can only belong to a Group if they are in the same Organisation as the Group

We use this validation on Membership to make sure user and organisation match when changing the membership.

```ruby
  def user_and_group_in_same_organisation
    return if user.nil? || group.nil?
    return if user.organisation == group.organisation

    errors.add :base, "User and group must have the same organisation"
  end
``` 

To enforce this, I think the we would need to add `organisation_id` to `Membership` and use it in composite foreign keys on `User` and `Group`. This would prevent users who are not part of the same organisation as the group being referenced.

### What happens without this?
If a `User` joins a `Group` but later changes their organisation, they will still be part of the group in the DB.

### How do we mitigate it?
I think we have to deal with this in the user and group model to update membership when organisation is changed.  This PR doesn't tackle it yet.

### Why not change the schema to include organisation?

It's a bit of a heavyweight change and commits us to a multi-tenant strategy that I'm not sure we are ready for. It's not completely clear how central organisation will end up being at this moment in the design/development so I think we should hold off.


app/models/group.rb
```diff
@@ -1,6 +1,9 @@
 class Group < ApplicationRecord
   belongs_to :organisation
 
+  has_many :memberships, dependent: :destroy
```
thomasiles, line 4: if a group is deleted, so are its memberships

app/models/membership.rb
```diff
@@ -2,4 +2,15 @@ class Membership < ApplicationRecord
   belongs_to :user
   belongs_to :group
   belongs_to :creator, class_name: "User"
+
+  validate :user_and_group_in_same_organisation
```
lfdebrux, line 6: Can we update the commit message for 80f4cd86a38b0df95b6009a36a04bf8a93dca8bc? It says 'Prevent a user being a member of a group more than once.', but the changes appear to be about a different validation.

lfdebrux (commented):
I've read through the code, looks good at first glance, I want to do some testing though...

db/migrate/20240223134039_create_memberships.rb
```diff
@@ -0,0 +1,13 @@
+class CreateMemberships < ActiveRecord::Migration[7.1]
+  def change
+    create_table :memberships do |t|
+      t.references :user, null: false, foreign_key: true, comment: "The user who is a member of the group"
+      t.references :group, null: false, foreign_key: true, comment: "The group that the user is a member of"
+      t.references :creator, null: false, foreign_key: { to_table: :users }, comment: "The user who created the membership"
```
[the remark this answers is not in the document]
lfdebrux: Sorry to bikeshed, thinking about it further, `creator` feels weird to me, what about `added_by` as a name for this attribute? It feels more natural to say one user added another, and I've seen that particular term used in other Rails codebases [[1](https://github.com/bullet-train-co/bullet_train-core/blob/9227d144096d0990d5abb4a90be994c2faeeb77f/bullet_train/db/migrate/20200219013834_add_added_by_to_membership.rb#L3)].

db/migrate/20240223134039_create_memberships.rb
```diff
@@ -0,0 +1,13 @@
+class CreateMemberships < ActiveRecord::Migration[7.1]
+  def change
+    create_table :memberships do |t|
+      t.references :user, null: false, foreign_key: true, comment: "The user who is a member of the group"
+      t.references :group, null: false, foreign_key: true, comment: "The group that the user is a member of"
```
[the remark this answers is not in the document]
thomasiles: I think we probably can get rid of them 👍

thomasiles:
Thanks for reviewing, I've updated it with the suggestions above 👍

app/models/membership.rb
```diff
@@ -0,0 +1,17 @@
...
+  belongs_to :user
+  belongs_to :group
+  belongs_to :added_by, class_name: "User"
+
+  validate :user_and_group_in_same_organisation
+  validates :user, uniqueness: { scope: :group }
+
+private
+
+  def user_and_group_in_same_organisation
```
chao-xian, line 11: The commit message says the code was from the Rails generator, but as this is custom it should probably go into a separate commit with the appropriate spec tests.

spec/factories/models/memberships.rb
chao-xian, line 1: The commit title says `...and move specs` but I wouldn't call this a spec...?

chao-xian (commented):
Just some commit tidy ups from me but looking good so far.

lfdebrux:
@thomasiles thanks for writing up the bits about organisation membership in more detail, that's very helpful.

Thinking about what to do if a user's organisation changes, it might be a bit drastic, but what if we just say for know that when a user moves from one organisation to another, they lose all of their group memberships? I think that would make some of the questions you had moot, what do you think?

It should be straightforward to add a step to the `on_user_update` method to achieve this:

https://github.com/alphagov/forms-admin/blob/7d6d76d28ff9edefaf23c38cd2fa3228debb3719/app/service/user_update_service.rb#L15-L18

thomasiles:
> @thomasiles thanks for writing up the bits about organisation membership in more detail, that's very helpful.
> 
> Thinking about what to do if a user's organisation changes, it might be a bit drastic, but what if we just say for know that when a user moves from one organisation to another, they lose all of their group memberships? I think that would make some of the questions you had moot, what do you think?
> 
> It should be straightforward to add a step to the `on_user_update` method to achieve this:
> 
> https://github.com/alphagov/forms-admin/blob/7d6d76d28ff9edefaf23c38cd2fa3228debb3719/app/service/user_update_service.rb#L15-L18

yes - i think this is the only thing which makes sense. I'd rather not make this commit any longer though so I think it's best to do it in another one.

lfdebrux (approved):

PR #978, opened 2024-02-27
Remove mismatched memberships

thomasiles:
### Remove old organisation groups when a member changes organisation

When a User changes organisation they should no longer be a member of the old organisation's groups.

### DIFF
--- a/spec/models/membership_spec.rb
+++ b/spec/models/membership_spec.rb
@@ -45,15 +45,44 @@ RSpec.describe Membership, type: :model do
     expect { membership.save(validate: false) }.to raise_error(ActiveRecord::RecordNotUnique)
   end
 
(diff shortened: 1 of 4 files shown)


spec/models/membership_spec.rb
```diff
@@ -44,4 +44,16 @@
     membership = build(:membership, user:, group:)
     expect { membership.save(validate: false) }.to raise_error(ActiveRecord::RecordNotUnique)
   end
+
+  it "removes mismatched memberships" do
```
lfdebrux, line 48: Can we have a test for `destroy_invalid_organisation_memberships` that checks it doesn't affect membership of other users?

DavidBiddle (approved):
Nice work! 🎉

PR #988, opened 2024-03-05
Refactor specs to reduce repetition of expected API request headers

lfdebrux:

It would be nice if we didn't have to rewrite/copy and paste the expected API request headers every time we want to write some new specs.

This PR makes it so.


thomasiles (approved):
Nice! I like the way you structured the commits for this 🏅

DavidBiddle (approved):
Love this!

PR #994, opened 2024-03-08
Restrict access to Groups to members

thomasiles:
### Restrict access to groups

Trello card: https://trello.com/c/OOy3MZps/1417-restrict-who-can-view-groups-to-group-members

Restrict access to list, view, modify and delete groups to only users who are members of that group.

Except for super_admins, who are able to access all groups.

This commit changes the group policy. It adds tests for the existing policy and the new scope and restrictions.

The policy scope delegates to a scope on the model because that seems a better place to implement and test it.

To support this work the group controller has been modified so that when a user creates a group, they are added to it.

The controller spec has also been updated because users will no longer be able to see all groups within their organisation, only those they are a member of.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -29,6 +29,7 @@ class GroupsController < ApplicationController
   def create
     @group = Group.new(group_params)
     @group.organisation = current_user.organisation
+    @group.memberships.build(user: current_user, added_by: current_user)
 
     authorize @group
 
--- a/app/policies/group_policy.rb
+++ b/app/policies/group_policy.rb
@@ -8,6 +8,7 @@ class GroupPolicy < ApplicationPolicy
     user.super_admin? || user.organisation_id == record.organisation_id
   end
   alias_method :edit?, :show?
+  alias_method :index?, :show?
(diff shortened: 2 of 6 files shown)


app/controllers/groups_controller.rb
```diff
@@ -29,6 +29,7 @@ def edit
   def create
     @group = Group.new(group_params)
     @group.organisation = current_user.organisation
+    @group.memberships.build(user: current_user, added_by: current_user)
```
thomasiles, line 32: we add ourselves to the group we are creating, otherwise we won't be part of it and have any access

app/policies/group_policy.rb
```diff
@@ -16,7 +17,7 @@ def resolve
       if user.super_admin?
         scope.all
       else
-        scope.where(organisation_id: user.organisation_id)
+        scope.for_user(user)
```
thomasiles, line 20: Super_admins can access every group.

Everyone else can only access the groups they belong to.

jamie-o-wilkinson (approved):
Looks good to me 👍 tests are tidy, especially for the scopes!

PR #995, opened 2024-03-08
Add group status

thomasiles:
### Add group status

Trello card: https://trello.com/c/5uZvQ5w1/1410-add-group-status-active-trial

See the trello card for details and design.

This PR adds a status to Groups which will be used to decide if forms within the group can be made live or not.

An enum is added to the Groups DB and the views updated to show the status of groups.

image

image


jamie-o-wilkinson (approved):
Looks good to me so far. Thanks for splitting up the commit history, that made it much easier to look through.

PR #996, opened 2024-03-11
Deny access for super-admins authenticated without Auth0 via Google workspace

jamie-o-wilkinson:

We want to ensure that all super-admins have logged in using MFA. We do this by checking that the Auth0 connection used by the super-admin user was the Google workspace connection. 

This is achieved by storing the value of the `auth0_connection_strategy` custom claim from the ID token from Auth0 in the warden session. We then can check that value in the existing `authenticate_and_check_access` before action.

For now, if this check fails we just show the access denied screen . We may change this is to something more useful in the future.

Trello card: https://trello.com/c/UdtU9PZC

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -29,6 +29,10 @@ class ApplicationController < ActionController::Base
     render "errors/forbidden", status: :forbidden, formats: :html
   end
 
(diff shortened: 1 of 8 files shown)


thomasiles (approved):
very clear 🏅

PR #998, opened 2024-03-12
Rename resource param to match Rails convention

lfdebrux:

Trello card: https://trello.com/c/16aEvxcI/1419-add-start-button-to-create-forms-in-group 

This will make having nested resource easier, and I want to nest forms under groups in a future PR.


thomasiles (commented):
👍

thomasiles (approved):

PR #999, opened 2024-03-12
Introduce way of synchronizing MailChimp lists with users

AP-Hunt:

Trello card: https://trello.com/c/iQOMKY1h/364-add-rake-task-to-automatically-add-new-users-to-mailchimp-mailing-lists

We have a database table full of user information, and we have some MailChimp lists that we'd like to contain all the same email addresses.

We're happy for the synchronization to happen out-of-band with regular user management activities, so we've built it as a Rake task that will be run on a schedule. The scheduling will be handled at the infrastructure level, in the `forms-deploy` repository.

### DIFF
--- a/config/settings/production.yml
+++ b/config/settings/production.yml
@@ -1,2 +1,8 @@
 # Use real authentication
 auth_provider: gds_sso
+
+mailchimp:
+  api_prefix: "us12"
+  lists:
+    - c9cc8511c6
+    - ef5e72c003
--- a/lib/tasks/mailchimp.rake
+++ b/lib/tasks/mailchimp.rake
@@ -1,0 +1,71 @@
+require "digest"
+require "MailchimpMarketing"
+
+namespace :mailchimp do
+  desc "Synchronise Mailchimp audiences with the users in the database"
+  task synchronize_audiences: :environment do
+    mailchimp_api_key = ENV["MAILCHIMP_API_KEY"]
+    mailchimp_server_prefix = ENV["MAILCHIMP_SERVER_PREFIX"]
+    mailchimp_lists = ENV["MAILCHIMP_LISTS"]
+
+    puts "Mailchimp server prefix: #{mailchimp_server_prefix}"
+    puts "Mailchimp lists: #{mailchimp_lists}"
+
+    mailchimp = MailchimpMarketing::Client.new
+    mailchimp.set_config({
+      api_key: mailchimp_api_key,
+      server: mailchimp_server_prefix,
+    })
+
+    list_ids = mailchimp_lists
+                  .split(",")
+                  .map(&:strip)
+
+    db_email_addresses = User.pluck(:email).to_set
+
+    puts "There are #{list_ids.length} lists to synchronize"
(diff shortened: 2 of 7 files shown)


lib/tasks/mailchimp.rake
```diff
@@ -0,0 +1,71 @@
+require "digest"
+require "MailchimpMarketing"
+
+namespace :mailchimp do
+  desc "Synchronise Mailchimp audiences with the users in the database"
+  task synchronize_audiences: :environment do
+    mailchimp_api_key = ENV["MAILCHIMP_API_KEY"]
+    mailchimp_server_prefix = ENV["MAILCHIMP_SERVER_PREFIX"]
+    mailchimp_lists = ENV["MAILCHIMP_LISTS"]
```
lfdebrux, line 9: Could we get these from the normal [settings](https://github.com/alphagov/forms-admin/blob/main/config/settings.yml) [mechanism](https://github.com/rubyconfig/config) rather than directly from the environment?

lfdebrux: No worries :) It's a thing we've added to our apps, it's not part of Rails out-of-the-box.

config/settings/production.yml
```diff
@@ -1,2 +1,6 @@
 # Use real authentication
 auth_provider: gds_sso
+
+mailchimp:
+  api_prefix: "us12"
+  lists: "c9cc8511c6,ef5e72c003"
```
lfdebrux, line 6: If we're defining the configuration here and not in the environment we can use a YAML array instead of having to split the string

AP-Hunt: Good point!

lfdebrux:
In Mailchimp we currently for each contact as well as email address we have first name, last name and organisation. Can this script also handle that?

AP-Hunt:
> In Mailchimp we currently for each contact as well as email address we have first name, last name and organisation. Can this script also handle that?

[The `users` database table](https://github.com/alphagov/forms-admin/blob/main/db/schema.rb#L98) and resulting model in Ruby records the user's name as a single field, but MailChimp has it as a first and last name fields. We could try to be clever with splitting on spaces, but I think we'd be asking to get it wrong a lot. 

So I think we _can_ do it, but I'm not convinced we _should_.

lfdebrux:
> > In Mailchimp we currently for each contact as well as email address we have first name, last name and organisation. Can this script also handle that?
> 
> [The `users` database table](https://github.com/alphagov/forms-admin/blob/main/db/schema.rb#L98) and resulting model in Ruby records the user's name as a single field, but MailChimp has it as first and last name fields. We could try to be clever with splitting on spaces, but I think we'd be asking to get it wrong a lot.
> 
> So I think we _can_ do it, but I'm not convinced we _should_.

That's a good point, I definitely agree we should not be doing the splitting. We might be able to configure Mailchimp to only have one name field.

But actually looking into our current Mailchimp audience in more detail it seems that for platform update emails I think we don't care too much about that extra information, so what you've got now should work.

lfdebrux (approved):

PR #1000, opened 2024-03-13
Make tables take up full width of page

lfdebrux:

We want more horizontal space for our tables of forms/groups, so this commit removes the two-thirds width grid containers.

This brings these pages into line with the current designs in Figma/our prototypes [[1]].

[1]: https://www.figma.com/file/D2DtaS68qRvVZgtxaBQNjd/User-management---moving-to-a-'group'-model-for-public-beta?type=design&node-id=124-7025&mode=design&t=0jENmoJ9JAJG8Gaa-0

<details>
<summary><h3>Screenshots</h3></summary>

##### Before

Screenshot of groups page /groups before this change

##### After

Screenshot of groups page /groups after this change

##### Before

Screenshot of group forms page /groups/:group_id before this change

##### After

Screenshot of group forms page /groups/:group_id after this change

</details>

### DIFF
--- a/app/views/groups/index.html.erb
+++ b/app/views/groups/index.html.erb
@@ -4,21 +4,17 @@
   <%= notice %>
 <% end%>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-l">
-      <%= t("page_titles.group_index") %>
-    </h1>
+<h1 class="govuk-heading-l">
+  <%= t("page_titles.group_index") %>
+</h1>
 
-    <%= render GroupListComponent::View.new(groups: @active_groups, title: t('groups.index.active_title'), empty_message: "You're not in any active groups.") %>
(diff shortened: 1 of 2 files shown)


hazalarpalikli:
Hi Laurence, there are a few things i have spotted:

On the group forms page (I am assuming these screens are for group admin) 
1- Notification banner is missing a link that says "Find out how to upgrade this group so you can make forms live"
2- On the same page we are also missing a second link underneath "Change the name of this group" that should say "Edit members of this group" 
3- We also need to have a CTA button underneath "edit members of this group" and it should say "Create a form:
4- We need to remove "delete group" button as we said we'll pause deleting groups for now.

I have tagged you on Figma for the design

lfdebrux:
> Hi Laurence, there are a few things i have spotted:

Thanks for the review @hazalarpalikli! Sorry I should have been clear when I tagged you, this PR is just about making the tables full width, I know there are other bits that are missing/need changing on the pages, but some of those changes can’t be made just yet (because the code to make the things happen behind the scenes is still being worked on).

But does the table look like you were expecting at full width?

thomasiles (approved):

PR #1001, opened 2024-03-13
Change GroupPolicy to only show members

thomasiles:
Change the GroupPolicy to enforce that only members of groups can show, modify or remove the group.

Trello card: 


lfdebrux (approved):
👍

PR #1002, opened 2024-03-13
Fix super admin user for developers

lfdebrux:

In commit bcac184 we restricted use of the super admin role to users signed in with Google SSO. However, running forms-admin locally generally Auth0 is not used. This commit adds a quick bypass for this scenario, so developers can still use super admin roles locally.


jamie-o-wilkinson (approved):
Silly thing to miss, was testing locally using Auth0 🤦‍♂️ 

Thank you!

PR #1005, opened 2024-03-14
Prevent specs from calling forms-api

lfdebrux:

When writing specs locally, it is easy to accidentally forget to mock ActiveResource requests to forms-api if that app is running at the same time.

This commit adds a line to our spec helper file to disable all network calls via ActiveResource. Now if we forget to mock something, an error will be raised even if forms-api is running.


thomasiles (approved):
good idea

PR #1006, opened 2024-03-14
Refactor form name form object

lfdebrux:

This PR does some refactoring of the `Forms::ChangeNameForm` class and associated views; this is partly to make re-use easier (in a future PR we want to use this form object in a third place), and partly because the way it is named kind of bothers me...


thomasiles (approved):
nice, interesting view tests

lfdebrux:
> nice, interesting view tests

Thanks, to be honest I cribbed them from the ones generated by rspec-rails 😅 

https://github.com/rspec/rspec-rails/blob/main/lib/generators/rspec/scaffold/templates/edit_spec.rb

PR #1008, opened 2024-03-14
Change Form access policy for forms in groups

thomasiles:
### Update form policy to allow access to group members

Trello card: https://trello.com/c/16aEvxcI/1419-add-start-button-to-create-forms-in-group

Change the Form policy. If the form is part of a group, allow access if the user is a member of the group, otherwise deny access, unless the user is a super_admin.


lfdebrux (approved):
🙌

PR #1013, opened 2024-03-18
Add form creation within groups

jamie-o-wilkinson:

We want members of a group to be able to create forms in that group. This PR adds the necessary controller, routing, policy, and view changes to add a button to create a form within a group.

We also took the approach in tests that forms within groups will be the new default, and have updated tests we were changing anyway to adopt this approach. Hopefully this means refactoring to remove the legacy "form without a group" case is easier in future. This also highlighted the need for feature testing of creation of groups, which will follow in a future PR.

Trello card: https://trello.com/c/16aEvxcI


thomasiles (approved):
nice chopping up!

PR #1014, opened 2024-03-18
Add feature test for creating Group

thomasiles:
## Add feature tests for Groups

To ensure the Group feature works as expected, we need a feature test.

This commit adds feature tests for creating a group.

It navigates directly to the goup page because it's not linked into the main application yet.

It also uses the back button to check that the group is listed in the user's group page. I added this because it's an issue I hit when adding groups and doesn't take much longer than just creating the group.


jamie-o-wilkinson (approved):
👏

PR #1021, opened 2024-03-20
Use email address as link text on list of users

DavidBiddle:

Trello card: https://trello.com/c/UOgSrjeL/1439-users-list-page-multiple-links-have-identical-link-text-and-context-no-name-set

Previously on the users page, we linked from the user's name to the edit user page. This failed guideline 2.4.4 of the Web Content Accessibility Guidelines, as it meant that when multiple users have the same name (either because their actual names are identical, or because they didn't have names set), there would be multiple links with the same link context.

This PR makes the user's email address the link instead of the user's name. This prevents the issue, as email addresses are unique in our service.

#### Before
Screenshot of a table with headings for name, email address, organisation, role and access. The entries for the name column are all hyperlinks, the other columns are plain text. Several of the entries in the name column have the identical link text 'No name set'

#### After
Screenshot of a table with headings for name, email address, organisation, role and access. The entries for the email address column are all hyperlinks, the other columns are plain text. Each entry in the email address column has unique link text.

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -18,10 +18,10 @@
     <%= table.with_body do |body|
       users.each do |user|
         body.with_row do |row|
+          row.with_cell( text: user.name || t("users.index.name_blank"))
           row.with_cell do
-            govuk_link_to(user.name || t("users.index.name_blank"), edit_user_path(user))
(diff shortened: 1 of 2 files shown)


chao-xian (approved):
Thanks David!

PR #1022, opened 2024-03-20
Add closed attribute to organisations

lfdebrux:

Trello card: https://trello.com/c/sLRD2UNh/1434-update-the-database-of-organisations-so-users-can-select-from-a-shorter-list 

We want to make it easier for users to choose the right organisation. Our list of organisations includes lots of departments and whatnots that no longer exist for various reasons; this information is recorded in the GOV.UK organisation API, but we previously we haven't been recording that information in our own database.

This PR adds a `closed` column to the organisation table that indicates whether an organisation is no longer extant.

I've called this status field `closed` even though this is a negative because it most closely matches what's in the [GOV.UK Signon codebase][1], and I thought it would be useful for our codebase to hew closely to theirs; if people find the default double negative too heinous though I'll change it!

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -6,5 +6,6 @@ class Organisation < ApplicationRecord
 
   has_many :mou_signatures
 
+  scope :not_closed, -> { where(closed: false) }
   scope :with_users, -> { joins(:users).distinct.order(:name) }
 end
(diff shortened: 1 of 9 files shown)


thomasiles (dismissed):
good work! lgtm and works well locally.

I think closed is a good name 👍

thomasiles (approved):

PR #1023, opened 2024-03-20
Add creator to group

jamie-o-wilkinson:

We want to be able to see who created each group, so that org and super admins can contact the correct person for initial group admin tasks.

This is done via a creator `belongs_to` association between the `Group` and `User` models. We then show the creator's name in the "Created by" column on the groups page.

Trello card: https://trello.com/c/KEYTEmgj

### DIFF
--- a/app/components/group_list_component/view.html.erb
+++ b/app/components/group_list_component/view.html.erb
@@ -21,7 +21,7 @@
       @groups.each do |group|
         body.with_row do |row|
           row.with_cell { govuk_link_to(group.name, group) }
-          row.with_cell(numeric: true) { t('groups.group_list.created_by_unknown') }
+          row.with_cell(numeric: true) { group.creator&.name.presence || t('groups.group_list.created_by_unknown') }
         end
       end
     end %>
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -1,6 +1,8 @@
 class Group < ApplicationRecord
   belongs_to :organisation
 
+  belongs_to :creator, class_name: "User", optional: true
+
   has_many :memberships, dependent: :destroy
   has_many :users, through: :memberships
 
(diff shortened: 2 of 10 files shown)


app/models/group.rb
```diff
@@ -17,6 +17,10 @@ def to_param
     external_id
   end
 
+  def creator
+    User.find_by(id: creator_id)
+  end
+
```
lfdebrux, line 23: Why not make this a proper association?

lfdebrux:
It might be worth updating the group factory to add a creator to groups in our specs. Also, it would be good to have creators for the groups we create in our database seed.

app/components/group_list_component/view.html.erb
```diff
@@ -21,7 +21,7 @@
       @groups.each do |group|
         body.with_row do |row|
           row.with_cell { govuk_link_to(group.name, group) }
-          row.with_cell(numeric: true) { t('groups.group_list.created_by_unknown') }
+          row.with_cell(numeric: true) { group.creator.present? ? group.creator&.name : t('groups.group_list.created_by_unknown') }
```
DavidBiddle, line 24: Not exactly equivalent but I think you could use [presence](https://apidock.com/rails/Object/presence) here to avoid the ternary operator:

```suggestion
          row.with_cell(numeric: true) { group.creator&.name.presence || t('groups.group_list.created_by_unknown') }
```

app/models/group.rb
jamie-o-wilkinson: I think I was distracted by wondering if we should be doing this via the `Membership` model...there is no good reason not to, so I've changed it to a `belongs_to` association.

thomasiles (approved):
LGTM 🏅

lfdebrux (approved):
Really nice and clean, good work! 🙌

PR #1030, opened 2024-03-21
Do not allow net requests in specs

lfdebrux:

We don't want our tests to make external HTTP requests.

We allow localhost so that Selenium/Webdriver messages still get through.


danworth (approved):
Oh nice catch 🏅

PR #1031, opened 2024-03-21
Admin: Ask user for name and org

thomasiles:
### Admin: Ask user for name and org

Trello card: https://trello.com/c/Apx9nwqS/1412-admin-ask-user-for-name-and-org-on-sign-in

This PR adds two new forms for capturing user organisation and name.

These will be used to ask the user for these values when we don't already have them on record. This PR only adds the forms, it does not show the user the screens unless they navigate directly to the URLs.

The forms are available to access for users who do not have name/org set. It is possible for them to directly access the forms and use them to update their own name and organisation. This isn't an issue, they are just doing it early, but worth pointing out.

### Organisation screen
image

### Name screen
image

### DIFF
--- a/app/controllers/account/names_controller.rb
+++ b/app/controllers/account/names_controller.rb
@@ -12,7 +12,7 @@ module Account
       if @name_form.submit
         redirect_to root_path
       else
-        render :edit
+        render :edit, status: :unprocessable_entity
       end
     end
 
--- a/app/controllers/account/names_controller.rb
+++ b/app/controllers/account/names_controller.rb
@@ -1,0 +1,29 @@
+module Account
+  class NamesController < ApplicationController
+    before_action :redirect_if_name_exists
+
+    def edit
+      @name_form = NameForm.new(user: current_user).assign_form_values
+    end
+
+    def update
+      @name_form = NameForm.new(account_name_form_params(current_user))
+
+      if @name_form.submit
+        redirect_to root_path
+      else
+        render :edit
+      end
+    end
+
+    def account_name_form_params(user)
+      params.require(:account_name_form).permit(:name).merge(user:)
+    end
+
+  private
+
+    def redirect_if_name_exists
+      redirect_to root_path if current_user.name.present?
+    end
+  end
+end
--- a/app/controllers/account/organisations_controller.rb
+++ b/app/controllers/account/organisations_controller.rb
@@ -1,0 +1,29 @@
+module Account
+  class OrganisationsController < ApplicationController
+    before_action :redirect_if_organisation_exists
+
+    def edit
+      @organisation_form = OrganisationForm.new(user: current_user).assign_form_values
+    end
+
+    def update
+      @organisation_form = OrganisationForm.new(account_organisation_form_params(current_user))
+
+      if @organisation_form.submit
+        redirect_to root_path
+      else
+        render :edit
+      end
+    end
+
+    def account_organisation_form_params(user)
+      params.require(:account_organisation_form).permit(:organisation_id).merge(user:)
+    end
+
+  private
+
+    def redirect_if_organisation_exists
+      redirect_to root_path if current_user.organisation.present?
+    end
+  end
+end
--- a/app/controllers/account/organisations_controller.rb
+++ b/app/controllers/account/organisations_controller.rb
@@ -12,7 +12,7 @@ module Account
       if @organisation_form.submit
         redirect_to root_path
       else
-        render :edit
+        render :edit, status: :unprocessable_entity
       end
     end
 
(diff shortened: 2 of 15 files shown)


app/controllers/account/organisations_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    before_action :redirect_if_organisation_exists
+
+    def edit
+      @organisation_form = OrganisationForm.new(user: current_user).assign_form_values
+    end
+
+    def update
+      @organisation_form = OrganisationForm.new(account_organisation_form_params(current_user))
+
+      if @organisation_form.submit
```
lfdebrux, line 12: It would be good to log somewhere what organisation the user selects, so we can do some monitoring in Splunk

app/form_objects/account/organisation_form.rb
```diff
@@ -0,0 +1,17 @@
+class Account::OrganisationForm < BaseForm
```
chao-xian, line 1: I think this could be a prior commit with the corresponding spec.

app/form_objects/account/name_form.rb
```diff
@@ -0,0 +1,17 @@
+class Account::NameForm < BaseForm
```
chao-xian, line 1: Again I think it's worth splitting the components into their own commits.

chao-xian (commented):
Think the commits could do with breaking down but lgtm so far. Also a typo in second commit message `orgnanisation` 

Great to see this work coming along so nicely!

app/controllers/account/organisations_controller.rb
thomasiles: The value saved in the DB? Wouldn't we use that for monitoring? I think the only value which can be submitted would be a valid org so this should always be set, i think.

app/form_objects/account/organisation_form.rb
thomasiles: something to consider for other features?

I'd rather get this merged and move on to other work than break the existing commits up.

app/views/account/organisations/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+        <%= f.govuk_error_summary %>
+      <% end %>
+
+      <%= render DfE::Autocomplete::View.new(
+        f,
+        attribute_name: :organisation_id,
+        form_field: f.govuk_collection_select(:organisation_id,
+                                              Organisation.order(:slug),
+                                              :id,
+                                              :name,
```
DavidBiddle, line 15: You've probably seen #1033 but flagging it just in case - I think it would make sense to update this to allow searching by abbreviation once that's merged. I'm happy for that to be in a subsequent PR though, if you want to get this one merged quicker.

app/controllers/account/names_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      @name_form = NameForm.new(user: current_user).assign_form_values
+    end
+
+    def update
+      @name_form = NameForm.new(account_name_form_params(current_user))
+
+      if @name_form.submit
+        redirect_to root_path
+      else
+        render :edit
```
DavidBiddle, line 15: I often forget to do this, but I think we usually try to return 422 for validation errors:
```suggestion
        render :edit, status: :unprocessable_entity
```

app/controllers/account/organisations_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        redirect_to root_path
+      else
+        render :edit
```
DavidBiddle, line 15: Same as above:
```suggestion
        render :edit, status: :unprocessable_entity
```

config/locales/en.yml
```diff
@@ -1,8 +1,21 @@
 ---
 en:
+  account:
+    organisation_details_summary: If your organisation is not listed
+    organisation_text_html: |
+      <p>If you're from a central government organisation that publishes content on the GOV.UK website but your organisation is not listed, please %{contact_link}.</p>
```
DavidBiddle, line 6: Curly quote nitpick:
```suggestion
      <p>If you’re from a central government organisation that publishes content on the GOV.UK website but your organisation is not listed, please %{contact_link}.</p>
```

app/views/account/organisations/edit.html.erb
thomasiles: thanks, will do

DavidBiddle (dismissed):
Code looks good and it works for me locally when I navigate directly to the new routes 🎉

app/controllers/account/organisations_controller.rb
thomasiles: added

DavidBiddle (approved):

PR #1032, opened 2024-03-21
Add ability to view archived form

SamJamCul:

https://trello.com/c/UyRNXHGh/1330-implement-archive-a-form-journey

This adds a controller action for showing an archived form. It also refactors the live form show page to make use of the same partial. 

You will need to manually set a form state to `archived` if you want to be able to view this page locally. This can be done in the api with a form in the live state having `.archive_live_form!` called on it. Then, viewing that form in forms-admin will render the show archived form page. 

_Find an archived form_
image

_Show an archived form_
image


spec/controller/application_controller_spec.rb
```diff
@@ -78,28 +81,56 @@ def index
...
       expect(Form).to have_received(:find).exactly(1).times
     end
   end
+
+  describe "#current_live_form" do
+    it "returns the current live form" do
+      params = ActionController::Parameters.new(form_id: id)
+      allow(Form).to receive(:find_live).with(id).and_return(form)
+      allow(controller).to receive(:params).and_return(params)
+      expect(controller.current_live_form).to eq form
```
chao-xian, line 106: Very minor and something that was already in this file, but there should be a newline before the expect for readability. That can be added in a separate commit.

app/controllers/forms/archived_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Forms
+  class ArchivedController < ApplicationController
+    after_action :verify_authorized
+    def show_form
```
chao-xian, line 4: Missing a newline before the def

app/service/form_list_service.rb
```diff
@@ -59,11 +59,10 @@ def rows
   end
 
   def form_name_link(form)
-    if form.has_live_version
-      govuk_link_to(form.name, live_form_path(form.id))
-    else
-      govuk_link_to(form.name, form_path(form.id))
-    end
+    return govuk_link_to(form.name, live_form_path(form.id)) if form.has_live_version
```
chao-xian, line 62: Nice use of guard clause!

chao-xian (commented):
At first glance, just some style things but looking good atm. Need to run it locally as well before I approve

app/controllers/forms/archived_controller.rb
DavidBiddle: Is this something we can/should make rubocop do?

app/controllers/application_controller.rb
```diff
@@ -103,6 +103,10 @@ def current_live_form
     @current_live_form ||= Form.find_live(params[:form_id])
   end
 
+  def current_archived_form
```
chao-xian, line 106: Move this into the ArchivedController (the others in here for example `current_live_form` if that can move, put that into the 2 places it's used).

app/controllers/forms/archived_controller.rb
chao-xian: You'd think so, yes.

spec/controller/application_controller_spec.rb
SamJamCul: They've now been separated properly

app/controllers/forms/archived_controller.rb
SamJamCul: Added the blank line, bit annoying that rubocop isn't catching this

app/controllers/application_controller.rb
SamJamCul: This is now moved as part of adding the archived controller

chao-xian (approved):

app/views/forms/_made_live_form.html.erb
```diff
@@ -11,59 +11,67 @@
...
+      <%= render FormUrlComponent::View.new(link_to_runner(Settings.forms_runner.url, form.id, form.form_slug, mode: :live))%>
+    <% else %>
+      <h3 class="govuk-heading-m"><%= t('made_live_form.previous_form_url') %></h2>
+      <p>
+        <%= link_to_runner(Settings.forms_runner.url, form.id, form.form_slug, mode: :live) %>
+      </p>
+    <% end %>
+
+    <h3 class="govuk-heading-m"><%= t('made_live_form.questions') %></h3>
+    <p><%= govuk_link_to t('made_live_form.questions_link', count: form.pages.count), questions_path %></p>
```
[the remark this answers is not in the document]
stephencdaly: Yes we haven't implemented this route yet, this will come in the next PR as we didn't want this one to get too big.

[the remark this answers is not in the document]
DavidBiddle: Figured you probably had this in hand!

DavidBiddle (approved):
Code looks good and works as expected on my local machine 🎉

PR #1036, opened 2024-03-25
Clean up output of tests

lfdebrux:

Currently our test output includes log lines from some of the scripts it is running, this PR quiets those so only the output from RSpec is present.


DavidBiddle (approved):
Nice change!

PR #1037, opened 2024-03-25
Implement route to show questions for an archived form

stephencdaly:

Trello card: https://trello.com/c/UyRNXHGh/1330-implement-archive-a-form-journey

Implement the `/forms/:form_id/archived/pages` route which shows the questions for an archived form. This page is identical to the page for a live form, but with different back links and status tag. So have used a partial to share the view code.

You will need to manually set a form state to archived if you want to be able to view this page locally. This can be done in the api with a form in the live state having .archive_live_form! called on it. 

Then, viewing that form in forms-admin will render the show archived form page. 

Click on the "View your form's X questions" to get to the route implemented by this PR

Screenshot 2024-03-25 at 12 58 30

Screenshot 2024-03-25 at 12 59 33


DavidBiddle (approved):
I think this looks really good! Works on my machine too 🎉

lfdebrux:
Nice change! One small comment, we usually do rebases to avoid the having "Merge branch 'main'" commits in branches

lfdebrux (approved):

PR #1038, opened 2024-03-25
DO NOT MERGE: Enable ask user name org

thomasiles:
### Redirect users to set name and organisation when we don't have them

Trello card: https://trello.com/c/yzbuocwP/1445-launch-ask-for-name-organisation-feature

This PR shouldn't be merged until an email has been sent to users telling them to expect to enter name and organisation.

It enables the feature using a before_action on the application controller to redirect all users on every request if name/org is not present.


thomasiles:
This has been replaced by this PR https://github.com/alphagov/forms-admin/pull/1051

PR #1048, opened 2024-03-27
Add role to group membership

jamie-o-wilkinson:

We want to use role based access control for actions that can be taken at the group level, e.g. manage users, make forms live. But first, we need the roles!

This PR adds the editor and group admin roles to group memberships, so that each user can have a role for each group they are a member of. We also give the creator of a group the group admin role. Restricting what each role can do will come later.

Trello card: https://trello.com/c/YuTJjp3p

### DIFF
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -4,6 +4,7 @@ class Membership < ApplicationRecord
   belongs_to :added_by, class_name: "User"
 
   enum :role, {
+    group_admin: "group_admin",
     editor: "editor",
   }
 
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -3,8 +3,13 @@ class Membership < ApplicationRecord
   belongs_to :group
   belongs_to :added_by, class_name: "User"
 
+  enum :role, {
+    editor: "editor",
+  }
+
   validate :user_and_group_in_same_organisation
   validates :user, uniqueness: { scope: :group }
+  validates :role, presence: true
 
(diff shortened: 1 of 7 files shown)


app/models/membership.rb
```diff
@@ -3,8 +3,14 @@ class Membership < ApplicationRecord
   belongs_to :group
   belongs_to :added_by, class_name: "User"
 
+  enum :role, {
+    group_admin: "group_admin",
```
jamie-o-wilkinson, line 7: I initially just called this "admin" rather than "group_admin" as I figured it is effectively namespaced inside memberships. I then reread the card and noticed it specified "group_admin" so swapped it out. Open to people's opinions on whether we need to disambiguate this from org and super admins here.

spec/requests/groups_controller_spec.rb
```diff
@@ -166,6 +166,14 @@
 
         expect(Group.last.creator).to eq(editor_user)
       end
+
+      it "gives the creator the group admin role" do
+        login_as_editor_user
+
+        post groups_url, params: { group: valid_attributes }
+
+        expect(Membership.last.role).to eq("group_admin")
```
lfdebrux, line 175: I'm probably being a little overcautious, but it might be worth tightening up this test a little?

```suggestion
        expect(Membership.last).to have_attributes user: editor_user, role: :group_admin
```

jamie-o-wilkinson: Good suggestion, have implemented!

lfdebrux (approved):

PR #1049, opened 2024-03-27
Add page titles for group pages

lfdebrux:

Some of the pages for the groups feature were missing page titles, this commit makes sure that each page has a unique title.


app/views/groups/edit.html.erb
```diff
@@ -1,5 +1,6 @@
+<% set_page_title(title_with_error_prefix(t(".title"), @group.errors.any?)) %>
```
DavidBiddle, line 1: Don't think I've seen this `.title` shorthand before - neat!

DavidBiddle (approved):

PR #1051, opened 2024-03-27
Enable ask user name org

thomasiles:
### Ask user for org and name on sign in, if we don't have it

Trello card: https://trello.com/c/yzbuocwP/1445-launch-ask-for-name-organisation-feature

This PR uses the omniauth callback to redirect the user to the org/name screens if needed. It also adds a `before_action` to stop users who haven't entered these details using the application.

Added the redirects to the omniauth callback means that the redirects are only triggered at sign in.

Using the omniauth callback as a trigger means the behaviour is less surprising. As it's closely tied to the auth process, it doesn't require opting out anywhere and only happens once on sign in.

It does have the disadvantage of not running when not using omniauth. This happens in the user research environment, which uses basic_auth directly instead of omniauth. It doesn't matter there because we directly set the user org and name.

If we replace our current omni-auth authentication scheme, we would need to take this into account. 


app/controllers/account/names_controller.rb
```diff
@@ -25,5 +25,9 @@ def account_name_form_params(user)
     def redirect_if_name_exists
       redirect_to root_path if current_user.name.present?
     end
+
+    def next_path
```
thomasiles, line 29: don't know if repeating this method is better than putting it in application controller. It's hard to find somewhere else to stick it.

DavidBiddle (dismissed):

thomasiles:
Two screenshots from dev:

Screenshot 2024-04-04 at 14 14 06
Screenshot 2024-04-04 at 14 07 16

jamie-o-wilkinson (dismissed):
Neat! 🥇

jamie-o-wilkinson (approved):
The description might want updating with how this is both post sign in and on every request, for posterity. Otherwise this looks good to me.

I've not been able to test locally, but approving on the proviso that this works in dev

thomasiles:
> The description might want updating with how this is both post sign in and on every request, for posterity. Otherwise this looks good to me.
> 
> I've not been able to test locally, but approving on the proviso that this works in dev

Testing in dev seems fine. I've tried various combinations of already being signed in or not with name and org removed from my account.

PR #1053, opened 2024-03-28
Make it easier for users to select their account organisation

lfdebrux:

- Show abbreviations
- Don't show closed organisations
- Sort by name

### DIFF
--- a/app/views/account/organisations/edit.html.erb
+++ b/app/views/account/organisations/edit.html.erb
@@ -10,7 +10,7 @@
         f,
         attribute_name: :organisation_id,
         form_field: f.govuk_collection_select(:organisation_id,
-                                              Organisation.not_closed.order(:slug),
+                                              Organisation.not_closed.order(:name),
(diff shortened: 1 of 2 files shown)


danworth (approved):
Looks good and the new test for not showing closed organisations is clear to understand 🏅 . 

Only a minor musing....I find I tend to err towards the idea that things are simpler to understand when avoiding the negative conditionals, so rather than `not_closed` something like `live` if such a property exists on the organisation. I don't think using `not_closed` here is confusing so don't go out of your way :)

danworth:
That's great Laurence, like I mentioned I don't have a problem with it in this instance, its clear to me too and from the previous PR its obviously been well considered. Thanks 👍

thomasiles (approved):

danworth:
Oh sorry Laurence I meant to approve it not just leave a comment!! Just seen Tom approve it and thought... that's odd.

PR #1054, opened 2024-03-28
Unarchive form

DavidBiddle:

Trello card: 

### DIFF
--- a/app/controllers/forms/unarchive_controller.rb
+++ b/app/controllers/forms/unarchive_controller.rb
@@ -1,0 +1,41 @@
+module Forms
+  class UnarchiveController < ApplicationController
+    after_action :verify_authorized
+
+    def new
+      authorize current_form, :can_make_form_live?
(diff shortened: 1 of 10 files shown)


thomasiles (approved):
This looks good and works for me locally 👍

app/controllers/forms/unarchive_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      params.require(:forms_make_live_form).permit(:confirm_make_live).merge(form: current_form)
+    end
+
+    def user_wants_to_make_form_live
+      @make_live_form.valid? && @make_live_form.made_live?
+    end
+
+    def make_form_live
+      @make_live_form.valid? && @make_form_live_service.make_live
+    end
```
chao-xian, line 39: Can we please abstract these into the MakeLiveForm model instead of being private methods in the controller.

PR #1062, opened 2024-04-02
Remove a negative contraction

hannahkc:
Remove a negative contraction.

Removing a negative contraction as they're not to style. 

Trello card: n/a


sarahseewhy (approved):
LGTM 🚀 

Thank you for catching that!

PR #1063, opened 2024-04-02
Move make live methods into make live form model

DavidBiddle:
Moves the user_wants_to_make_form_live and make_form_live methods out of the make_live_controller and the unarchive_controller and into the make_live_form model.

Trello card: https://trello.com/c/li50etsD/1432-implement-form-unarchiving (tidying up)


spec/form_objects/forms/make_live_form_spec.rb
```diff
@@ -30,4 +30,48 @@
       end
     end
   end
+
+  describe "#user_wants_to_make_form_live" do
```
chao-xian, line 34: Beautiful! TYSM

chao-xian (approved):
Thank you for making this change! Can we also reword the commit message with a "why" that the controller shouldn't have such private methods that are really belonging in the domain of the model itself pls?

DavidBiddle:
Thanks @chao-xian, have added that context to the commit message 👍

PR #1065, opened 2024-04-03
Remove usages of form.has_live_version

stephencdaly:
Replace usages of the `form.has_live_version` property with calls to the `form.is_live?` method which uses the form state.

The `has_live_version` property will be removed from the form API response as this property is confusing now that we have archived forms which technically have a live "version" (a snapshot of the form stored in the `made_live_forms` table) that is disabled.

This is related to trello card: https://trello.com/c/UVCBeTs8/1433-allow-user-to-create-a-draft-for-an-archived-form


chao-xian (approved):
Nice! Is it worth deploying to dev and checking the e2e tests?

stephencdaly:
> Nice! Is it worth deploying to dev and checking the e2e tests?

Are there instructions for doing this? I thought they might be here but this page seems to have disappeared https://github.com/alphagov/forms-team/wiki/Deploying-code-changes-AWS

PR #1066, opened 2024-04-03
Stop adding users without access to mailing lists

DavidBiddle:

Trello card: https://trello.com/c/1B4eVcBh/1711-fix-the-updates-for-platform-users-mailchimp-mailing-list-so-denied-users-are-unsubscribed-automatically

This commit adds a `.where` to the database query which should stop users without access from being added to the mailing list. This means that users who no longer use the platform (e.g. people who've moved teams or people whose organisations can't use the platform) won't get emails which aren't relevant to them.

Also adds tests to make sure these users aren't added to the lists and that any existing users without access will be removed from the list.

### DIFF
--- a/lib/tasks/mailchimp.rake
+++ b/lib/tasks/mailchimp.rake
@@ -17,7 +17,7 @@ namespace :mailchimp do
       server: mailchimp_server_prefix,
     })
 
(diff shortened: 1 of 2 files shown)


jamie-o-wilkinson (approved):
Nice one. Thanks! 🥇

PR #1068, opened 2024-04-03
Don't show delete button for archived form

stephencdaly:

Trello card: https://trello.com/c/UVCBeTs8/1433-allow-user-to-create-a-draft-for-an-archived-form

It should only be possible to delete a form if it is in draft state, so only show the button when a form is in this state.


DavidBiddle (approved):

thomasiles (approved):
👍

PR #1074, opened 2024-04-04
Handle paginated Mailchimp API response

DavidBiddle:

Trello card: https://trello.com/c/1B4eVcBh/1466-fix-the-updates-for-platform-users-mailchimp-mailing-list-so-denied-users-are-unsubscribed-automatically

We had some issues with the Mailchimp task failing to unsubscribe people from the mailing lists when they weren't active users in the database. The cause for this is that the Mailchimp API paginates its response when requesting a list of subscribers. By default it will only return 10 results. This means the comparisons of existing members vs users in the database in our rake task didn't actually return the right results - we were comparing the list of users in the database against the first 10 subscribers of each Mailchimp list.

This PR adds some logic to handle the pagination, and requests the maximum page size (1000 results) in each request.

### DIFF
--- a/lib/tasks/mailchimp.rake
+++ b/lib/tasks/mailchimp.rake
@@ -29,9 +29,22 @@ namespace :mailchimp do
       puts "Found Mailchimp list: #{target_list['name']}"
       puts "Mailchimp list has #{target_list['stats']['member_count']} members"
 
-      existing_members = mailchimp.lists.get_list_members_info(list_id)
-      list_email_addresses = existing_members["members"]
-                               .map { |member| member["email_address"] }
+      existing_members = []
+
+      # Set up API pagination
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
nice!

I guess there are never that many users subscribed so we don't need to rate limit the mailchimp calls.

DavidBiddle:
@thomasiles good point, thanks for raising it! I hadn't really thought about the API limits but Mailchimp say:

> The Marketing API has a limit of 10 simultaneous connections. You’ll receive a [429 error](https://mailchimp.com/developer/marketing/docs/errors/#error-glossary) if you reach the limit. At exceptionally high volumes, you may receive an HTTP 429 or 403 without a JSON body.

So I think we should be alright - I don't think we ever have multiple connections open at the same time, so I think we should be OK even if we see a big increase in users.

PR #1075, opened 2024-04-04
Update feedback link

chao-xian:

Update feedback link to point to a survey for user research instead of the contact email.

[Trello card](https://trello.com/c/xJd9DCp5)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-# Rubocop ate my lunch
 module ApplicationHelper
   def page_title(separator = " – ")
     [content_for(:title), "GOV.UK Forms"].compact.join(separator)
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -33,7 +32,7 @@ def link_to_runner(runner_url, form_id, form_slug, mode: :preview_draft)
   end
 
   def contact_url
-    "mailto:govuk-forms-support@govuk.zendesk.com"
+    t("contact_url")
```
DavidBiddle, line 35: I don't think we can change this here - it's used in the `contact_link` helper below. That's used in other places (like [the 403 page](https://github.com/alphagov/forms-admin/blob/2ef3760b74fca7ddfd252c414f88462b0fa09191/app/views/errors/forbidden.html.erb#L6)) where I think we still want the email link rather than this survey

chao-xian: Ah interesting! OK thanks I'll rework it

DavidBiddle: The refactor looks good overall, but I think to update the link in the phase banner we'll need to change it somewhere more specific

DavidBiddle: Update looks good!

DavidBiddle (approved):
Nice refactor & works for me locally 🎉

PR #1076, opened 2024-04-04
Fix question text page title

DavidBiddle:

Trello card: https://trello.com/c/RkF3fgU1/1462-whats-your-question-page-has-the-wrong-title

Previously this page had the same title as the date settings page (presumably some copy-pasted content that wasn't replaced). As well as being a bug this was a violation of WCAG 2.4.2 Page Titled, because the page title did not accurately describe the page.

This PR makes the title match the h1/form field text.

Also includes adding some view tests for this file.


thomasiles (approved):
nice!

PR #1079, opened 2024-04-05
Add initial Group members screen

thomasiles:
### Add new page to show members of a Group

This PR starts the work for managing members of a group.

It only tackles the basic journey of showing the members of a group. This is what members of a group with the role editor will see. Other members, group admins, org admins and super admins will all have options for adding and removing users fro the group.

image

When the group doesn't have any members, I've added some new content (which can be changed). I think this screen won't happen often but super admins and org admins might see it if all members are removed from a group:
image

Link from the group page:
image

### DIFF
--- a/app/controllers/group_members_controller.rb
+++ b/app/controllers/group_members_controller.rb
@@ -1,0 +1,11 @@
+class GroupMembersController < ApplicationController
+  before_action :set_group
+
+  def index; end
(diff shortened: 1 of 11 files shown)


DavidBiddle (approved):
Code looks great, works on my machine, and good content too 💯

PR #1080, opened 2024-04-05
Update change link hidden text

DavidBiddle:

Trello card: https://trello.com/c/tJswYdpy/1471-page-settings-summary-change-link-accessible-name-does-not-match-the-visible-label

The main point to this PR is to update the change link visually hidden text in the page settings summary at the bottom of the edit question page.

As well as doing this, there are a few small refactors/cleanup jobs included, namely:
- Using translated values for these bits of text
- Changing the format of the 'Change answer type' link for consistency
- Some cleanup in the test files
- Moving some logic out of the view file

(see individual commit messages for details)


thomasiles (approved):
very neat!

PR #1084, opened 2024-04-08
Update content for unarchiving a form

chao-xian:

Content that was added as part of the unarchiving journey was (quite pragmatically) added by devs as placeholder. We now have the proper content from our Content Designer and this PR updates to that.

[Trello card](https://trello.com/c/K2gHWdLY)

## Screenshots

<details>
<summary>When making an archived form live again (without making a draft first)</summary>

localhost_3000_forms_4_unarchive
</details>

<details>
<summary>When making a draft of an archived form live</summary>

localhost_3000_forms_1_make-live (1)

localhost_3000_forms_4_make-live (1)

</details>

<details>

<summary>For the draft of an archived form task list page</summary>

localhost_3000_forms_9

</details>

### DIFF
--- a/app/service/form_task_list_service.rb
+++ b/app/service/form_task_list_service.rb
@@ -32,7 +32,10 @@ class FormTaskListService
 private
 
   def create_or_edit
-    @form.is_live? ? "edit" : "create"
+    return "edit" if @form.is_live?
+    return "edit" if @form.is_archived?
+
+    "create"
   end
 
   def create_form_section
@@ -116,7 +119,7 @@ private
 
   def make_form_live_section
     section = {
-      title: I18n.t("forms.task_list_#{create_or_edit}.make_form_live_section.title"),
+      title: live_title_name,
       section_number: 4,
       subsection: false,
     }
@@ -132,11 +135,29 @@ private
 
   def make_form_live_section_tasks
     [{
-      task_name: I18n.t("forms.task_list_#{create_or_edit}.make_form_live_section.make_live"),
-      path: @form.all_ready_for_live? ? make_live_path(@form.id) : "",
+      task_name: live_task_name,
+      path: live_path,
       status: @task_statuses[:make_live_status],
       active: @form.all_ready_for_live?,
     }]
+  end
+
+  def live_title_name
+    return I18n.t("forms.task_list_create.make_form_live_section.title") if @form.is_archived?
+
+    I18n.t("forms.task_list_#{create_or_edit}.make_form_live_section.title")
+  end
+
+  def live_task_name
+    return I18n.t("forms.task_list_create.make_form_live_section.make_live") if @form.is_archived?
+
+    I18n.t("forms.task_list_#{create_or_edit}.make_form_live_section.make_live")
+  end
+
+  def live_path
+    return "" unless @form.all_ready_for_live?
+
+    make_live_path(@form.id)
   end
 
   def statuses_by_user
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -73,6 +73,7 @@ en:
         <ul class="govuk-list govuk-list--bullet">
           <li>links to the form will no longer work - make sure any links to the form have been removed</li>
           <li>anyone who is part way through completing the form will lose their progress and get an error page</li>
+          <li>we’ll send an email to %{submission_email} to let them know the form has been archived</li>
         </ul>
         <p>You will still be able to preview the form and see its information and settings. You can also make a new draft of the form and make it live again if you need to.</p>
       radios_legend: Are you sure you want to archive this form?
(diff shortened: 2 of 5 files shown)


config/locales/en.yml
```diff
@@ -65,6 +65,7 @@ en:
         <ul class="govuk-list govuk-list--bullet">
           <li>links to the form will no longer work - make sure any links to the form have been removed</li>
           <li>anyone who is part way through completing the form will lose their progress and get an error page</li>
+          <li>we'll send an email to %{submission_email} to let them know the form has been archived</li>
```
hannahkc, line 68:
```suggestion
          <li>we’ll send an email to %{submission_email} to let them know the form has been archived</li>
```

hannahkc, line 68:
```suggestion
          <li>we’ll send an email to %{submission_email} to let them know the form has been archived</li>
```

hannahkc:
The content on the "When making a draft of an archived form live" page is wrong. It should be the same as the content on the "When making an archived form live again (without making a draft first)" page. 

(because there won't be any impact on people who are filling in the live form as there is no live form).

chao-xian:
@hannahkc thanks for checking I'll get on it

DavidBiddle (dismissed):
Code looks good and works for me locally :tada:

Edit: didn't see Hannah's comment!

chao-xian:
@hannahkc @DavidBiddle I've realised what I missed out now - the link to making live wasn't right from the task list for a draft for an archived form. Should be correct now

app/models/form.rb
```diff
@@ -34,6 +34,10 @@ def is_archived?
     state.to_sym.in?(%i[archived archived_with_draft])
   end
 
+  def is_archive_with_draft?
+    state.to_sym == :archived_with_draft
+  end
```
lfdebrux, line 39: The commit message adding this method says it will be needed in the task list, but I can't see it used anywhere 🤔

app/service/form_task_list_service.rb
```diff
@@ -132,13 +132,26 @@ def make_form_live_section
...
+    return I18n.t("forms.task_list_#{create_or_edit}.make_form_live_section.make_changes_live") if @form.is_archived?
+
+    I18n.t("forms.task_list_#{create_or_edit}.make_form_live_section.make_live")
+  end
+
+  def live_path
+    return "" unless @form.all_ready_for_live?
+    return unarchive_path(@form.id) if @form.is_archived?
+
+    make_live_path(@form.id)
```
DavidBiddle, line 152: The make_live controller which the make_live_path points to already has logic for switching between the normal make live view and a new `make_archived_draft_live` view. Rather than adding this logic here I think the simplest way to fix the issue Hannah described would be to change the content in that view to match that in the unarchive view.

The way this is implemented in the most recent commit means that:
- the `make_archived_draft_live.html.erb` view is never rendered, so it and the switching logic in the make_live_controller are redundant
- the back link and vallidation on the unarchive page won't take the user back to the task list, they'll take them back to the archived form page.

I'll leave it up to you whether `make_archived_draft_live.html.erb` should be removed or not, but if we do remove it we need to preserve the different back link/validation behaviour depending on where the user has come from.

app/models/form.rb
chao-xian: ah crumbs I thought I removed it as part of dropping a commit - rebase error

app/service/form_task_list_service.rb
chao-xian: Right I think I've got it this time...

DavidBiddle (dismissed):
Looks good and works for me locally 🎉

hannahkc:
The 'Make your form live again' screenshots look grand to me. 

I can't see the bottom of the draft of an archived form task list in the attached screenshot. If the final section and task are both 'Make your form live' (NOT 'Make your changes live') then this is all good.

hannahkc:
Oh wait sorry, the tasklist DOES have the wrong content. It has the right title - Edit your form - but it looks to be using the section and task content for the 'Create a form' tasklist that's only for when you are FIRST making a form. 

It should be using the section and task content from the 'Edit your form' tasklist. So, for example, the first sections should be: 

- Edit the name of your form
- Edit your questions
- Edit the declaration for people to agree to
- Edit the information about what happens next

The two changes we needed to this were to remove the inset text at the top and to change the section title and task title for the last section to 'Make your form live'. 

Sorry!

chao-xian:
@hannahkc has reviewed the screens as running on my machine and all looks good.

thomasiles (dismissed):
LGTM - I haven't checked the content but I'm assuming you and Hannah covered that.

Might be worth sticking a few tests in the task_list_service_spec to cover the new stuff when you get a chance.

chao-xian:
@thomasiles great shout missed that

thomasiles (approved):
🏅

PR #1088, opened 2024-04-10
Mailchimp mou

DavidBiddle:

Trello card: https://trello.com/c/04ivsTJK/1459-mou-mailing-list-has-lots-more-subscribers-than-it-should-do


thomasiles (approved):
adding `mailchimp_list_synchronizer.rb` makes everything so much clearer 🥇

PR #1095, opened 2024-04-12
Delete members from groups

thomasiles:
### Allow group admins to remove editors from groups

We add a new membership controller to avoid adding too many routes to the nested resource group_member. There is a new policy to permit "admins" to remove members and a new service to avoid sticking too much logic into the view.

As a group admin:

image

The designs:
image

### DIFF
--- a/app/service/group_member_service.rb
+++ b/app/service/group_member_service.rb
@@ -15,7 +15,7 @@ class GroupMemberService
   end
 
   def show_actions?
-    current_user.super_admin? || current_user.is_organisations_admin?(group.organisation) || current_user.is_group_admin?(group)
+    @show_actions ||= current_user.super_admin? || current_user.is_organisations_admin?(group.organisation) || current_user.is_group_admin?(group)
   end
 
   def rows
--- a/app/service/group_member_service.rb
+++ b/app/service/group_member_service.rb
@@ -15,11 +15,11 @@ class GroupMemberService
   end
 
   def show_actions?
-    @show_actions ||= current_user.super_admin? || current_user.is_organisations_admin?(group.organisation) || current_user.is_group_admin?(group)
+    rows.any? { |row| row.actions.any? }
   end
 
   def rows
-    group.memberships.map(&method(:row))
+    @rows ||= group.memberships.map(&method(:row))
   end
 
 private
--- a/app/service/group_member_service.rb
+++ b/app/service/group_member_service.rb
@@ -1,0 +1,46 @@
+class GroupMemberService
+  Row = Struct.new(:name, :email, :role, :actions, :membership)
+
+  attr_reader :group, :current_user
+
+  class << self
+    def call(**args)
+      new(**args)
+    end
+  end
+
+  def initialize(group:, current_user:)
+    @group = group
+    @current_user = current_user
+  end
+
+  def show_actions?
+    current_user.super_admin? || current_user.is_organisations_admin?(group.organisation) || current_user.is_group_admin?(group)
+  end
+
+  def rows
+    group.memberships.map(&method(:row))
+  end
+
+private
+
+  def row(membership)
+    Row.new(
+      name: membership.user.name,
+      email: membership.user.email,
+      role: membership.role,
+      actions: actions(membership),
+      membership:,
+    )
+  end
+
+  def actions(membership)
+    actions = []
+
+    if Pundit.policy(current_user, membership).destroy?
+      actions << :delete
+    end
+
+    actions
+  end
+end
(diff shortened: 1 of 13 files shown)


DavidBiddle (approved):

spec/requests/memberships_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+
+describe "/memberships", type: :request do
+  let(:user) { create(:user) }
+  let(:group) { create(:group) }
+  let(:role) { :group_admin }
```
SamJamCul, line 6: Could we rename this to make it easier to separate out the logged-in-user's role from the membership-being-deleted's role

```suggestion
  let(:logged_in_user_role) { :group_admin }
```

spec/requests/memberships_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  before do
+    create(:membership, user: editor_user, group:, role:)
```
SamJamCul, line 9:
```suggestion
    create(:membership, user: editor_user, group:, role: logged_in_user_role)
```

spec/requests/memberships_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      expect {
+        delete membership_path(membership)
+      }.to change(Membership, :count).by(-1)
+
+      expect(response).to redirect_to(group_members_path(group.external_id))
+    end
+
+    context "when user is not a group admin" do
+      let(:role) { :editor }
```
SamJamCul, line 25:
```suggestion
      let(:logged_in_user_role) { :editor }
```

spec/requests/memberships_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      membership = create(:membership, user:, group:, added_by: editor_user)
+
+      expect {
+        delete membership_path(membership)
+      }.to change(Membership, :count).by(-1)
+
+      expect(response).to redirect_to(group_members_path(group.external_id))
+    end
+
+    context "when user is not a group admin" do
```
SamJamCul, line 24:
```suggestion
    context "when logged in user is not a group admin" do
```

SamJamCul (approved):

app/service/group_member_service.rb
```diff
@@ -0,0 +1,46 @@
...
+      new(**args)
+    end
+  end
+
+  def initialize(group:, current_user:)
+    @group = group
+    @current_user = current_user
+  end
+
+  def show_actions?
```
SamJamCul, line 17: Should we just check if there are any actions that have been added to the rows, rather than the user's role? If we've added an action in our row setup, we'd definitely want to `show_actions?` anyway

thomasiles: I thought this originally but I think we need to show an empty actions column when the current_user is a group admin and there are no actions. This happens when there are no editors in the group.

SamJamCul: Cool, makes sense then!

app/service/group_member_service.rb
```diff
@@ -0,0 +1,46 @@
...
+    current_user.super_admin? || current_user.is_organisations_admin?(group.organisation) || current_user.is_group_admin?(group)
+  end
+
+  def rows
```
SamJamCul, line 21: Is it worth storing this in some way, as we're calling it a couple of times in the view, and it'll be running through the row setup each time

spec/views/group_members/index.html.erb_spec.rb
```diff
@@ -2,25 +2,21 @@
 
 RSpec.describe "group_members/index", type: :view do
   let(:organisation) { build(:organisation, slug: "Department for testing group members") }
-  let(:user1) { build(:user, organisation:) }
-  let(:user2) { build(:user, organisation:) }
   let(:group) { create(:group, name: "Group 1", organisation:) }
   let(:add_editor) { false }
 
+  let(:rows) { [OpenStruct.new({ name: "user1", email: "user1@gov.uk", role: :editor, actions: [] })] }
```
SamJamCul, line 8: Would it be possible to also test whether a `:delete` action has triggered the button appearing?

thomasiles: I've changed the tests to check for this.

app/service/group_member_service.rb
thomasiles: I misread this but now rows is stored.

app/service/group_member_service.rb
thomasiles: wait! I checked with the designers and they said don't show the column if there are no actions, so you were right.

I've updated the code and tests.

SamJamCul (approved):
Sorry about the delay, I think this is good to go now

PR #1096, opened 2024-04-16
Rake tasks for pushing commits to the dev environment

Rkotyank:
What problem does this pull request solve?

[Trello card](https://trello.com/c/Z8G6L6zt/388-add-dev-environment-support-tooling-to-all-repos)

To push a commit to a dev environment, we invoke an image building pipeline to produce an image with a specific prefix ("dev-"), which triggers the dev deployment pipelines.

This commit adds two Rake tasks:

push_to_dev
reset_dev_to_main

The former deploys the commit at the tip of the current branch to dev. The latter deploys the commit at the tip of main, after updating the remotes to get the very latest commits.

### DIFF
--- a/lib/tasks/deploy.rake
+++ b/lib/tasks/deploy.rake
@@ -1,0 +1,77 @@
+require "English"
+require "json"
+require "net/http"
+require "shellwords"
+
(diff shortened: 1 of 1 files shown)


sarahseewhy (approved):
LGTM 🚀

PR #1102, opened 2024-04-17
Add new member emails

thomasiles:
### Send notification emails when a user is added to a group

When a user is added to a group, they should receive an email notifying them.

This PR adds a new mailer which is used when a new membership is created. A new template has been added to notify with conditional text for the two cases: a new editor or a new group admin.

The design:
image

A sample email sent when joining a group (it's not shown in this image but the template is now "an editor" not "a editor")
image

### DIFF
--- a/app/controllers/group_members_controller.rb
+++ b/app/controllers/group_members_controller.rb
@@ -29,6 +29,7 @@ private
   end
 
   def group_member_params
-    params.require(:group_member_form).permit(:member_email_address).merge(group: @group, creator: current_user)
+    ## TODO: We are passing in host here because the admin doesn't know it's own URL to use in emails
+    params.require(:group_member_form).permit(:member_email_address).merge(group: @group, creator: current_user, host: request.host)
   end
 end
--- a/app/form_objects/group_member_form.rb
+++ b/app/form_objects/group_member_form.rb
@@ -18,10 +18,10 @@ class GroupMemberForm < BaseForm
       return false
     end
 
-    if new_membership.save
-      send_notification_email
-      true
-    end
+    new_membership.save!
+    send_notification_email
+
+    true
   end
 
 private
--- a/app/form_objects/group_member_form.rb
+++ b/app/form_objects/group_member_form.rb
@@ -1,7 +1,8 @@
 class GroupMemberForm < BaseForm
   include ActiveModel::Validations::Callbacks
+  include Rails.application.routes.url_helpers
 
-  attr_accessor :member_email_address, :group, :creator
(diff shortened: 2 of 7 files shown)


SamJamCul (approved):
Looking good!

app/form_objects/group_member_form.rb
```diff
@@ -17,7 +18,10 @@ def save
       return false
     end
 
-    new_membership.save!
+    if new_membership.save
+      send_notification_email
+      true
+    end
```
lfdebrux, line 24:
```suggestion
    new_membership.save!
    send_notification_email
    
    true
```

thomasiles: yes, that makes sense but the controller relies on the return value:

  @group_member_form = GroupMemberForm.new(group_member_params)
    if @group_member_form.save
      redirect_to group_members_path(@group.external_id)
    else
      render :new, status: :unprocessable_entity
    end

lfdebrux (approved):
Works on my machine!

app/controllers/group_members_controller.rb
```diff
@@ -29,6 +29,7 @@ def set_group
   end
 
   def group_member_params
-    params.require(:group_member_form).permit(:member_email_address).merge(group: @group, creator: current_user)
+    ## TODO: We are passing in host here because the admin doesn't know it's own URL to use in emails
```
DavidBiddle, line 32: This is a TODO but I'm not entirely sure what needs to be done - is there a card that can be referenced here?

DavidBiddle (commented):
Works for me locally and code looks good 🎉 had one small question about a TODO comment

app/controllers/group_members_controller.rb
thomasiles: yes, I should make one.

It's related to the point we discussed about the admin's URL in tech catchup on monday: https://docs.google.com/document/d/1iHfUoCzgg-2Mk9wbi5o46fFVRak_R6nNQ2m8OqSz_hs/edit

thomasiles: https://trello.com/c/zSQq9xee/1773-configure-admin-to-use-url-from-env-in-email-config

thomasiles:
> Works for me locally and code looks good 🎉 had one small question about a TODO comment

thanks, I've added two trello cards after the tech standup to cover this. It's not a big ticket so the cleanup shouldn't be too difficult.

https://trello.com/b/e28NIDBJ/govuk-forms-product-backlog-private-beta

I marked our conversation as resolved to merge this. I'm happy to discuss it further though if its not clear or more work needs to be done.

PR #1108, opened 2024-04-23
Use task list component

DavidBiddle:

Trello card: https://trello.com/c/z0qYqjJE/1397-replace-custom-task-list-component-with-govuk-frontend-code

Adapts our task list component to make use of the component that's included in the Design System. 

Main differences are:
- The whole of each row is now a link, with a special hover style
- The tag colours have changed
- The 'cannot start yet' and 'completed' statuses appear as plain text rather than tags
- We no longer indent each list

Also includes a few small refactors, mainly to handle less of the logic in the view.html.erb file (see individual commits for details).

#### Before
Screen Shot 2024-04-23 at 16 52 04

#### After
Screen Shot 2024-04-23 at 16 51 31

### DIFF
--- a/app/components/task_list_component/view.html.erb
+++ b/app/components/task_list_component/view.html.erb
@@ -24,7 +24,7 @@
           <%= govuk_task_list(id_prefix: section.title.gsub(" ", "-").downcase, classes: "app-task-list__items") do |task_list| %>
             <% section.rows.each do |row| %>
               <% task_list.with_item(classes: "app-task-list__item") do |item| %>
-                <% item.with_title(text: row.task_name, href: row.active ? row.get_path : nil, hint: row.hint_text&.html_safe ) %>
+                <% item.with_title(text: row.task_name, href: row.get_path, hint: row.hint_text&.html_safe ) %>
                 <% item.with_status(text: govuk_tag(text: I18n.t("task_statuses.#{row.status}"), colour: row.get_status_colour), cannot_start_yet: row.status == :cannot_start) %>
               <% end %>
             <% end %>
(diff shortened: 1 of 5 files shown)


app/components/task_list_component/view.html.erb
```diff
@@ -3,55 +3,32 @@
...
-                  classes: "app-task-list__tag",
-                  colour: row.get_status_colour,
-                  html_attributes: { id: row.status_id },
-                ) %>
-            <% end %>
-
-            <% if row.hint_text.present? %>
-              <br>
-              <span class="app-task-list__hint govuk-hint"><%= row.hint_text.html_safe %></span>
+            <% task_list.with_item(classes: "app-task-list__item") do |item| %>
```
stephencdaly, line 25: Same with `app-task-list__item`, can this be removed?

DavidBiddle: Done!

app/components/task_list_component/view.html.erb
```diff
@@ -3,55 +3,32 @@
...
         <div class="app-task-list__body">
           <%= simple_format(section.body_text) %>
         </div>
-        <% end %>
+      <% end %>
 
-        <% unless section.rows.empty? %>
-        <ul class="app-task-list__items">
+      <% unless section.rows.empty? %>
+        <%= govuk_task_list(id_prefix: section.title.gsub(" ", "-").downcase, classes: "app-task-list__items") do |task_list| %>
```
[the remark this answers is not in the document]
DavidBiddle: Done!

stephencdaly (approved):
Nice work! I've tested locally and it all seems to work.

PR #1113, opened 2024-04-24
Update notification banners to use h3s where appropriate

DavidBiddle:

Trello card: https://trello.com/c/V7Xpdemm/1477-trial-account-banner-content-has-been-styled-as-a-heading-but-is-not-programmatically-a-heading

Most of our notification banners contain a heading and some other content, and we were using the govuk-components gem's `with_heading` helper to add the heading. Unfortunately, the `with_heading` helper generates a `p` element by default, because the design system allows the `govuk-notification-banner__heading` to be used for both [large notification text which isn't semantically a heading](https://design-system.service.gov.uk/components/notification-banner/whole-service/) and as [a heading for other content](https://design-system.service.gov.uk/components/notification-banner/success/). 

This PR supplies the `with_heading` function with the new `tag` argument. We now set the `tag` value to:
- `h3` if the notification banner also contains other content as well as the heading
- `p` if the notification banner doesn't have any other text in.

Also includes:
- rejigging the trial group message to use `with_heading` for consistency.
- removing the MakeLiveWarning component, a component which seems to only exist to supply a nitification abnner we no longer use.

No screenshots as there's no visual change.


thomasiles (approved):
nice! works locally and tags look right 👍

PR #1116, opened 2024-04-25
Add base form object for forms that confirm an action

stephencdaly:

Trello card: https://trello.com/c/XCLShuwW/1483-upgrade-groups-org-admin-can-upgrade-a-group-without-a-request

As part of this trello card, we are adding a page with Yes/No radio buttons for the user to confirm that they want to archive the form.

We already have a few pages for different actions where we ask the user to select "Yes" or "No" to confirm whether they want to continue with the action or not, and we follow slightly different patterns for some of these. Add a base form object that can be inherited for forms that need simple yes/no radios so this code is re-used.


app/views/forms/delete_confirmation/delete.html.erb
```diff
@@ -5,7 +5,8 @@
     <%= f.govuk_error_summary %>
   <% end %>
   <span class="govuk-caption-l"><%= @item_name %></span>
-  <%= f.govuk_collection_radio_buttons :confirm_deletion, @confirm_deletion_options, :value,
+  <%= f.govuk_collection_radio_buttons :confirm,
```
chao-xian, line 8: I'm not clear what's going on here?

chao-xian (commented):
Looks really sensible and great overall but I'm not clear on where we call the radio button component what's going on there

app/views/forms/delete_confirmation/delete.html.erb
stephencdaly: I wasn't either, and I'm still not entirely, but it's following this https://govuk-form-builder.netlify.app/form-elements/radios/#populating-values-labels-and-hints-with-procs

chao-xian: ah ok... thanks for pointing me to that! had no idea.

chao-xian (approved):

PR #1127, opened 2024-04-30
Rename form objects

stephencdaly:

Rename form objects to have suffix "Input".

The suffix "Form" for form objects which represent the input fields on a HTML form was difficult to work with, as we have a "Form" entity represing a GOV.UK Forms form, which is used throughout the application.

Use the suffix "Input" for form objects, to hopefully avoid this point of confusion.


lfdebrux (approved):
Looks good to me, I've tested locally and found no issues.

PR #1128, opened 2024-04-30
Allow admins to change the role of group members

thomasiles:

Add the extra buttons to allow organisation and super admins to change the role of group members.

image


app/policies/membership_policy.rb
```diff
@@ -3,6 +3,10 @@ def destroy?
     user.super_admin? || organisation_admin? || (group_admin? && record.editor?)
   end
 
+  def update?
```
thomasiles, line 6: we are using update as as the policy because it's a standard route and makes sense given the current requirements.

It could be that we want something more specific in the future, e.g `can_change_role?` but for now I think this is the simplest option?

app/views/group_members/index.html.erb
```diff
@@ -28,9 +28,21 @@
...
+
+              if member.actions.include?(:make_group_admin)
+                actions << govuk_button_to(t(".make_group_admin"), membership_path(member.membership), method: :patch, params: { membership: { role: Membership.roles[:group_admin] }}, secondary: true)
               end
+
+              if member.actions.include?(:make_editor)
+                actions << govuk_button_to(t(".make_editor"), membership_path(member.membership), method: :patch, params: { membership: { role: Membership.roles[:editor] }}, secondary: true)
+              end
+
+              safe_join(actions)
```
thomasiles, line 45: This view code isn't the nicest but separating the actions seemed like the best option.

I first tried generating the HTML for the buttons in a service but that requires passing the CSRF token, because the buttons are acutally forms.

That didn't seem any clearer than keeping the logic of what buttons are required out of the view, which this does.

config/brakeman.ignore
```diff
@@ -1,5 +1,28 @@
...
+        "type": "method",
+        "class": "MembershipsController",
+        "method": "membership_params"
+      },
+      "user_input": ":role",
+      "confidence": "Medium",
+      "cwe_id": [
+        915
+      ],
+      "note": "Changing role for Membership is allowed."
```
thomasiles, line 24: brakeman gets upset because it sees the word "role" and thinks it's dangerous to attach it to a form.

I think this is fine though. It's not possible to set it to admin and gain privileges given the policy and strong params.

app/views/group_members/index.html.erb
jamie-o-wilkinson: Tom and I had a chat about this and couldn't immediately come up with a more elegant solution...

jamie-o-wilkinson (approved):
LGTM

app/controllers/memberships_controller.rb
```diff
@@ -1,13 +1,35 @@
 class MembershipsController < ApplicationController
+  before_action :set_membership, only: %i[destroy update]
```
aliuk2012, line 2: Why not create `membership` as a memorised method then its only created and stored the first time its used instead of creating before_actions

```suggestion
  def membership
    @membership ||= Membership.find(params[:id])
  end
```

app/views/group_members/index.html.erb
thomasiles: hello! 👋 

My initial approach was to follow that pattern and generate the headings, rows and action links in another service and just pass them into the component in the view.

I hit an issue with delete though and haven't found a satisfactory solution. The buttons are actually forms and need to have the CSRF token included in them, otherwise they don't work. The current CSRF token will be int them if they are generated in a view, a helper, the controller or a view component.

The only way I could find to use a service would be to pass it explicitly when creating the service. This makes things a bit ugly, now the service feels like its responsible for something which is really a view concern and is harder to test.

I also tried putting it into a view component but remembered your advice about keeping them as 'dumb' containers, so that didn't feel great either.

Any suggestions? (and thanks for commenting, it's great to hear from you)

aliuk2012: Hey @thomasiles 👋 

🤔  oh yes, I see what you mean about forms and CSRF issues. Maybe a better way to output the forms would be to use pundit policy in the view to control which buttons are shown. There are benefits in that you do not have to have a bunch of `member.actions.include?` dotted around the codebase and if you ever need to refactor its a case of refactoring the pundit policy. It also means that you don't have to build up an array of actions and then forced to use safe_join and pundit controls authorisation for the control actions and the UI components that the user will see.

pseudo example:

```
govuk_button_to(t(".remove_member"), membership_path(member.membership), method: :delete, secondary: true) if policy(membership).delete?
govuk_button_to(t(".make_group_admin"), membership_path(member.membership), method: :patch, params: { membership: { role: Membership.roles[:group_admin] }}, secondary: true) if policy(membership).make_admin?

```

thomasiles:
This was superceded by https://github.com/alphagov/forms-admin/pull/1155 where the comments where taken into account.

PR #1131, opened 2024-05-01
Fix Editing Multiple Forms

danworth:
When draft questions are created they are separated by their form_id and user_id. The `clear_draft_questions_data` was originally deleting all draft forms for the user regardless of the form they were associated with. This commit:
1. when clearing draft forms it clears them for the current form and user only.
2. moves the `clear_draft_questions_data` function and the `before_action` call into the PagesController to which it is more closely related. This should help future understanding and avoids users inadvertently clearing draft questions when visiting other pages unrelated to draft questions in another tab.

The above is done in a single commit. Attempting to introduce the form_id into the condition of `clear_draft_question_data` whilst it remains in the `ApplicationController` would require wide spread changes to tests to mock the active resource for getting the form. It doesn't seem valuable to do this when the intention is to move it to the PagesController immediately afterwards.

Trello card: https://trello.com/c/ELMRrABD/382-end-to-end-test-locking
https://trello.com/c/EVSrkyxX/1518-fix-parallel-end-to-end-tests

**First problem:** Allows the same user to edit multiple forms by clearing draft questions by user and form id. Essentially the problem was:
1. Add a question to form 1
2. In another tab add a question to form 2. this invoked `clear_draft_question_data` clearing all draft questions for the user, including the draft question above for form 1 because it didn't consider the form_id.
4. Go back to the first tab and the state has been lost, so attempting to progress throws various errors depending on the next page but one of them is `No answer type set for draft question (RuntimeError)`

**Second problem:** Allows users to visit non PageController pages without inadvertently clearing draft questions. This is achieved by moving the `clear_draft_question_data` into the PagesController rather than it being invoked in the ApplicationController on almost all pages. I also believe this helps with comprehension since the clearing draft questions is the responsibility of the PagesController.
1. Add a question to form 1
2. In another tab visit almost any other page which invoked `clear_draft_question_data`
3. Go back to the first tab and the state has been lost, so attempting to progress throws various errors depending on the next page but one of them is `No answer type set for draft question (RuntimeError)`

I've tested this locally and deployed to dev whilst running multiple e2e tests at once (previous to this change those tests threw the errors seen during deployments).

image

I don't think there is ever a need to clear all draft questions without knowing the form so I have not kept that but I'm open to opinions; I don't think you can get the pages controller without being signed in and referencing a particular form. **Are there any nasty edge cases lurking here?**

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,8 +11,6 @@ class ApplicationController < ActionController::Base
 
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
-  before_action :clear_draft_questions_data
-
   add_flash_types :success
 
   rescue_from ActiveResource::ResourceNotFound do
@@ -55,10 +53,6 @@ class ApplicationController < ActionController::Base
     payload[:page_id] = params[:page_id] if params[:page_id].present?
     payload[:session_id_hash] = Digest::SHA256.hexdigest session.id.to_s if session.exists?
     payload[:trace_id] = request.env["HTTP_X_AMZN_TRACE_ID"].presence
-  end
-
-  def clear_draft_questions_data
-    current_user.draft_questions.destroy_all if current_user.present?
   end
 
   def set_request_id
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,6 +1,6 @@
 class PagesController < ApplicationController
   before_action :check_user_has_permission
-  skip_before_action :clear_draft_questions_data, except: %i[index move_page]
+  before_action :clear_draft_questions_data, only: %i[index move_page]
   after_action :verify_authorized
 
   def index
@@ -29,6 +29,10 @@ class PagesController < ApplicationController
   end
 
 private
+
+  def clear_draft_questions_data
+    DraftQuestion.destroy_by(form_id: current_form.id, user_id: current_user.id) if current_user.present? && current_form.present?
+  end
 
   def check_user_has_permission
     authorize current_form, :can_view_form?
--- a/app/controllers/heartbeat_controller.rb
+++ b/app/controllers/heartbeat_controller.rb
@@ -1,5 +1,5 @@
 class HeartbeatController < ApplicationController
-  skip_before_action :authenticate_and_check_access, :check_maintenance_mode_is_enabled, :clear_draft_questions_data, :redirect_if_account_not_completed
+  skip_before_action :authenticate_and_check_access, :check_maintenance_mode_is_enabled, :redirect_if_account_not_completed
 
   def ping
     render(body: "PONG")
--- a/spec/controller/application_controller_spec.rb
+++ b/spec/controller/application_controller_spec.rb
@@ -63,23 +63,6 @@ describe ApplicationController, type: :controller do
     end
   end
 
-  describe "#clear_draft_questions_data" do
-    let(:user) { create(:user) }
-
-    it "destroys draft questions when user is present" do
-      allow(controller).to receive(:current_user).and_return(user)
-      create_list(:draft_question, 3, user:)
-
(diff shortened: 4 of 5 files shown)


app/controllers/pages_controller.rb
```diff
@@ -1,6 +1,6 @@
 class PagesController < ApplicationController
   before_action :check_user_has_permission
-  skip_before_action :clear_draft_questions_data, except: %i[index move_page]
+  before_action :clear_draft_questions_data, only: %i[index move_page]
```
lfdebrux, line 3: I like this change, it is easier to understand for me. I'm pretty sure it means the same thing now you've moved the action method?

lfdebrux (approved):
Thanks for looking at this @danworth!

I like where you're going with this, it certainly does feel more predictable. There are some edge cases I can think of, but I think they were edge cases already 😂 I think what you've got has got to be an improvement.

app/controllers/pages_controller.rb
danworth: I think its clearer to me, but there is a change in when the `clear_draft_questions_data` is invoked which I would welcome you're opinion on. 

When it was in the ApplicationController is was being invoked whenever a user visited almost any page (it was excluded in the heartbeat controller). This I believe was causing a problem mentioned above under "second problem". If a user was editing a question and happened to visit any other page in another tab it would clear the draft question they were working on which doesn't seem right and I think causes some of the [draft question exceptions](https://govuk-forms.sentry.io/issues/4200026265/?project=6576190). It makes sense to me that we only clear draft questions, when they save their question, visit the new method (or as it was also on `move_page`, not sure why but that was already there). It's also clearer to me to have this in the pages controller because that's its concern. It was confusing when I looked at it originally in the ApplicationController and wondering why we're clearing draft questions on very unrelated pages, like viewing the MOU for example.

To be honest, I would err towards that we explicitly call `clear_draft_questions_data` from the appropriate controller methods when there are few specific reasons to call it (unlike ensuring authentication which is wide spread and safer to `before_action` as the ApplicationController level). We currently have a mix because we call it explicitly in the update questions (https://github.com/alphagov/forms-admin/blob/main/app/controllers/pages/questions_controller.rb#L48). This to me is simpler to understand what that controller method does, but using the `before_action` with `only` seems the more Railsy way :)

app/controllers/application_controller.rb
```diff
@@ -11,8 +11,6 @@ class ApplicationController < ActionController::Base
 
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
 
-  before_action :clear_draft_questions_data
```
SamJamCul, line 14: I did work on this a bit with @aliuk2012 , this may have been here as a legacy from a time when incomplete questions were stored in the session, so we were a lot more concerned about storing an indefinite amount of stuff.

app/controllers/pages_controller.rb
lfdebrux: Sorry, yes, I meant it's the same within the PagesController class and subclasses. I agree there's not much need to clear the draft question in pages that aren't related to editing a particular form. We could end up with some storage leaking over time, but I don't think it would be that much of an issue (famous last words).

Personally I'm not too bothered either way about using `before_action` or not, but googling for 'ruby rails when not to use before_action' suggests that there a different schools of thought out there...

danworth: yeah I'm happy to leave it with `before_action` if that's the norm. Ok to revolve this conversation?

lfdebrux: Resolved :D

thomasiles (approved):
Nice! This is a bit of a scary change but I don't think there are any situations where clearing the draft question outside of the pages mattered.

There will be differences but I think they are all good. Draft questions will now persist between login sessions, but I think thats better.

I can't find anything thats broken testing it locally.

🏅

PR #1133, opened 2024-05-02
Remove success banners for trial groups

stephencdaly:
We don't want to show success banners for trial groups, as we already display a notification banner when a group is in trial mode and we do not multiple banners. So:

- don't display a banner when a group is created
- only display a banner when the group's name is changed if the group is active.

Updated the content for success banner when the name is changed.

Trello card: https://trello.com/c/KUmN1mEt/1695-update-new-group-page-and-edit-group-name-page-to-match-designs

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -37,7 +37,7 @@ class GroupsController < ApplicationController
     authorize @group
 
     if @group.save
-      redirect_to @group, success: t("groups.success_messages.create")
+      redirect_to @group
     else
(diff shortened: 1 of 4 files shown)


DavidBiddle (approved):

DavidBiddle (approved):

stephencdaly (commented):
Even thought I created the PR, @DavidBiddle picked it up to merge. I then spotted an issue with a factory trait I added which he fixed. I can confirm this is good to merge ✅

PR #1139, opened 2024-05-08
Add group upgrade requests

stephencdaly:

Trello card: https://trello.com/c/j8yP0WRm/1484-request-group-upgrade-add-request-upgrade

When the user is a group admin for the group and a standard user in the organisation (not organisation admin or super admin), there is a link in the trial banner for the group to request an upgrade:

Screenshot 2024-05-08 at 15 25 52

When the link is clicked, a page is displayed with a green button to send the request:

Screenshot 2024-05-08 at 15 26 07

When the button is clicked, the status of the group is updated to a new status `upgrade_requested` and the user is shown a success page:

Screenshot 2024-05-08 at 15 29 02

An email is sent to all organisation admins for the organisation, informing them of the request. This email contains a link to visit the group page:

Screenshot 2024-05-08 at 15 29 41

For the MVP, we do not change the trial group banner for group admins when the request is pending, and group admins are able to submit multiple requests. This might be revisited later.

As part of this PR, we have also updated the content for the trial banner for the different user roles in the organisation and group to match the designs.

When the user is an organisation admin:

Screenshot 2024-05-08 at 15 32 49

When the user is an editor for the group:

Screenshot 2024-05-08 at 15 25 24

### DIFF
--- a/app/mailers/group_upgrade_mailer.rb
+++ b/app/mailers/group_upgrade_mailer.rb
@@ -1,4 +1,4 @@
-class GroupUpgradedMailer < GovukNotifyRails::Mailer
+class GroupUpgradeMailer < GovukNotifyRails::Mailer
   def group_upgraded_email(upgraded_by_user:, to_email:, group:, group_url:)
     set_template(Settings.govuk_notify.group_upgraded_template_id)
 
--- a/app/mailers/group_upgrade_mailer.rb
+++ b/app/mailers/group_upgrade_mailer.rb
@@ -12,4 +12,19 @@ class GroupUpgradeMailer < GovukNotifyRails::Mailer
 
     mail(to: to_email)
   end
+
+  def group_upgrade_requested_email(requester_name:, requester_email_address:, to_email:, group_name:, view_request_url:)
+    set_template(Settings.govuk_notify.group_upgrade_requested_template_id)
+
+    set_email_reply_to(Settings.govuk_notify.zendesk_reply_to_id)
+
+    set_personalisation(
+      requester_name:,
+      requester_email_address:,
+      group_name:,
+      view_request_url:,
+    )
+
+    mail(to: to_email)
+  end
 end
--- a/app/mailers/group_upgrade_mailer.rb
+++ b/app/mailers/group_upgrade_mailer.rb
@@ -1,12 +1,12 @@
 class GroupUpgradeMailer < GovukNotifyRails::Mailer
-  def group_upgraded_email(upgraded_by_user:, to_email:, group:, group_url:)
+  def group_upgraded_email(upgraded_by_name:, to_email:, group_name:, group_url:)
     set_template(Settings.govuk_notify.group_upgraded_template_id)
 
     set_email_reply_to(Settings.govuk_notify.zendesk_reply_to_id)
 
     set_personalisation(
-      upgraded_by_name: upgraded_by_user.name,
-      group_name: group.name,
+      upgraded_by_name:,
+      group_name:,
       group_url:,
     )
 
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -16,4 +16,8 @@ class Organisation < ApplicationRecord
       name
     end
   end
+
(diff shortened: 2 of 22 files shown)


app/models/organisation.rb
```diff
@@ -4,6 +4,8 @@ class Organisation < ApplicationRecord
   has_many :forms
   has_many :users
 
+  has_many :admin_users, -> { where(role: :organisation_admin) }, class_name: "User"
```
jamie-o-wilkinson, line 7: We wondered about whether this should be done in the Organisation or User model. We settled on putting it in the Organisation model as it felt more like canonical Rails. It also avoids adding a class method to the User model, which seemed nicer.

jamie-o-wilkinson (dismissed):

spec/policies/group_policy_spec.rb
```diff
@@ -54,10 +54,6 @@
       it "permits group creation, viewing, and editing" do
         expect(policy).to permit_only_actions(%i[show new edit create update])
       end
-
-      it "does not allow add_editor or upgrade" do
-        expect(policy).to forbid_only_actions(%i[add_editor upgrade])
-      end
```
lfdebrux, line 60: Why was this test removed?

app/mailers/group_upgrade_mailer.rb
```diff
@@ -1,12 +1,12 @@
 class GroupUpgradeMailer < GovukNotifyRails::Mailer
-  def group_upgraded_email(upgraded_by_user:, to_email:, group:, group_url:)
+  def group_upgraded_email(upgraded_by_name:, to_email:, group_name:, group_url:)
```
lfdebrux: What's the reason behind this refactor, is it for a particular reason or following a particular principle/pattern? It would be good to have it in the commit message.

app/models/organisation.rb
lfdebrux: I think this makes sense to me to have the method on the organisation model, my only question is whether `has_many` is the right tool for this job; would a simple method also do the job? As I understand it adding this relation means we also have methods for changing/adding to admin users, which I don't think we need?

app/service/group_service.rb
```diff
@@ -14,26 +14,43 @@ def upgrade_group
...
     @group.upgrade_requested!
+    send_group_upgrade_requested_emails
   end
 
 private
 
   def send_group_upgraded_emails
     @group.memberships.each do |membership|
-      send_group_updated_email(membership.user) if notify_member?(membership)
+      send_group_updated_email(membership.user.email) if notify_member?(membership)
```
lfdebrux, line 24: Should this change have been in the previous commit?

spec/policies/group_policy_spec.rb
stephencdaly: We removed it as there is a test above that tests that only certain actions are permitted, so already tests this. We previously removed the other tests where there was both `permit_only_actions` and `forbid_only_actions` for the same context, so this was the only one left

DavidBiddle (dismissed):
Code looks good, have tried locally and all seems to work, including the emails 🎉

app/service/group_service.rb
stephencdaly: It could be, but not sure its worth the effort moving it. It's related to the changes here enough and is isolated so I don't think it's too difficult to follow?

app/mailers/group_upgrade_mailer.rb
stephencdaly: Yeh just following the convention of not passing around more than we need to. Kelvin did a similar refactor recently so thought it was worthwhile following that principle while we were editing this class. We can add something to the commit message.

app/models/organisation.rb
jamie-o-wilkinson: Good point, switching over to just use a method

DavidBiddle (dismissed):

lfdebrux (dismissed):

DavidBiddle (approved):

PR #1141, opened 2024-05-09
Prevent editors from editing groups

SamJamCul:

https://trello.com/c/5iV1sCvx/1504-prevent-editors-from-changing-the-group-name-and-the-members

This removes the link from a group for changing the name of the group for editor users. 
It also changes the text for the link to the group members page for editors. 
Under the hood, editors have been restricted from the group `edit` and `update` policy, as there weren't any use cases for them.

See comparisons:

### Previous page / current page for non-editors
image

### New version for editor members

image


thomasiles (approved):
nice!

PR #1143, opened 2024-05-09
Make optional question setting use radio buttons

DavidBiddle:

Trello card: https://trello.com/c/WjhNHghp/1482-make-this-question-optional-checkbox-isnt-marked-as-optional

Replaces the 'Make this question optional' checkbox in the question form with a new binary choice radio button so it's clearer to the user what the checkbox will do. As with the checkbox, this radio button option won't appear for selection questions.

Inlcudes a slight reworking of the boolean cast logic in the page model.

When doing this work I noticed that in the forms-api database the value of `is_optional` for a mandatory questionis set to `nil`. This is wrong - it should be `false`. It doesn't break anything at the moment because `nil` is falsey, but we should change it. Once this PR is merged, any mandatory questions that are created or edited will have `is_optional` set to `false`. We'll have to make a migration in forms-api to make the column non-nullable and change the existing `nil` values to `false`.

#### Before
Question settings section with a single checkbox. The checkbox has the label 'Make this question optional so people can leave it blank' and a hint '‘(optional)’ will be added to the end of the question text'

#### After
The same section with a radio button question. The question is 'Should this question be mandatory or optional?', and the two options are 'Mandatory' and 'Optional'. This serves as the section heading. The 'Optional' answer has an additional hint: 'We’ll add ‘(optional)’ to the end of the question text.'

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -443,6 +443,9 @@ en:
           phone_number: Ask the question the way you would in person. For example, ‘What’s your phone number?’
           radio: Ask the question the way you would in person. For example, ‘What country do you live in?’
           single_line: Ask the question the way you would in person. For example, ‘What’s your reference number?’
+      pages_question_input:
+        is_optional_options:
+          'true': We’ll add ‘(optional)’ to the end of the question text.
     label:
       account_name_input:
         name: Enter your full name
@@ -534,6 +537,9 @@ en:
         default_goto_page_id: Select a question or page
       pages_question_input:
         hint_text: Hint text (optional)
+        is_optional_options:
+          'false': Mandatory
+          'true': Optional
         question_text: Question text
       user_upgrade_request:
         met_requirements_options:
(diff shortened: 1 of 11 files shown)


config/locales/en.yml
```diff
@@ -441,6 +443,9 @@ en:
           phone_number: Ask the question the way you would in person. For example, ‘What’s your phone number?’
           radio: Ask the question the way you would in person. For example, ‘What country do you live in?’
           single_line: Ask the question the way you would in person. For example, ‘What’s your reference number?’
+      pages_question_input:
+        is_optional_options:
+          'true': If you make this question optional, users will be able to leave it blank. ‘(optional)’ will be added to the end of the question text.
```
hannahkc, line 448:
```suggestion
          'true': We‘ll add ‘(optional)’ to the end of the question text.
```

thomasiles (approved):
nice! This looks like it covers everything and works when testing locally 🏅 .

hannahkc (commented):
Hi David

Rather than having the question as a sub heading, can we use it as the heading instead of 'Question Settings'?

This was discussed in the trello card, but if it's tricky to do as part of this we can do it separately?

DavidBiddle:
@hannahkc done!

hannahkc (approved):

PR #1144, opened 2024-05-10
Add buttons for changing group member's role

thomasiles:
## Add buttons for changing group member's role

This is a simpler implementation of [this PR](https://github.com/alphagov/forms-admin/pull/1128).

I've removed the service and used code in the view. I've used a new PR to make comparison easier.

image


DavidBiddle (approved):
Code looks good, approach makes sense, and works on my local machine 🎉 

Especially appreciated the clear explanation for the brakeman change in the commit message!

PR #1151, opened 2024-05-13
Add upgrade request review

stephencdaly:

Trello card: https://trello.com/c/UTcd8lwc/1486-request-group-upgrade-add-approve-request

When an upgrade has been requested for a group, display a notification banner for organisation/super admins to review the request:

Screenshot 2024-05-13 at 18 15 57

When the link is clicked, a confirmation page is shown to accept or reject:

Screenshot 2024-05-13 at 18 16 02

If no option is chosen, an error message is shown:

Screenshot 2024-05-13 at 18 16 07

When "Yes" is selected, the group is upgraded to `active` and the user is redirected to the group page with a success banner. The same email that is sent for when a group is upgraded without a request (already implemented) is sent to all group admins.

Screenshot 2024-05-13 at 18 16 14

When "No" is selected, the group status is changed back to `trial`. An email is sent to all group admins to tell them the request was rejected.

Screenshot 2024-05-14 at 15 36 38

### DIFF
--- a/app/policies/group_policy.rb
+++ b/app/policies/group_policy.rb
@@ -5,18 +5,18 @@ class GroupPolicy < ApplicationPolicy
   alias_method :create?, :new?
 
   def show?
-    user.super_admin? || user.is_organisations_admin?(record.organisation) || user.groups.include?(record)
+    organisation_admin_or_super_admin? || user.groups.include?(record)
   end
 
   def edit?
-    user.super_admin? || user.is_organisations_admin?(record.organisation) || group_admin?
+    organisation_admin_or_super_admin? || group_admin?
   end
 
   alias_method :update?, :edit?
   alias_method :add_editor?, :edit?
 
   def upgrade?
-    user.super_admin? || user.is_organisations_admin?(record.organisation)
+    organisation_admin_or_super_admin?
   end
 
   alias_method :add_group_admin?, :upgrade?
@@ -25,7 +25,15 @@ class GroupPolicy < ApplicationPolicy
     group_admin? && !record.active?
   end
 
+  def review_upgrade?
+    organisation_admin_or_super_admin? && record.upgrade_requested?
+  end
+
 private
+
+  def organisation_admin_or_super_admin?
+    user.super_admin? || user.is_organisations_admin?(record.organisation)
+  end
 
   def group_admin?
     record.memberships.find_by(user:)&.group_admin?
--- a/app/service/group_service.rb
+++ b/app/service/group_service.rb
@@ -12,6 +12,11 @@ class GroupService
     send_group_upgraded_emails
   end
 
+  def reject_upgrade
+    @group.trial!
+    send_upgrade_rejected_emails
+  end
+
   def request_upgrade
     @group.upgrade_requester = @current_user
     @group.upgrade_requested!
@@ -22,13 +27,19 @@ private
 
   def send_group_upgraded_emails
     @group.memberships.each do |membership|
-      send_group_updated_email(membership.user.email) if notify_member?(membership)
+      send_group_upgraded_email(membership.user.email) if notify_member?(membership)
+    end
+  end
+
+  def send_upgrade_rejected_emails
+    @group.memberships.each do |membership|
+      send_upgrade_rejected_email(membership.user.email) if notify_member?(membership)
     end
   end
 
   def send_group_upgrade_requested_emails
     @group.organisation.admin_users.each do |user|
-      send_group_update_requested_email(user.email)
+      send_group_upgrade_requested_email(user.email)
     end
   end
 
@@ -36,7 +47,7 @@ private
     membership.group_admin? && membership.user.id != @current_user.id
   end
 
-  def send_group_updated_email(to_email)
+  def send_group_upgraded_email(to_email)
     GroupUpgradeMailer.upgraded_email(
       to_email:,
       upgraded_by_name: @current_user.name,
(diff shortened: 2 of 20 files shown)


app/policies/group_policy.rb
```diff
@@ -5,26 +5,34 @@ def new?
...
-    user.super_admin? || user.is_organisations_admin?(record.organisation)
+    organisation_admin_or_super_admin?
   end
 
   def request_upgrade?
     group_admin? && !record.active?
   end
 
+  def review_upgrade?
+    organisation_admin_or_super_admin? && record.upgrade_requested?
```
stephencdaly, line 27: For convenience, we are checking that the group is in the correct status for an upgrade request to be reviewed as part of the policy. This allows us to re-use the policy in the show group view to decide which trial group notification banner to display.

This means that if an organisation admin attempts to visit the page to review an upgrade request when the request has already been approved or rejected, they will get a 403 error and see a page with the message "You do not have permission to view this page."

Whilst it might be more ideal to show a more specific error, or a 404 in this scenario this is very much an edge case. For the MVP, we decided not to customise these error pages in the designs but can revisit this later.

Any thoughts on this welcomed.

spec/features/groups/request_upgrade_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    click_link_or_button "Send request to upgrade"
+    expect(page.find("h1")).to have_text "Your upgrade request has been sent"
+    expect_page_to_have_no_axe_errors(page)
+
+    login_as_organisation_admin_user
+
+    visit groups_path
+    expect(page).to have_css ".govuk-notification-banner", text: "You have one request to upgrade a trial group."
+    expect_page_to_have_no_axe_errors(page)
```
stephencdaly, line 34: On most other feature specs we follow the pattern of having private methods that have names describing each step of the test.

For this test in particular, we would end up with names like `i_visit_the_groups_page_as_an_organisation_admin` as we have to repeat the same actions as different users. We thought this made it harder follow the test with no benefit of reducing duplication so didn't follow this pattern.

Thoughts welcome on this.

lfdebrux (commented):
This looks good to me on a first reading. There's some nice refactors in here!

spec/features/groups/request_upgrade_spec.rb
thomasiles: I think thats sensible 👍

thomasiles (approved):
looks really good and works locally. I've left a question but I don't think its important.

great work!

app/service/group_service.rb
```diff
@@ -12,7 +12,13 @@ def upgrade_group
     send_group_upgraded_emails
   end
 
+  def reject_upgrade
+    @group.trial!
+    send_upgrade_rejected_emails
+  end
+
```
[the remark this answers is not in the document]
stephencdaly: Me and @jamie-o-wilkinson did discuss this and landed on not clearing it. I can't exactly remember our reasoning but I think we thought it didn't matter too much either way. As you say, it won't be used anywhere afterwards - you're not able to visit the page that would show the requester name if the group status isn't `upgrade_requested`.

 If we think it makes more sense to clear it, we can do that.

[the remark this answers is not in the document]
thomasiles: I think that makes sense 👍

app/policies/group_policy.rb
DavidBiddle: This feels OK for now - I think we've done similar things with policies in the past

spec/features/groups/request_upgrade_spec.rb
DavidBiddle: I think another option would be to have generic methods (e.g. `i_visit_the_groups_page`) which do the page visiting and the axe check, but keep the bits that change between roles outside of the method.

```ruby
def i_visit_the_groups_page
  visit groups_path
  expect_page_to_have_no_axe_errors(page)
end
```

and in the scenario:
```ruby
    login_as_editor_user
    i_visit_the_groups_page
    expect(page.find("h1")).to have_text "Your groups"
    
    login_as_organisation_admin_user
    i_visit_the_groups_page
    expect(page).to have_css ".govuk-notification-banner", text: "You have one request to upgrade a trial group."
```

Additionally I think you could also try putting the role-specific assertions into their own methods - wouldn't do anything to reduce duplication, but would potentially make the scenarios easier to read:

```ruby
def i_should_see_the_your_groups_text
  expect(page.find("h1")).to have_text "Your groups"
end

def i_should_see_the_upgrade_request_banner
  expect(page.find("h1")).to have_text "Your groups"
end
```

so the scenario would look like this:
```ruby
    login_as_editor_user
    i_visit_the_groups_page
    i_should_see_the_your_groups_text    

    login_as_organisation_admin_user
    i_visit_the_groups_page
    i_should_see_the_upgrade_request_banner
```

DavidBiddle: I'll leave it up to you whether you think that's worthwhile - I don't think this spec is particularly difficult to read anyway

spec/views/groups/show.html.erb_spec.rb
```diff
@@ -7,14 +7,20 @@
...
     assign(:group, group)
     assign(:forms, forms)
 
     without_partial_double_verification do
-      allow(view).to receive(:policy).and_return(instance_double(GroupPolicy, upgrade?: upgrade?, edit?: edit?, request_upgrade?: request_upgrade?))
+      double = instance_double(GroupPolicy,
+                               upgrade?: upgrade?,
+                               edit?: edit?,
+                               request_upgrade?: request_upgrade?,
+                               review_upgrade?: review_upgrade?)
```
DavidBiddle, line 22: Does the shorthand work here?
```suggestion
                               upgrade?:,
                               edit?:,
                               request_upgrade?:,
                               review_upgrade?:)
```

DavidBiddle (approved):
Added a couple of minor comments but looks really good overall and works for me locally 🎉

spec/views/groups/show.html.erb_spec.rb
stephencdaly: Unfortunately not in this case, I think it's because of the question marks 🤔

spec/features/groups/request_upgrade_spec.rb
stephencdaly: Thanks for the suggestions. What you've suggested makes sense to me. In absence of a strong opinion, I think we'll leave it as it is because as you say it's not particularly difficult to read.

I've put in a topic in the dev meeting next week about this just to see if anyone has any strong preferences or good reasoning for using this style of having methods in the tests.

PR #1155, opened 2024-05-14
Add new group member role

thomasiles:
### Add members to groups with role

Allow super admins and organisation admins to add members to a group as either editors or group admins.

image

### DIFF
--- a/app/policies/group_policy.rb
+++ b/app/policies/group_policy.rb
@@ -19,6 +19,8 @@ class GroupPolicy < ApplicationPolicy
     user.super_admin? || user.is_organisations_admin?(record.organisation)
   end
 
+  alias_method :add_group_admin?, :upgrade?
+
   def request_upgrade?
     group_admin? && !record.active?
   end
--- a/app/views/group_members/new.html.erb
+++ b/app/views/group_members/new.html.erb
@@ -1,4 +1,8 @@
-<% set_page_title(t(".title")) %>
+<% if show_role_options %>
+  <% set_page_title(t(".title_show_role_options")) %>
+<% else %>
+  <% set_page_title(t(".title")) %>
+<% end %>
 <% content_for :back_link, govuk_back_link_to(group_members_path(@group)) %>
 
 <div class="govuk-grid-row">
@@ -6,7 +10,11 @@
     <h1 id="heading" class="govuk-heading-l">
       <span class="govuk-caption-l"><%= @group.name %></span>
       <span class="govuk-visually-hidden"> - </span>
-      <%= t(".title") %>
+      <% if show_role_options %>
+        <%= t(".title_show_role_options") %>
+      <% else %>
+        <%= t(".title") %>
+      <% end %>
     </h1>
 
     <%= t(".body_html") %>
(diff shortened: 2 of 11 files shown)


app/policies/group_policy.rb
```diff
@@ -15,6 +15,10 @@ def edit?
   alias_method :update?, :edit?
   alias_method :add_editor?, :edit?
 
+  def add_group_admin?
+    user.super_admin? || user.is_organisations_admin?(record.organisation)
+  end
```
stephencdaly, line 20: Is this one we can add as an alias, as `upgrade?` has the same permissions check?

spec/policies/group_policy_spec.rb
```diff
@@ -66,7 +66,7 @@
       end
 
       it "forbids upgrade" do
```
stephencdaly, line 68: Should we update the name of this test to reflect the updated behaviour?

config/locales/en.yml
```diff
@@ -466,6 +468,14 @@ en:
         confirm: Are you sure you want to make your form live?
       group_member_input:
         member_email_address: Enter the email address of the person you want to add
+        role: Select the role for this person
```
stephencdaly, line 472: The content here doesn't match the content in the Figma - which one is correct?

spec/input_objects/group_member_input_spec.rb
```diff
@@ -7,8 +7,9 @@
   let(:user) { create(:user, organisation: group.organisation) }
 
   describe "validations" do
-    it "is valid with an email address" do
+    it "is valid with an email address and role" do
```
stephencdaly, line 10: Should we add another test to check there is an error if the role is missing?

stephencdaly (commented):
Changes look good and I've tested locally. I've just left a few minor comments.

Also, should the error summary appear at the top of the page?

Screenshot 2024-05-14 at 15 53 16

config/locales/en.yml
stephencdaly: This is what I'm looking at in the Figma, for reference

Screenshot 2024-05-14 at 15 57 11

thomasiles: totally right! the figma was correct and I've updated it

app/views/group_members/new.html.erb
```diff
@@ -1,22 +1,41 @@
...
+<% if show_role_options %>
+  <% set_page_title(t(".title_show_role_options")) %>
+<% else %>
+  <% set_page_title(t(".title")) %>
+<% end %>
 <% content_for :back_link, govuk_back_link_to(group_members_path(@group)) %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 id="heading" class="govuk-heading-l">
```
stephencdaly, line 10: The error summary should come above the page heading. On other pages, we put the `h1` inside the `form` to achieve this.

Screenshot 2024-05-16 at 11 02 54

thomasiles: Thanks! sorry wasn't thinking properly, I've updated it.

stephencdaly (approved):
All looks good to me!

thomasiles:
> All looks good to me!

Thanks for your patience and careful reviewing! 🙏

PR #1156, opened 2024-05-15
Refactor pundit stubs in request specs

thomasiles:
### Remove Pundit stubs in request specs

A recent update to Pundit, the gem we use for authorisation, means our request tests no longer pass.

This is because the way we were stubbing calls to Pundit no longer works. See this [dependabot PR](https://github.com/alphagov/forms-admin/pull/1140) for more details.

After [some discussion with the technical team](https://docs.google.com/document/d/1iHfUoCzgg-2Mk9wbi5o46fFVRak_R6nNQ2m8OqSz_hs/edit#heading=h.c0rqps9ls6o) we decided that request specs should not stub Pundit but arrange the tests so that the authorisation is tested and passes the tests too.

This PR works through the request specs and removes the stubs to Pundit. Instead the setup has been modified so that the tests pass. Some Pundit stubs were not doing anything and so could be removed without changing the test behaviour.

This PR doesn't do the Pundit update. Rebasing the dependabot PR when this is merged should allow them to pass.

There are other Pundit stubs within our code which don't need to change. This are stubbing the policy object within views, which the Pundit update doesn't break.

### DIFF
--- a/spec/requests/forms/change_name_controller_spec.rb
+++ b/spec/requests/forms/change_name_controller_spec.rb
@@ -20,8 +20,6 @@ RSpec.describe Forms::ChangeNameController, type: :request do
       mock.put "/api/v1/forms/2", post_headers
     end
 
-    allow(Pundit).to receive(:authorize).and_return(true)
-
     login_as user
   end
 
--- a/spec/requests/memberships_controller_spec.rb
+++ b/spec/requests/memberships_controller_spec.rb
@@ -38,9 +38,10 @@ describe "/memberships", type: :request do
 
   describe "PUT #update" do
     let(:membership) { create(:membership, user:, group:) }
+    let(:current_user) { organisation_admin_user }
 
     before do
-      login_as_super_admin_user
+      login_as current_user
     end
 
     context "with valid parameters" do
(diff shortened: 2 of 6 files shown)


spec/requests/forms/change_name_controller_spec.rb
```diff
@@ -20,8 +20,6 @@
       mock.put "/api/v1/forms/2", post_headers
     end
 
-    allow(Pundit).to receive(:authorize).and_return(true)
-
```
lfdebrux, line 24: Nice

spec/requests/memberships_controller_spec.rb
```diff
@@ -38,9 +38,10 @@
 
   describe "PUT #update" do
     let(:membership) { create(:membership, user:, group:) }
+    let(:logged_in_user) { organisation_admin_user }
```
lfdebrux, line 41: I think we call this `current_user` in other tests, might be worth using the same here for consistency?

lfdebrux (approved):
This is a big improvement, thanks for doing this. I have one small comment but happy either way.

spec/requests/memberships_controller_spec.rb
thomasiles: yes, logged_in_user is a clunky name, I'll change it.

jamie-o-wilkinson (approved):
I think this is actually much nicer than stubbing out Pundit. Thank you! 🥇

spec/requests/memberships_controller_spec.rb
thomasiles: I've changed it 👍

lfdebrux (approved):

PR #1164, opened 2024-05-17
Add feature flag for groups

lfdebrux:

Put the groups feature behind a feature flag. When the feature flag is disabled:
- prevent access to all groups routes
- back links from forms will link to the root URL rather than the group URL, regardless of whether the form is in a group
- policies for forms will not look at the user's role in a group

In this PR, we have not yet made the feature flag configurable per organisation. This will come in a follow-up PR.

### DIFF
--- a/app/controllers/forms/delete_confirmation_controller.rb
+++ b/app/controllers/forms/delete_confirmation_controller.rb
@@ -62,7 +62,7 @@ module Forms
     end
 
     def delete_form(form)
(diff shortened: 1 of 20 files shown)


thomasiles (dismissed):
LGTM - works well locally and it looks like all the cases for back links etc are covered.

jamie-o-wilkinson (dismissed):
Works for me locally. Nice one! 🥇

stephencdaly (approved):

PR #1169, opened 2024-05-20
Add validation for org admin orgs

SamJamCul:
Adds a validation to check that organisation_admins are part of an organisation.

We're not sure if there are any existing org admins that don't have an org, so would want to confirm that there aren't before adding this validation

We're also not sure if there are any better ways of using factories to build valid organisation_admin users, as it's a bit fiddly signing an mou for their org so that they're created in a valid state

Trello card: https://trello.com/c/wgcwhH8p/1527-prevent-a-user-from-being-made-an-org-admin-if-their-organisation-hasnt-agreed-to-an-mou

image

### DIFF
--- a/spec/factories/models/organisations.rb
+++ b/spec/factories/models/organisations.rb
@@ -8,5 +8,11 @@ FactoryBot.define do
     initialize_with do
       Organisation.create_with(govuk_content_id:, name:).find_or_initialize_by(slug:)
     end
+
+    trait :with_signed_mou do
+      after(:build) do |organisation|
+        organisation.mou_signatures << (create :mou_signature_for_organisation, organisation:)
+      end
+    end
   end
 end
--- a/spec/requests/users_controller_spec.rb
+++ b/spec/requests/users_controller_spec.rb
@@ -40,9 +40,9 @@ RSpec.describe UsersController, type: :request do
         login_as_super_admin_user
 
         organisations = [
-          create(:organisation, slug: "test-org"),
-          create(:organisation, slug: "ministry-of-tests"),
-          create(:organisation, slug: "department-for-testing"),
+          create(:organisation, :with_signed_mou, slug: "test-org"),
+          create(:organisation, :with_signed_mou, slug: "ministry-of-tests"),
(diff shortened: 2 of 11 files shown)


hannahkc:
Content looks bang on thank you!

spec/requests/users_controller_spec.rb
```diff
@@ -253,7 +253,7 @@
 
       User.roles.reject { |role| role == "trial" }.each_value do |role_value|
```
chao-xian, line 254: Please make this consistent with the other file where we are more explicit with our specs.

spec/factories/models/organisations.rb
```diff
@@ -8,5 +8,11 @@
     initialize_with do
       Organisation.create_with(govuk_content_id:, name:).find_or_initialize_by(slug:)
     end
+
+    trait :with_signed_mou do
+      after(:build) do |organisation|
+        organisation.mou_signatures << (create :mou_signature_for_organisation, organisation:)
+      end
+    end
```
stephencdaly, line 16: I don't see any major issues with this. But if there's likely to only be a few specs that depend on an MOU existing for the user/organisation, is it clearer to just build the `mou_signature` in the spec rather than having it as a trait?

SamJamCul: It's mainly to clean up tests so that they're more readable. In this case, we added a trait to the user factory which is bundled along with creating an `organisation_admin` user, and that triggers a sequence of factory traits to make sure that user is in a valid state. Doing all of that piece by piece for tests that involve an organisation admin would just get repetitive, and wouldn't result in additional coverage.

spec/requests/users_controller_spec.rb
SamJamCul: This has been changed so that we aren't looping through the roles, instead we're testing specific roles and actions in each test.

stephencdaly (approved):
Looks good to me! There aren't any users that are organisation admins on production yet, so should be fine on that one unless we start adding ones that haven't agreed to an MOU

PR #1170, opened 2024-05-20
Add rake task to update the organisation for an MOU signature

stephencdaly:

Trello card:
https://trello.com/c/syeAKkzY/1535-manually-change-mou-agreement-from-dhsc-to-ohid
https://trello.com/c/vJsmEFPl/1534-manually-change-mou-agreement-from-home-office-to-cce

Adds a rake task that, given the User name and Organisation name for an MOU, will update the organisation for the MOU to the target organisation.

The primary intended use for this is when a user and their forms are moved to a sub-organisation and the MOU signature needs to be moved with them to the sub-organisation to reflect the organisation that the user agreed to the MOU for.

### DIFF
--- a/spec/lib/tasks/mou_signatures.rake_spec.rb
+++ b/spec/lib/tasks/mou_signatures.rake_spec.rb
@@ -1,0 +1,58 @@
+require "rake"
+
+require "rails_helper"
+
+RSpec.describe "mou_signatures.rake" do
+  before do
+    Rake.application.rake_require "tasks/mou_signatures"
+    Rake::Task.define_task(:environment)
+  end
+
+  describe "mou_signatures:update_organisation" do
+    subject(:task) do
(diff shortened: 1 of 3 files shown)


spec/lib/tasks/mou_signatures.rake_spec.rb
```diff
@@ -0,0 +1,60 @@
+require "rake"
+
+require "rails_helper"
+
+RSpec.describe "mou_signatures.rake" do
+  before do
+    # Rake.application.options.trace = true
```
chao-xian, line 7: Don't think you meant to leave this in?

stephencdaly: I copied this from the other rake task test. Have removed here and in the file I copied from too.

chao-xian (approved):

PR #1172, opened 2024-05-21
Fix record not found for groups routes

lfdebrux:

Requests to any URLs with an invalid group_id (such as `/groups/foobar`) currently result in a 500 error instead of a 404 error.

The 500 is caused when the method `set_group` assigns the instance variable `@group` is assigned to `nil`, leading Pundit to explode with the error:

```
unable to find policy `NilClassPolicy` for `nil`
```

This bug was introduced multiple times, whenever we used the Rails controller scaffold generator. We thought that the files created by the `rails generate scaffold` tool would follow best practice, however by failing to use `find!` (note the exclamation mark) in the template the generated controller would not raise a 404 as expected. This bug was then propagated by us and not caught even as we modified the code.

We also weren't testing this unhappy path because the template from the RSpec generator only covers the happy paths, and we missed this at review time.

This bug may be present in other places in our codebase, but for now I'm just focusing on places related to the groups feature.

[1]: https://github.com/rails/rails/blob/747a03ba7722b6f0a7ce42e86cea83cf07a2e6ef/railties/lib/rails/generators/rails/scaffold_controller/templates/controller.rb.tt#L52
[2]: https://github.com/rspec/rspec-rails/blob/4ab08fa954fe4e0ae6f2302fa7201dcad3f11521/lib/generators/rspec/scaffold/templates/request_spec.rb

### DIFF
--- a/app/controllers/group_forms_controller.rb
+++ b/app/controllers/group_forms_controller.rb
@@ -31,7 +31,7 @@ class GroupFormsController < ApplicationController
 private
 
   def set_group
-    @group = Group.find_by(external_id: params[:group_id])
+    @group = Group.find_by!(external_id: params[:group_id])
   end
 
   def name_input_params(form)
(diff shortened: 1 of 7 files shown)


thomasiles (approved):
good catch 👍

PR #1174, opened 2024-05-21
Update super admin hint text in en.yml

hannahkc:

Edited the hint for the 'super admin' radio option in the Edit user page to make it clear that only members of the GOV⁠.⁠UK Forms team should be made super admins. 

Also simplified the language of the hints for the 'permitted' and 'denied' radio options.

- Does it look like Hannah broke anything? 
- Are there any typos? 
- Are there any straight quotes?


danworth (approved):
Looks good 🏅

PR #1176, opened 2024-05-22
Update root path to depend on groups feature flag

stephencdaly:

Add a new HomepageController with an index action that is called when the root path is visited.

If the groups feature flag is enabled, redirect to the groups index page.

If the groups feature flag is disabled, directly call the Forms controller index action. We're doing this rather than redirecting because we don't want to add a new route for the Forms controller index action, which will be removed soon anyway.

### DIFF
--- a/app/controllers/homepage_controller.rb
+++ b/app/controllers/homepage_controller.rb
@@ -1,0 +1,9 @@
+class HomepageController < ApplicationController
+  def index
+    if FeatureService.enabled? :groups
+      redirect_to groups_path
+    else
+      FormsController.dispatch(:index, request, response)
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


app/controllers/homepage_controller.rb
```diff
@@ -0,0 +1,9 @@
+class HomepageController < ApplicationController
+  def index
+    if FeatureService.enabled? :groups
+      redirect_to groups_path
+    else
+      FormsController.dispatch(:index, request, response)
```
SamJamCul, line 6: Nifty!

SamJamCul (approved):

thomasiles (approved):

PR #1177, opened 2024-05-22
Add task to insert MOU signature

lfdebrux:

We have a couple of cases where an MOU signature has been signed off the platform, and we want to add them in now.

This commit adds a task to help a dev on support do this.

### DIFF
--- a/lib/tasks/mou_signatures.rake
+++ b/lib/tasks/mou_signatures.rake
@@ -1,7 +1,21 @@
 namespace :mou_signatures do
+  desc "Add MOU signature that was done offline"
+  task :create, %i[user_email organisation_name agreed_at_date] => :environment do |_, args|
+    usage_message = "usage: rake mou_signatures:create[<user_email>, <organisation_name>, <agreed_at_date>]".freeze
+    abort usage_message if args[:user_email].blank? || args[:organisation_name].blank? || args[:agreed_at_date].blank?
+
+    user = User.find_by!(email: args[:user_email])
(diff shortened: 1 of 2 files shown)


lib/tasks/mou_signatures.rake
```diff
@@ -1,7 +1,21 @@
 namespace :mou_signatures do
+  desc "Add MOU signature that was done offline"
+  task :create, %i[user_email organisation_name agreed_at_date] => :environment do |_, args|
```
DavidBiddle, line 3: Is there a reason we're using the user's name in one task and their email in the other? Feels like it might make sense to use the same in both tasks. 

I'd prefer using email for both (since we treat them as unique) but I'm open to using name.

stephencdaly: We decided we needed to use email as it has to be unique - you could end up with 2 users with the same name. I didn't think about this when I was adding the first task for updating an MOU. Agree that the other one could be updated to use the email.

I could do this as I wrote the first task. It could be a separate PR if everything else in this one looks fine?

DavidBiddle: Makes sense, thanks Stephen! Think it's fine to be a separate PR, so I'll approve this one. 

I'm happy to pick up the work on the update task if you've got other stuff on today!

DavidBiddle (approved):

DavidBiddle:
Code looks good and have tested locally 🎉

PR #1179, opened 2024-05-22
Add ability to enable groups feature flag for individual organisations

stephencdaly:

Allow specifying overrides for a feature flag per organisation by providing an object in the settings YAML file as the value for the feature, with optional keys matching the slugs for individual organisations (with underscores instead of dashes). If a key exists for the organisation for the logged in user, the value of this key is used over the value of the `enabled` key.

Example:

```
features:
  groups:
    enabled: false
    organisations:
      government_digital_service: true
```

Update where we check the groups feature flag to pass in the logged in user to the FeatureService so we check whether the feature flag has an override for the user's organisation.

### DIFF
--- a/spec/service/feature_service_spec.rb
+++ b/spec/service/feature_service_spec.rb
@@ -1,52 +1,170 @@
 require "rails_helper"
 
 describe FeatureService do
-  describe ".enabled?" do
-    context "when eature is enabled" do
+  describe "#enabled?" do
+    subject :feature_service do
+      described_class.new(user)
+    end
+
+    let(:organisation) { build :organisation, id: 1, slug: "a-test-org" }
+    let(:user) { build :user, id: 1, organisation: }
+
+    context "when the feature key has a boolean value" do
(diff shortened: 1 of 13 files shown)


spec/service/feature_service_spec.rb
```diff
@@ -1,52 +1,170 @@
...
+      end
+
+      context "when a key exists for the organisation overriding the feature and the user has not been provided to the service" do
+        let(:feature_service) { described_class.new(nil) }
+
+        before do
+          Settings.features[:some_feature] = Config::Options.new(enabled: false, organisations: { a_test_org: true })
+        end
+
+        it "is enabled" do
```
jamie-o-wilkinson, line 112: Should this be something to effect of:
```ruby
it "raises an error about a missing user"
```
?

jamie-o-wilkinson (commented):
One suggestion for test names (possible typo?) but otherwise looks good.

jamie-o-wilkinson (approved):

PR #1181, opened 2024-05-23
Restrict upgrade routes

thomasiles:

Trello card: https://trello.com/c/dvoXCPwg/1538-stop-group-upgrades-when-an-organisation-doesnt-have-an-mou-org-admins

Forbids group upgrades when a group's organisation has not signed an MOU
Forbids group admins from requesting an upgrade when a group's organisation does not have any organisation admins.

Also added some extra content for when a super admin views a trial group that can't be upgraded.

image

### DIFF
--- a/app/views/groups/show.html.erb
+++ b/app/views/groups/show.html.erb
@@ -13,6 +13,9 @@
           <% nb.with_heading(text: t("groups.show.trial_banner.heading"), tag: "h3") %>
           <%= t("groups.show.trial_banner.upgrade.body_html") %>
           <%= govuk_link_to t("groups.show.trial_banner.upgrade.link"), upgrade_group_path(@group) %>
+        <% elsif @current_user.super_admin? %>
+          <% nb.with_heading(text: t("groups.index.no_mou_banner.heading"), tag: "h3") %>
+          <%= t("groups.index.no_mou_banner.body_html", contact_link: contact_link) %>
         <% elsif policy(@group).request_upgrade? %>
           <% nb.with_heading(text: t("groups.show.trial_banner.heading"), tag: "h3") %>
           <%= t("groups.show.trial_banner.request_upgrade.body_html") %>
--- a/app/views/groups/show.html.erb
+++ b/app/views/groups/show.html.erb
@@ -5,28 +5,35 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <%= govuk_notification_banner(title_text: t("banner.default.title")) do |nb| %>
-        <% if policy(@group).review_upgrade? %>
-          <% nb.with_heading(text: t("groups.show.trial_banner.review_upgrade.heading"), tag: "h3") %>
-          <%= t("groups.show.trial_banner.review_upgrade.body_html", upgrade_requester_name: @group.upgrade_requester.name) %>
-          <%= govuk_link_to t("groups.show.trial_banner.review_upgrade.link"), review_upgrade_group_path(@group) %>
-        <% elsif policy(@group).upgrade? %>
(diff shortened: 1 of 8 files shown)


spec/features/groups/request_upgrade_spec.rb
stephencdaly, line 1: Should this commit be squashed into the commit that breaks the tests? Do we follow the principle that a commit shouldn't break tests? If not then ignore me.

app/views/groups/show.html.erb
```diff
@@ -13,6 +13,9 @@
           <% nb.with_heading(text: t("groups.show.trial_banner.heading"), tag: "h3") %>
           <%= t("groups.show.trial_banner.upgrade.body_html") %>
           <%= govuk_link_to t("groups.show.trial_banner.upgrade.link"), upgrade_group_path(@group) %>
+        <% elsif @current_user.super_admin? %>
+          <% nb.with_heading(text: t("groups.index.no_mou_banner.heading"), tag: "h3") %>
+          <%= t("groups.index.no_mou_banner.body_html", contact_link: contact_link) %>
```
stephencdaly, line 18: Why do only super admins see the paragraph about the MOU? Should all user types see this if there is no MOU signed?

stephencdaly (commented):
I won't withhold approval due to my comment about traits, just something to consider/discuss.

I do have a question about the content in the banner that the different users see though.

app/views/groups/show.html.erb
SamJamCul: Super admins are the only ones who will need to be shown that they can't upgrade the group solely because the organisation doesn't have an MOU. 

The prior `elsif` checks to see if the user can upgrade the group, but that will be false if there's no MOU - which means we need to still catch super admins here, as there can't be any organisation admins if there's no MOU.

We were getting a bit concerned with the readability for this particular if-chain...

spec/features/groups/request_upgrade_spec.rb
SamJamCul: I don't think we've followed that principle, but it'll be easy enough to reorder these so that they're non-breaky

app/views/groups/show.html.erb
stephencdaly: Aha yeh ok that makes sense! Didn't consider the fact that you can't have org admins without an MOU. Yeh this if chain is getting quite difficult to follow!

The thing that's slightly confusing is that on the Figma there are 2 different notification banners. There's one for "If their org has no MOU or org admin yet", and one for "If their org has no org admin". Are we definitely using the right one for group admins?

stephencdaly: If we should always show the MOU version of the banner if there is no MOU signed, would it be clearer to have a check in the first "if" case that an MOU is signed and if not use the `no_mou_banner` content? Would that make this easier to follow?

stephencdaly (approved):
Looks good to me!

spec/features/groups/request_upgrade_spec.rb
```diff
@@ -2,8 +2,9 @@
 
 feature "Request an upgrade for a group", type: :feature, feature_groups: true do
   let!(:group) do
-    create(:group, organisation: editor_user.organisation).tap do |group|
+    create(:group, :org_has_org_admin, organisation: editor_user.organisation).tap do |group|
```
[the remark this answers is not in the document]
SamJamCul: I think I agree with you here, but I think this is all tied up in our somewhat inexplicable use of the `test-org` in a lot of our tests. There's a ticket [here](https://trello.com/c/dFI2PPaq/1835-review-and-refactor-the-way-we-use-factorybot-in-the-tests-for-the-admin) which is meant to represent us getting that sorted out.

PR #1182, opened 2024-05-24
Restrict making group forms live

jamie-o-wilkinson:
This PR prevent forms within groups from being made live if the group is not active, or if the user doesn't have permission to administer the group (group, org, and super admins). 

It also includes changes to the task list, to indicate to the user why the form cannot be made live, and point them in the direction of a group admin or the group page as appropriate.

Trello card: https://trello.com/c/6nLLXGzi

#### Inactive group, editor user
image

#### Inactive group, admin users
image

For group admins, the link will go directly to the request upgrade page. For org and super admins, it'll go to the group page.

#### Active group, editor user
image

#### Active group, admin user
image

### DIFF
--- a/app/policies/form_policy.rb
+++ b/app/policies/form_policy.rb
@@ -66,6 +66,10 @@ class FormPolicy
     false
   end
 
+  def can_administer_group?
+    user.super_admin? || user.is_organisations_admin?(form.group.organisation) || user.is_group_admin?(form.group)
+  end
+
 private
 
   def user_is_form_creator
@@ -74,9 +78,5 @@ private
 
   def users_organisation_owns_form
     user.organisation.present? ? user.organisation_id == form.organisation_id : false
-  end
-
-  def can_administer_group?
-    user.super_admin? || user.is_organisations_admin?(form.group.organisation) || user.is_group_admin?(form.group)
   end
 end
--- a/app/policies/form_policy.rb
+++ b/app/policies/form_policy.rb
@@ -58,7 +58,13 @@ class FormPolicy
     form_has_two_or_more_pages && form_has_qualifying_pages
   end
 
-  alias_method :can_make_form_live?, :can_change_form_submission_email?
+  def can_make_form_live?
+    # TODO: we should remove the check the form is within a group when we remove the feature flag
+    return can_change_form_submission_email? unless FeatureService.new(user).enabled?(:groups) && form.group.present?
+    return can_change_form_submission_email? if form.group.active? && can_administer_group?
+
+    false
+  end
 
 private
 
@@ -68,5 +74,9 @@ private
 
   def users_organisation_owns_form
     user.organisation.present? ? user.organisation_id == form.organisation_id : false
+  end
+
+  def can_administer_group?
+    user.super_admin? || user.is_organisations_admin?(form.group.organisation) || user.is_group_admin?(form.group)
   end
 end
--- a/app/service/form_task_list_service.rb
+++ b/app/service/form_task_list_service.rb
@@ -124,10 +124,30 @@ private
     if Pundit.policy(@current_user, @form).can_make_form_live?
       section[:rows] = make_form_live_section_tasks
     else
-      section[:body_text] = I18n.t("forms.task_list_create.make_form_live_section.if_not_permitted.body_text")
+      section[:body_text] = make_form_live_section_body_text
     end
 
     section
+  end
+
+  def make_form_live_section_body_text
+    # TODO: we should remove the check the form is within a group when we remove the feature flag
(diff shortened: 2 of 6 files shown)


app/policies/form_policy.rb
```diff
@@ -58,7 +58,15 @@ def can_add_page_routing_conditions?
     form_has_two_or_more_pages && form_has_qualifying_pages
   end
 
-  alias_method :can_make_form_live?, :can_change_form_submission_email?
+  def can_make_form_live?
+    return can_change_form_submission_email? unless FeatureService.enabled?(:groups)
+
+    if form.group.active? && can_administer_group?
```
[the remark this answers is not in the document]
DavidBiddle: Definitely feels better than a ternary, other option would be doing the guard clause thing again but I'm not sure how necessary that is here

app/policies/form_policy.rb
```diff
@@ -58,7 +58,19 @@ def can_add_page_routing_conditions?
...
+      can_change_form_submission_email?
+    else
+      false
+    end
+  end
```
DavidBiddle, line 69: I think this would be the guard clausey version:
```suggestion
  def can_make_form_live?
    return can_change_form_submission_email? unless FeatureService.enabled?(:groups)
    return can_change_form_submission_email? if form.group.active? && can_administer_group?
    
    false
  end
```

app/service/form_task_list_service.rb
```diff
@@ -124,12 +124,26 @@ def make_form_live_section
...
   end
 
+  def make_form_live_section_body_text
+    return I18n.t("forms.task_list_create.make_form_live_section.if_not_permitted.body_text") unless FeatureService.enabled?(:groups)
+    return I18n.t("forms.task_list_create.make_form_live_section.group_not_active.group_admin.body_text", upgrade_path: group_upgrade_url) if !@form.group.active? && Pundit.policy(@current_user, @form).can_administer_group?
+    return I18n.t("forms.task_list_create.make_form_live_section.group_not_active.group_editor.body_text", group_members_path: group_members_path(@form.group)) unless @form.group.active?
+
+    I18n.t("forms.task_list_create.make_form_live_section.user_cannot_administer.body_text", group_members_path: group_members_path(@form.group)) unless Pundit.policy(@current_user, @form).can_administer_group?
+  end
+
```
DavidBiddle, line 140: We could put the two different @form.group.active? cases into a method - do you reckon this makes it any easier to read? Feel free to ignore if not
```suggestion
  def make_form_live_section_body_text
    return I18n.t("forms.task_list_create.make_form_live_section.if_not_permitted.body_text") unless FeatureService.enabled?(:groups)
    return inactive_group_message unless @form.group.active?

    I18n.t("forms.task_list_create.make_form_live_section.user_cannot_administer.body_text", group_members_path: group_members_path(@form.group)) unless Pundit.policy(@current_user, @form).can_administer_group?
  end

  def inactive_group_message
    return I18n.t("forms.task_list_create.make_form_live_section.group_not_active.group_admin.body_text", upgrade_path: group_upgrade_url) if Pundit.policy(@current_user, @form).can_administer_group?

    I18n.t("forms.task_list_create.make_form_live_section.group_not_active.group_editor.body_text", group_members_path: group_members_path(@form.group))
  end
```

app/policies/form_policy.rb
jamie-o-wilkinson: I think that looks nicer...I'll do that 👍

app/service/form_task_list_service.rb
jamie-o-wilkinson: Yeah I like this suggestion too, it was getting too complicated at a glance I think

DavidBiddle (approved):

stephencdaly (commented):
Looks good - just wondering whether we should check that the form is in a group wherever we check the groups feature flag. This will probably mainly affect us while we're doing internal testing before we apply the migration.

app/policies/form_policy.rb
```diff
@@ -58,7 +58,12 @@ def can_add_page_routing_conditions?
     form_has_two_or_more_pages && form_has_qualifying_pages
   end
 
-  alias_method :can_make_form_live?, :can_change_form_submission_email?
+  def can_make_form_live?
+    return can_change_form_submission_email? unless FeatureService.enabled?(:groups)
```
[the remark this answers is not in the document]
DavidBiddle: We talked about this when developing it and decided to leave it out (on the logic that if this happens something has been migrated wrongly and we want it to get flagged to Sentry). But if it'll get in the way of our testing maybe we should add the check - what do you think @jamie-o-wilkinson?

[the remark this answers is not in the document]
jamie-o-wilkinson: Since everywhere else we're using the feature flag we do this check, we've added it to our new uses too. We've also added TODOs to make sure we remove the check along with the feature flag 👍

stephencdaly (approved):
Looks good to me!

PR #1196, opened 2024-05-30
Make new and edit group views match designs

DavidBiddle:

Trello card: https://trello.com/c/KUmN1mEt/1557-update-new-group-page-and-edit-group-name-page-to-match-designs

Covers the design changes in the ticket, namely:
- updating the heading text on both the new asnd edit group pages
- adding body text to the new group page
- adding a validation message when the name is left blank.

The card also specifies that there should be changes to the notification banner messages - those aren't covered here. There's a separate PR covering that (https://github.com/alphagov/forms-admin/pull/1133), and I don't think there's any reason they have to be merged at once.

##### Before
A page with a back link reading 'back', main heading 'New group', and a form with the label 'Name'.

##### After
A page with a back link reading 'Back to your forms', main heading 'Create a new group for forms', fiollowed by a large block of explanatory text including a list and inset text, and a form field with the label 'Enter a name for your new group'.

##### Before
A page with a back link reading 'Back to A trial group', main heading 'Edit group', and a form with the label 'Edit'.

##### After
A page with a back link reading 'Back to A trial group', and a form field with a label reading 'Change the name of this group'. The form field's label is also the page's main heading, and has the caption 'A trial group'.


thomasiles (approved):
looks good locally, nice!

PR #1205, opened 2024-06-03
Remove link to request upgrade when org has no org admins

stephencdaly:

Trello card: https://trello.com/c/slE3Jq7H/1569-remove-link-to-request-an-upgrade-from-task-list-for-trial-groups-with-no-org-admin

Don't display the link to request a group upgrade in the "Make your form live" section for a form if the organisation that the group belongs to does not have any organisation admins. This is because there will be nobody to request an upgrade from.

Instead, rely on the user seeing the various other banners that will tell them what to do to make their group active.

Screenshot 2024-06-03 at 15 29 12

It might be easier to hide whitespace changes in the diff when reviewing, as I added some extra contexts that changed indentation.

### DIFF
--- a/app/service/form_task_list_service.rb
+++ b/app/service/form_task_list_service.rb
@@ -139,6 +139,7 @@ private
   end
 
   def inactive_group_message
+    return I18n.t("forms.task_list_create.make_form_live_section.group_not_active.no_org_admin") if @form.group.organisation.admin_users.blank?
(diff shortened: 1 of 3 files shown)


DavidBiddle (approved):

DavidBiddle (commented):
Code looks good and works for me locally!

PR #1209, opened 2024-06-04
Replace Jest with Vitest for JavaScript testing

aliuk2012:

Using Jest and Vite means that you have effectively two different pipelines, one for testing and one for compiling, and each pipeline would need separate config files.  Vitest is a JavaScript testing framework that works natively with Vite without having a separate pipeline & config.  [Here is some great documentation on why Vitest exists](https://vitest.dev/guide/why.html).

As you can see there is not a lot of change from Jest but while running a few benchmarks I did find that Vitest was ever so slightly faster than Jest but I think the main benefit is having one single config to control and test frontend assets.

If you are willing to accept this PR, I'm happy to replicate this for forms-runner and forms-product-page (which seems to include puppeteer) 

Trello card: 


package.json
```diff
@@ -24,8 +24,6 @@
     "@testing-library/dom": "^10.1.0",
```
DavidBiddle, line 24: I suspect we might be able to drop `@babel/preset-env` - I think we were only using it to make Jest play nicely with ESM. If so, we can also drop the babel.config.json in the root folder.

package.json
```diff
@@ -13,7 +13,7 @@
     "Safari >= 12"
   ],
   "scripts": {
-    "test": "jest",
+    "test": "vitest",
```
DavidBiddle, line 16: I think this needs to be:
```suggestion
    "test": "vitest run",
```
since [Vitest runs in watch mode by default](https://vitest.dev/guide/cli.html#vitest) (so this task will never finish in the CI pipeline as it's currently written).

vite.config.js
```diff
@@ -23,5 +25,8 @@ export default defineConfig({
       ),
       '@images': path.resolve(__dirname, 'app/frontend/images')
     }
+  },
+  test: {
+    globals: true
```
DavidBiddle, line 30: I think there's an argument for not using these globals, but we can handle that separately if we think it's a good idea. Doing it like this makes sense for migrating from Jest so I'm happy with it in this PR.

DavidBiddle (commented):
Looks like a sensible change - thanks for the PR!

I've raised a couple of small things but really like this overall. Especially like how much simpler it makes the promises stuff!

package.json
aliuk2012: good spot but it does exit in CI pipeline https://github.com/Laings-Online-Limited/forms-admin/actions/runs/9366514608/job/25784128335?pr=1#step:11:1 

As it js testing and lining etc are run now together as part of vitest there's no real need to have this step in CI https://github.com/alphagov/forms-admin/blob/main/.github/workflows/test.yml#L69-L70

aliuk2012: https://vitest.dev/guide/cli.html#vitest

>vitest
>Start Vitest in the current directory. Will enter the watch mode in development environment and **run mode in CI automatically.**

DavidBiddle: Ooh, smart!

vite.config.js
aliuk2012: 💯  but to keep this PR smaller and focused, lets keep it as is and do another PR for globals

package.json
aliuk2012: Removed now

aliuk2012:
All done, @DavidBiddle  I think you might need to kickstart the CI because I'm an external contributor but once this has been merged I'll take a look at the other repos.

DavidBiddle (approved):
Looks good and tests run locally! 🎉

PR #1210, opened 2024-06-04
WIP: 1551 migration create a default group for each organisation

thomasiles:

Trello card: 

### DIFF
--- a/app/service/form_service.rb
+++ b/app/service/form_service.rb
@@ -11,4 +11,23 @@ class FormService
 
     form_path(@form.id)
   end
+
+  def add_to_default_group!(current_user)
+    return if current_user.trial?
+
+    org = current_user.organisation
+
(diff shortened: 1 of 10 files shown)


app/service/form_service.rb
```diff
@@ -30,4 +30,12 @@ def add_to_default_group!(current_user)
 
     org.default_group.memberships.find_or_create_by!(user: current_user, role: :editor, added_by: current_user)
   end
+
+  def assign_owner!(current_user)
```
stephencdaly, line 34: I'm not sure about this method name, I don't really understand what's happening here from this. Would it also make sense to squash the next commit into this commit, as this commit by itself isn't particularly informative about what's going on here. Looks like we're extracting a method?

stephencdaly:
We have these changes on another branch so closing this PR

PR #1213, opened 2024-06-05
Put forms in default groups

thomasiles:

Trello card: https://trello.com/c/ZbW37yS5/1551-migration-create-a-default-group-for-each-organisation

This PR prepares for switching the groups feature flag on by creating default groups for forms created without the feature turned on.

It includes changes to the way forms are created to work both with and without the feature flag and a rake task to migrate existing forms.

### DIFF
--- a/app/service/form_service.rb
+++ b/app/service/form_service.rb
@@ -11,4 +11,23 @@ class FormService
 
     form_path(@form.id)
   end
+
+  def add_to_default_group!(current_user)
+    return if current_user.trial?
+
+    org = current_user.organisation
+
+    if org.default_group.nil?
+      status = org.mou_signatures.present? ? :active : :trial
+      org.create_default_group!(name: "#{org.name} forms", organisation: org, status:)
+      org.save!
+    end
+
+    group_form = GroupForm.new(group: org.default_group)
+
+    group_form.form_id = @form.id
+    group_form.save!
+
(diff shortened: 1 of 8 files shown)


app/service/form_service.rb
```diff
@@ -11,4 +11,23 @@ def path_for_state
 
     form_path(@form.id)
   end
+
+  def add_to_default_group!(current_user)
+    return if current_user.trial?
+
+    org = current_user.organisation
+
+    if org.default_group.nil?
```
chao-xian, line 20: This could probably be in a separate method

DavidBiddle: Yeah, this feels sensible to me too

DavidBiddle (approved):
Code looks really good, and everything works on my machine 🎉 don't have anything to comment on other than echoing @chao-xian's point about extracting the default group creation into a separate method.

app/service/form_service.rb
thomasiles: I think this is a really good point and refactoring this code would make sense if we were planning on keeping it around beyond the migration.

As we've tested it along with our rake task we'd prefer to keep it as is for now, if thats ok.

lfdebrux: Worth noting that we did try having it in a different method when writing the code originally, but we found that it didn't bring much benefit.

PR #1219, opened 2024-06-07
Add trial group banner for case where organisation has no admins

lfdebrux:

Trello card: https://trello.com/c/te6u0prQ/1577-add-trial-group-banner-for-case-where-organisation-has-signed-mou-but-no-organisation-admins 

 In the case where an organisation has a signed MOU but no organisation admins, we don't allow users to request an upgrade. However the message we were showing in the group trial banner in this case doesn't make sense for group admins, as the logic in the view was falling through to the case where the user did not have permissions to request an upgrade.

This PR adds a case for when the organisation has signed an MOU but has no organisation admins, using the content defined in https://docs.google.com/document/d/1qRiztg6c33jRx7gvGEgq9i2w5JdDS_Vq4AtuxjSucJU/edit#heading=h.fzrys43i4ye1. 

Note that we have to duplicate some of the logic that is currently in the GroupPolicy#request_upgrade? method, to be able to tell why the upgrade cannot be requested. This might be an argument for keeping policy classes strictly related to authorisation.

### Screenshots

Screenshot of the groups page when the group is a trial group, the current user is a group admin for that group, and the organisation has signed an MOU but has no organisation admins. The trial banner reads This is a ‘trial’ group You can create forms in this group and test them, but you cannot make them live. Speak to your organisation’s GOV.UK publishing team or Link contact the GOV.UK Forms team to find out how to make live forms. The link "contact the GOV.UK Forms team" leads to mailto:govuk-forms-support@govuk.zendesk.com

### DIFF
--- a/app/views/groups/show.html.erb
+++ b/app/views/groups/show.html.erb
@@ -18,6 +18,10 @@
             <% nb.with_heading(text: t("groups.show.trial_banner.heading"), tag: "h3") %>
             <%= t("groups.show.trial_banner.request_upgrade.body_html") %>
             <%= govuk_link_to t("groups.show.trial_banner.request_upgrade.link"), request_upgrade_group_path(@group) %>
+          <% elsif @group.memberships.find_by(user: @current_user)&.group_admin? && !@group.organisation.admin_users.present? %>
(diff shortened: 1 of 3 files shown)


DavidBiddle (approved):
Code all looks good and I can see the new banner locally 🎉

PR #1220, opened 2024-06-07
Add rake default_groups:create_dry_run

thomasiles:
Wrap the rake task for migrating existing forms and editor users into default groups in a transaction.

This prevents changes to the database being committed if an exception is raised when running.

For example, the task will try to create a default group for any organisation which doesn't have one. If that organisation already has a group with same name, an error will be thrown and the task aborted. This would leave the DB with some organisations having default groups and some without.

Using a transaction means that either all the changes will happen, or none of them.

This is used to create a dry_run version of the task. At the end of the dry run task, a Rollback exception is raised. This means no changes are committed to the database.

Trello card: 

### DIFF
--- a/lib/tasks/default_groups.rake
+++ b/lib/tasks/default_groups.rake
@@ -1,35 +1,51 @@
 namespace :default_groups do
   desc "Create default groups for organisations"
   task create: :environment do
-    Organisation.joins(:users).where(users: { role: :editor }).distinct.each do |org|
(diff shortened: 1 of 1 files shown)


lfdebrux (dismissed):
Tested locally, works for me.

This is very nice, both in concept and execution.

I think this will be super useful, even if we don't end up needing it for this migration we should reuse this pattern in other future data migrations.

lfdebrux (approved):

PR #1221, opened 2024-06-07
Always log to STDOUT in production

lfdebrux:

We always want to log to standard out in production, instead of relying on the `RAILS_LOG_TO_STDOUT` environment variable. As well as allowing us to simplify our app configuration in forms-deploy [[1]] it also brings our app configuration closer in line with the Rails 7.1 defaults [[2]] and our other apps [[3], [4]].

[1]: https://github.com/alphagov/forms-deploy/blob/96cab32c19ab68515c357d6de7c9d3f801022ee8/infra/modules/forms-admin/main.tf#L106
[2]: https://github.com/rails/rails/blob/v7.1.2/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt#L62-L65
[3]: https://github.com/alphagov/forms-api/blob/88a0ad5818a1b626cd494fd8bfccb2392bccc66d/config/environments/production.rb#L47-L49
[4]: https://github.com/alphagov/forms-api/blob/88a0ad5818a1b626cd494fd8bfccb2392bccc66d/config/environments/production.rb#L47-L49

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -47,12 +47,6 @@ module FormsAdmin
     # logging use the Logger::Formatter.new.
     config.log_formatter = JsonLogFormatter.new
 
-    if ENV["RAILS_LOG_TO_STDOUT"].present?
-      config.logger = ActiveSupport::Logger.new($stdout)
-      config.logger.formatter = config.log_formatter
-
-    end
-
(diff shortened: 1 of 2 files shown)


DavidBiddle (approved):
This change seems sensible to me 🎉

PR #1227, opened 2024-06-11
Enable groups feature for GDS organisation

lfdebrux:

Trello card: https://trello.com/c/ZEki1GAK/1553-migration-enable-groups-feature-for-gds 

We want to test the groups feature internally; this PR enables the feature flag for our organisation.

### DIFF
--- a/config/settings/production.yml
+++ b/config/settings/production.yml
@@ -1,6 +1,8 @@
 features:
   groups:
     enabled: false
+    organisations:
+      government_digital_service: true
 
 # Use real authentication
 auth_provider: gds_sso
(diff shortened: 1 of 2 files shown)


config/settings/production.yml
```diff
@@ -1,6 +1,8 @@
 features:
   groups:
     enabled: false
+    organisation:
```
stephencdaly, line 4: Should be `organisations`

lfdebrux: Thanks!

lfdebrux: Fixed.

stephencdaly (approved):

PR #1229, opened 2024-06-11
WIP: Add app changes and tests for forms

thomasiles:

Trello card: https://trello.com/c/t4bUdTbz/1555-migration-create-a-default-group-for-each-trial-user


DavidBiddle (approved):
Code looks good and think I've got it working locally 🎉 

PR description might benefit from a bit more detail/context but there's information in the invidual commit messages so I don't think that's a blocker.

lfdebrux:
> Code looks good and think I've got it working locally 🎉
> 
> PR description might benefit from a bit more detail/context but there's information in the invidual commit messages so I don't think that's a blocker.

Thanks for the review! I've added a bit more detail to the PR description, a good point well made 😁

DavidBiddle:
Extra info looks good - thanks @lfdebrux!

PR #1236, opened 2024-06-14
Create a default group and add forms for trial users who set name and organisation

thomasiles:
### Create a default group and add forms for trial users who set name and organisation

Trello card: https://trello.com/c/t4bUdTbz/1555-migration-create-a-default-group-for-each-trial-user

We have not migrated the forms owned by trial users who do not have name or organisation set into groups.

This PR adds code which runs when a user sets name or org. It checks that they are a trial user and creates a new default group if required, adds the user as group admin and adds any forms owned by the user which are not already in groups.

As this only happens when a trial user signs in to the service, it will take a while for all forms to be added to groups.

### DIFF
--- a/app/service/default_group_service.rb
+++ b/app/service/default_group_service.rb
@@ -1,0 +1,45 @@
+class DefaultGroupService
+  def create_trial_user_default_group!(user)
+    return unless user.trial? && user.name.present? && user.organisation.present?
+
+    forms = Form.where(creator_id: user.id).to_h { [_1.id, _1] }
+    form_ids = forms.keys
+    group_form_ids = GroupForm.where(form_id: form_ids).pluck(:form_id)
+    not_group_form_ids = form_ids.to_set - group_form_ids
+
+    if not_group_form_ids.blank?
+      Rails.logger.info "DefaultGroupService: Trial user '#{user.name}' does not have any forms not in groups, skipping creating default group"
+      return
+    end
+
+    Rails.logger.info "DefaultGroupService: Trial user '#{user.name}' default group creation starting"
+
+    default_trial_group = Group.find_or_create_by!(
+      creator_id: user.id,
+      name: "#{user.name}’s trial group",
+      organisation_id: user.organisation_id,
+      status: :trial,
+    ) do |new_group|
+      Rails.logger.info "DefaultGroupService: Created default group '#{new_group.name}', with creator '#{new_group.creator.email}'"
+    end
+
+    default_trial_group.memberships.find_or_create_by!(
+      user:,
+      role: :group_admin,
+      added_by: user,
+    ) do |new_membership|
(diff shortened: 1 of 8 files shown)


app/service/default_group_service.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    Rails.logger.info "default_groups: trial user #{user.name}"
+
+    default_trial_group = Group.find_or_create_by!(
+      creator_id: user.id,
+      name: "#{user.name}’s trial group",
+      organisation_id: user.organisation_id,
+      status: :trial,
+    ) do |new_group|
+      Rails.logger.info "default_groups: created #{new_group.name}, with creator #{new_group.creator.email}"
```
stephencdaly, line 23: Is there a way we can avoid logging user names and email addresses while still logging something useful? I'm not sure the exact rules around logging PII for our internal users but it feels safer not to unless we absolutely need to.

Could we log the user `id` and/or `uid` instead?

spec/service/default_group_service_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    end
+
+    before do
+      ActiveResource::HttpMock.respond_to do |mock|
+        mock.get "/api/v1/forms?creator_id=#{user.id}", headers, forms_response.to_json, 200
+      end
+    end
+
+    context "when user has a trial role" do
+      it "creates a group if the user does not already have one" do
```
stephencdaly, line 22: "if the user does not already have one" doesn't sound so certain about what it's testing

Using contexts would make it easier to read:

```suggestion
      when "the user does not already have a trial group" do
          it "creates a group" do
```

spec/service/default_group_service_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      it "adds their trial forms to the group" do
+        expect {
+          default_group_service.create_trial_user_default_group!(user)
+        }.to change(GroupForm, :count).by(1)
+
+        expect(GroupForm.last).to have_attributes(form_id: form.id)
+      end
+
+      it "raises an exception if it tries to add their trial forms to an active group" do
```
stephencdaly, line 56: This could be clearer about what scenario this can occur under. Something like:

```suggestion
      when "a default group already exists with active status" do
          when "the user has forms without a group" do
              it "raises an exception" do
```

stephencdaly (commented):
Changes look good, just some minor comments about logging and test names

app/service/default_group_service.rb
lfdebrux: We already log email address for all users in the admin app

stephencdaly: Ah ok, if this isn't something new and we've decided it's ok then don't see an issue.

spec/service/default_group_service_spec.rb
jamie-o-wilkinson: I've rearranged the contexts in a way I _think_ is more logical. See what you think!

PR #1238, opened 2024-06-17
Fix default_groups:create for organisations with only super admin users

lfdebrux:

Trello card: https://trello.com/c/82ZeBA9C/1598-investigate-forms-that-have-not-been-migrated-to-groups 

We have some organisations in production where the only users are super admins, so they got missed by the `create_organisation_default_groups` script. This commit fixes this.


thomasiles (approved):
nice

PR #1243, opened 2024-06-18
Add Google analytics snippet

DavidBiddle:

Trello card: https://trello.com/c/EmmWXBP5/1587-add-google-analytics-to-admin-to-track-what-form-builders-are-doing

This card adds Google analytics to any page, as long as:
- the user is not a super admin
- the flag (`Settings.analytics_enabled`) is `true`

You can test whether GA is running on an particular page by opening your browser console and seeing if typing `window.dataLayer` returns anything - if it returns `undefined` it isn't running, if it returns an array of event objects it is running.


stephencdaly (approved):
All looks good to me, it's well tested and I ran the check locally, and it looks like it works as expected

PR #1245, opened 2024-06-19
Fix feature tests for group flag enabled

thomasiles:

Trello card: https://trello.com/c/EsoNSIL4/1573-migration-improve-automated-tests-for-groups-feature

Enable the groups feature flag in test by default and change any tests which need it. This PR also removes unnecessary toggling of the feature flag in tests, now that it is on by default.


spec/features/users/set_organisation_spec.rb
```diff
@@ -33,13 +18,6 @@
     then_the_users_organisation_name_is_updated
   end
 
-  scenario "Super admin can change their own organisation" do
-    given_i_am_a_super_admin_in_the_government_digital_service_organisation
-    when_i_change_my_organisation_to_test_org
-    then_i_cannot_see_the_government_digital_service_forms
-    but_i_can_see_the_test_org_forms
-  end
```
lfdebrux, line 41: We felt these tests were unnecessary?

spec/views/forms/show.html.erb_spec.rb
```diff
@@ -36,23 +36,21 @@
     expect(rendered).to have_selector(".app-task-list__summary", text: "You’ve completed 12 of 20 tasks.")
   end
 
-  context "when form state is draft", feature_groups: true do
```
lfdebrux, line 39: Is there any value to keeping this context? Why was it included in the first place?

spec/service/form_task_list_service_spec.rb
```diff
@@ -422,6 +424,46 @@
               expect(section_rows.first[:task_name]).to eq "Make your form live"
               expect(section_rows.first[:path]).to eq "/forms/1/make-live"
             end
+
+            context "when form is ready to make live" do
```
lfdebrux, line 428: These changes look good, but I think go beyond what I'd call a refactor... does the commit message need to be updated?

spec/features/users/set_organisation_spec.rb
thomasiles: yeah sorry we should have commented on this. We decided this wasn't really a very useful test to have as there is nothing special about changing organisation and being super admin.

Within groups, we already test that super admin can see other organisation groups.

spec/views/forms/show.html.erb_spec.rb
thomasiles: It adds context to the form status, which is draft by default.

I think we could go either way with this one. If we decide to add the context back, I think there are other tests that should be under it. We were erring on the side of less nesting.

spec/service/form_task_list_service_spec.rb
jamie-o-wilkinson: Good point, I don't think I properly considered the commit message. I've changed it to "Add missing tests for FormTaskListService when groups are enabled"

lfdebrux (approved):

PR #1248, opened 2024-06-20
Enable groups feature in development

thomasiles:
The groups feature flag should be on by default in development. Currently the flag is not enabled because of a typo in the development settings file.

This file fixes the typo to correctly enable the flag.

Trello card: 

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -1,4 +1,4 @@
-feature:
+features:
   groups:
     enabled: true
 


lfdebrux (approved):
😱 thanks for fixing this!

danworth (approved):
Wonder if there's a nice practical way of avoiding this in future, a simple test that looks for unused settings keys perhaps? Would also help to keep the file clean as flags are removed from our codebase.

thomasiles:
> Wonder if there's a nice practical way of avoiding this in future, a simple test that looks for unused settings keys perhaps? Would also help to keep the file clean as flags are removed from our codebase.

Th [config gem can do validation of the settings](https://github.com/rubyconfig/config?tab=readme-ov-file#validation) to make sure they match a schema. I think that could pick it up, with the cost of a bit of work in making and maintaining the schema.

PR #1250, opened 2024-06-20
Ensure trial and editor users have the same permissions when groups feature is enabled

lfdebrux:

Trello card: https://trello.com/c/SoKlQmep/1597-migration-ensure-trial-editor-and-standard-roles-are-equivalent-when-groups-feature-is-enabled 

We need to be sure that when the groups feature is enabled that users with the trial or editor role will have the same permissions, as shortly afterwards we want to get rid of the trial role and rename the editor role to 'standard'.

This means that any restrictions that apply to users with the trial role should be lifted when the groups feature is enabled.

### DIFF
--- a/app/policies/form_policy.rb
+++ b/app/policies/form_policy.rb
@@ -35,7 +35,7 @@ class FormPolicy
   def can_view_form?
     return true if user.super_admin?
 
-    if FeatureService.new(user).enabled?(:groups) && form.group.present?
+    if groups_enabled?
       return user.groups.include?(form.group) || user.is_organisations_admin?(form.group.organisation)
     end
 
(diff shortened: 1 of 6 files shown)


jamie-o-wilkinson (approved):
LGTM 🥇

app/policies/form_policy.rb
```diff
@@ -72,6 +72,10 @@ def can_administer_group?
 
 private
 
+  def groups_enabled?
```
chao-xian, line 77: Lovely. Clearly named method that tells you what it's doing before I have to read the code, and when I do read the code I know if it's actually doing what's expected based on the name. 👏  more please!

chao-xian (approved):

PR #1252, opened 2024-06-20
Prevent forms without groups being made live if groups are enabled

jamie-o-wilkinson:
Previously, we were allowing forms without a group to be made live. Although these wouldn't be listed in groups and the forms index would've been unreachable, if a user had the url for an existing form they could have made the form live.

Trello card: https://trello.com/c/EsoNSIL4


jamie-o-wilkinson:
Closing as superseded by #1282

PR #1253, opened 2024-06-20
Change memberships table to be sorted by user name

lfdebrux:

Currently the order of users in the table on the "Members of this group" page is whatever is returned by the database, which is not consistent; so when editing users the table can jump around between page loads, which is a bit disorientating.

This PR fixes things by ordering by name. This is a naive way, but at least is more likely to have consistent results.

#### Before

https://github.com/alphagov/forms-admin/assets/503614/4cb6522b-7ad6-4665-88df-d38a9a504bf1

(A screen recording showing editing users in a group before this change. The order of the users in the table changes each time a user is made a group admin.)

#### After

https://github.com/alphagov/forms-admin/assets/503614/3d60a9e4-f298-41f1-aaa6-101362713ffb

(A screen recording showing editing users in a group after this change. The order of the users in the table is the same each time a user is made a group admin.)

### DIFF
--- a/app/views/group_members/index.html.erb
+++ b/app/views/group_members/index.html.erb
@@ -23,7 +23,7 @@
     end %>
 
     <%= table.with_body do |body| %>
-      <% @group.memberships.each do |membership| %>
+      <% @group.memberships.joins(:user).order("users.name").each do |membership| %>
         <%= body.with_row do |row| %>
           <%= row.with_cell { membership.user.name } %>
           <%= row.with_cell { membership.user.email } %>
--- a/app/views/group_members/index.html.erb
+++ b/app/views/group_members/index.html.erb
@@ -23,7 +23,7 @@
     end %>
 
     <%= table.with_body do |body| %>
-      <% @group.memberships.joins(:user).order("users.name").each do |membership| %>
+      <% @group.memberships.ordered.each do |membership| %>
         <%= body.with_row do |row| %>
           <%= row.with_cell { membership.user.name } %>
           <%= row.with_cell { membership.user.email } %>
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -5,7 +5,11 @@ class Group < ApplicationRecord
 
   belongs_to :upgrade_requester, class_name: "User", optional: true
 
-  has_many :memberships, dependent: :destroy
+  has_many :memberships, dependent: :destroy do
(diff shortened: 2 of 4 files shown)


app/views/group_members/index.html.erb
```diff
@@ -23,7 +23,7 @@
     end %>
 
     <%= table.with_body do |body| %>
-      <% @group.memberships.each do |membership| %>
+      <% @group.memberships.joins(:user).order("users.name").each do |membership| %>
```
stephencdaly, line 26: I wonder whether the ordering should be in the model, maybe as a method - rather than being in the view?

stephencdaly (approved):
Nice spot!

Fix looks good, just a question about where the ordering logic should live. But happy for this to be merged as is.

app/views/group_members/index.html.erb
lfdebrux: Yeah, my initial thought was that it would be nice to have the ordering on the model, but to be honest the main reason I didn't do that is because I didn't know what to call it 😅 

Actually my first thought was that we could just put a scope on the `users` association in the `Group` model, so it's always sorted, but then I realised that we do actually need to iterate by memberships to get the user's role in the group. I then thought about putting the scope on the memberships, but that feels like it would be inefficient.

Anyway it would be fairly straightforward to add an [association extension](https://guides.rubyonrails.org/association_basics.html#association-extensions) that does this little snippet, similar to what we do in https://github.com/alphagov/forms-admin/commit/877e26b1be4cea82b30e48e52b18736c847e9c64, but yeah I don't know what to call it 😅 Any suggestions?

stephencdaly: `ordered_memberships`?

lfdebrux: @stephencdaly ooh interesting idea, I hadn't considered that as an option, I was sort of thinking we needed to shoehorn "by user name" in there somehow.

What about `memberships.ordered` (having an association extension instead of a plain method)?

stephencdaly: Oh, I didn't properly look at what the association extensions did. Yeh that would make sense to me!

lfdebrux: Cool I've made that change now!

stephencdaly (approved):
Really nice!

PR #1255, opened 2024-06-21
Fix warning about scoped order

lfdebrux:

Currently when running the `organisations:summary` task we get a warning from Rails about `Scoped order is ignored, it's forced to be batch order.`, because we're using `find_each` but the `with_users` scope adds an `ORDER BY` clause.

This PR fixes things by removing the order relation with [unscope](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-order).

When testing locally the warning appears in `log/development.log`, but in production it is emitted to stdout.


DavidBiddle (approved):
Change looks good and I can no longer see the warning in the logs 🎉

PR #1270, opened 2024-06-26
Change forms:move to also update form organisation

lfdebrux:

Trello card: https://trello.com/c/7EkOqWib/1605-add-rake-task-to-move-forms-from-one-group-to-another 

When moving a form to a group in a different organisation to the one it currently belongs to, we should update the form's organisation attribute.

Not doing so shouldn't have any effect when the groups feature is enabled, but until we remove the organisation attribute we should keep things consistent.

### DIFF
--- a/lib/tasks/forms.rake
+++ b/lib/tasks/forms.rake
@@ -34,7 +34,7 @@ def move_forms(form_ids, group_id)
     group_form = GroupForm.find_or_initialize_by(form_id:)
 
     if group_form.group == group
-      Rails.logger.info "forms:move: keeping form #{fmt_form(form)} in #{fmt_group(group)}"
+      Rails.logger.info "forms:move: keeping #{fmt_form(form)} in #{fmt_group(group)}"
       next
     elsif group_form.persisted?
       Rails.logger.info "forms:move: moving #{fmt_form(form)} from #{fmt_group(group_form.group)} to #{fmt_group(group)}"
@@ -51,5 +51,5 @@ def fmt_form(form)
 end
 
 def fmt_group(group)
-  "group #{group.id} (\"#{group.name}\", #{group.organisation.name}, #{group.creator&.name || 'GOV.UK Forms Team'})"
+  "group #{group.external_id} (\"#{group.name}\", #{group.organisation.name}, #{group.creator&.name || 'GOV.UK Forms Team'})"
 end
--- a/lib/tasks/forms.rake
+++ b/lib/tasks/forms.rake
@@ -43,6 +43,13 @@ def move_forms(form_ids, group_id)
     end
 
     group_form.update!(group:)
+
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/forms.rake_spec.rb
```diff
@@ -78,6 +79,48 @@
...
+
+          it "does not change the group organisation" do
+            expect(ActiveResource::HttpMock.requests).not_to include update_organisation_request
+          end
+        end
+
+        context "and in a different organisation to the group" do
+          let(:form_organisation) { create :organisation, slug: "other-org" }
+
+          it "changes the group organisation" do
```
jamie-o-wilkinson, line 118:
```suggestion
          it "changes the form's organisation" do
```

jamie-o-wilkinson (commented):
One suggestion, I think it's a typo in a test name. Otherwise, looks good! 👍

spec/lib/tasks/forms.rake_spec.rb
lfdebrux: Ah nice catch, thanks!

lfdebrux: Fixed in https://github.com/alphagov/forms-admin/compare/378d2171551d4ba942f0b442a73c89f3a578928a..341b71b436924a596d1f21a5588069dc2772b18b

lib/tasks/forms.rake
```diff
@@ -43,6 +43,13 @@ def move_forms(form_ids, group_id)
     end
 
     group_form.update!(group:)
+
+    next unless form.organisation && form.organisation != group.organisation
+
+    Rails.logger.info "forms:move: updating #{fmt_form(form)} organisation from #{form.organisation.name} to #{group.organisation.name}"
+
+    form.organisation = group.organisation.name
```
DavidBiddle, line 51: This doesn't seem to work on my machine - the task logs that it's updating the organisation, but the organisation doesn't actually change. 

I think this line might be the issue - you might need to send the ID rather than the name. I can see in the API log that the `PUT` request being sent contains the old organisation ID
```suggestion
    form.organisation_id = group.organisation.id
```

lfdebrux: Ah. Thanks for catching that! I wonder why my test didn't work 🤔

DavidBiddle: That's weird!

lfdebrux: I've updated the tests to use our custom `have_been_updated` matcher.

DavidBiddle (approved):

lib/tasks/forms.rake
DavidBiddle: All working for me locally now 👍

PR #1276, opened 2024-06-28
Delete the old homepage

lfdebrux:

Trello card: https://trello.com/c/tUpB9Xpe/1613-remove-formsindex-and-linked-classes 

This PR removes the old "homepage" which showed the user's forms. This page is no longer used, it's been replaced by the groups page.

We can also remove components that were only used on the old homepage.


thomasiles (approved):
nicely laid out in commits 👏

PR #1282, opened 2024-07-01
Refactor policies with groups disabled

thomasiles:
### Change the Form policies now that the groups feature is enabled

Trello card: https://trello.com/c/QoZWJP6I/1614-cleanup-forms-admin-policies-to-focus-on-the-groups-model

Now that the groups feature is the default, the policies for accessing forms can be cleaned up.

The policies no longer need to consider the older model, which relied on organisation membership and trial roles.

One of the policies, `can_change_form_submission_email?`, is no longer needed. This policy used to prevent trial users changing the submission email.

Now, anyone who can access the group is allowed to change the submission email. This matches the other Form attributes so the specific policy is no longer required.

The other large change to application code is in the `FormTaskListService`. This used to offer different tasks if the user was a trial user. This code and it's tests has been removed.

## Request and Feature test changes
Changing the FormPolicy means the current request and feature tests must be rewritten so that forms belong to groups and that the current user belongs to this group. Without this change even unrelated tests will fail with a permissions error.

This means a large amount of tests have been changed in this commit. It's mostly a mechanical change to how the tests are arranged.

Ideally, these changes would be made in another PR to reduce the size of this one. We didn't do that because without the changes to the policy, the test setup wouldn't be checked by tests.


spec/requests/forms/what_happens_next_controller_spec.rb
```diff
@@ -66,7 +72,7 @@
     end
 
     context "when the user is not authorised to view the form" do
-      let(:form_organisation_id) { 999 }
+      let(:user) { build :user }
```
thomasiles, line 75: These tests have been changed so that the form is in the same group, but we try to access it with another user who  isn't a member.

This is easier to change than making the test setup conditional.

spec/requests/forms_controller_spec.rb
```diff
@@ -74,31 +76,6 @@
         expect(response.status).to eq(403)
       end
     end
-
-    context "with a user with a trial account" do
```
thomasiles, line 78: We no longer need to test for trial accounts.

stephencdaly (dismissed):
All the changes make sense and look good to me. @jamie-o-wilkinson and I gave it a fair kick locally and everything seems to be working as expected.

jamie-o-wilkinson (dismissed):
LGTM, although I did work on it so I'm probably biased 😄 

One thing @stephencdaly and I discussed was that retrospectively we probably could have split this into multiple PRs.

jamie-o-wilkinson (approved):

PR #1296, opened 2024-07-05
Ensure DeleteConfirmationController actions are authorized by Pundit

lfdebrux:


DavidBiddle (approved):

PR #1297, opened 2024-07-05
Make find invited user by email case insensitive

stephencdaly:

Trello card: https://trello.com/c/JBBA5kyk/1623-make-email-matching-case-insensitive-when-adding-someone-to-a-group

When users are created in Signon and Auth0, the email is downcased when it is stored.

When inviting users to join groups, we should therefore ignore the case when locating the user to invite. Previously, we were showing an error saying that the user does not have a Forms account when the case did not match the stored email for the user.

### DIFF
--- a/app/input_objects/group_member_input.rb
+++ b/app/input_objects/group_member_input.rb
@@ -30,7 +30,7 @@ class GroupMemberInput < BaseInput
   end
 
   def invited_user
-    @invited_user ||= User.find_by(email: member_email_address)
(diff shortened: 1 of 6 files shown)


chao-xian (changes requested):
Please don't approve/merge this yet as this might not be the right solution

chao-xian:
A unique constraint on email addresses would make any clash issues not our problem.

chao-xian (approved):

PR #1302, opened 2024-07-08
Add default context to all application logs

stephencdaly:

Trello card: https://trello.com/c/tjnJrmjR/7-improve-default-information-we-add-to-logs

We already use Lograge to add attributes about the request context, such as the `request_id`, `user_id` etc to the request logs for incoming requests to the app.

However, ad-hoc application logs and default Rails logs such as for requests made to other services did not get any default context attributes added to them.

To make request context attributes available to all log lines, this commit adds an implementation of the `ActiveSupport::CurrentAttributes` abstract super class, which provides a thread-isolated attributes singleton, which resets automatically before and after each request.

Introduces a custom logger class - `ApplicationLogger` to use for application logs. This makes use of the `CurrentAttributes` implementation to extract common logging fields and appends them to the JSON object for a log event.

By configuring the application to use `ApplicationLogger`, all logs will automatically get these default attributes, making them much more useful and findable in Splunk searches.

The Lograge configuration has also been updated to extract these fields from the `CurrentLoggingAttributes` so we get the same set of fields on both the application and request logs.

This PR also proposes removing `EventLogger` in favour of just using `Rails.logger.info` as it seemed like `EventLogger` was added to mirror the EventLogger in runner but I was unsure of the advantage this offered in the case of admin over logging more descriptive messages using the `message` field of a structured log line. This is something that I'm happy to open a discussion over.


app/models/current_logging_attributes.rb
```diff
@@ -0,0 +1,19 @@
...
+      user_id:,
+      user_email:,
+      user_organisation_slug:,
+      user_ip:,
+      request_id:,
+      form_id:,
+      page_id:,
+      session_id_hash:,
+      trace_id:,
+    }.compact
```
lfdebrux, line 17: Can we use `attributes` here instead?

stephencdaly: Good point, yes we can!

app/input_objects/account/organisation_input.rb
```diff
@@ -22,9 +22,7 @@ def assign_form_values
 private
 
   def log_organisation_chosen_event
-    EventLogger.log({
-      event: "organisation_chosen",
```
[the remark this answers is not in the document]
stephencdaly: I think it makes sense for runner logs when we're thinking about important events in the lifecycle of filling out a form, but I'm unsure of the advantage in this case over just logging a descriptive string that can be searched.

Is the intention to discourage people from putting too much detail in the log "message"/"event" and instead attaching it to the log as JSON attributes?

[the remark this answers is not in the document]
lfdebrux: > I think it makes sense for runner logs when we're thinking about important events in the lifecycle of filling out a form, but I'm unsure of the advantage in this case over just logging a descriptive string that can be searched.

Hmm, I think it's mainly a subjective choice, there are pros and cons for each style.

In this particular case I would say that the advantage of `organisation_chosen` over `User chose their organisation` is that the first one can be searched without worrying about quoting spaces or worrying about the search term catching other log events. But the second is definitely easier to read and understand.

> Is the intention to discourage people from putting too much detail in the log "message"/"event" and instead attaching it to the log as JSON attributes?

I'm not sure, I don't think I've ever heard it articulated like that, although it is possible that was the intention.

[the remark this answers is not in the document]
lfdebrux: Makes sense!

Of course, there's no reason a JSON log line couldn't have both an `event` field and a human readable message.

config/environments/production.rb
```diff
@@ -46,8 +46,8 @@
   config.force_ssl = true
 
   # Log to STDOUT by default
-  config.logger = ActiveSupport::Logger.new($stdout)
-    .tap { |logger| logger.formatter = config.log_formatter }
+  # config.logger = ActiveSupport::Logger.new($stdout)
+  #   .tap { |logger| logger.formatter = config.log_formatter }
```
lfdebrux, line 50: I'd prefer if we kept the logging configuration in `config/environments/production.rb`, so that it's closer to what is in the default Rails configuration.

Or, if we don't want to do that, we should probably delete these lines rather than commenting them out.

stephencdaly: The config should be the same for production and development, which is why I've put the configuration in `application.rb` rather than here.

I can delete these lines rather than comment them out - I saw that we deliberately comment out a lot of the defaults to make it clear that we don't want to use those defaults so wasn't sure what to do here. If you think it's better to delete rather than comment our I can do that.

app/controllers/application_controller.rb
```diff
@@ -39,20 +40,19 @@ def check_maintenance_mode_is_enabled
     end
   end
 
-  def append_info_to_payload(payload)
-    super
-    payload[:host] = request.host
+  def set_logging_attributes
+    CurrentLoggingAttributes.host = request.host
+    CurrentLoggingAttributes.request_id = request.request_id
+    CurrentLoggingAttributes.session_id_hash = Digest::SHA256.hexdigest session.id.to_s if session.exists?
```
danworth, line 46: If `session.exists?` then is it certain that `session.id` exists? Otherwise we'll get a NPE on `session.id.to_s`, Might be safer to use the safe reference like `session.id&.to_s`. I can see this is copied from the existing but worth checking I think. My concern with common logging is that it's invoked on every request and if its wrong then it can cause unfortunate issues.

app/input_objects/account/organisation_input.rb
```diff
@@ -22,9 +22,7 @@ def assign_form_values
...
-      user_id: user.id,
+    Rails.logger.info("User chose their organisation", {
```
danworth, line 25: Please check with Anne whether is affects her reporting if this message content and key changes; I ask because if it's currently in the `event` field then it might be being used.

spec/lib/application_logger_spec.rb
```diff
@@ -0,0 +1,72 @@
+require "rails_helper"
```
danworth, line 1: Nice tests 👍

danworth (commented):
I think using the `ActiveSupport::CurrentAttributes` is a nice improvement on passing our own logging context around 🎉 . Making it simpler to log ad-hoc messages with the necessary common fields to aid debugging and observability is also a nice change 🏅 

There's been some back-and-forth on where logging configuration should exist with consideration to `application.rb` and the various environment override files e.g. `production.rb` so please check with @chao-xian on his preferred direction. A couple of comments, one more important one to check with Anne before changing what is currently logged within an `event` field.

I presume this has been deployed into dev and seen working as expected under various conditions. I see the addition of unit tests which is great but worth checking that it all renders correctly all the way through to Splunk (there are have been mishaps in the past and its difficult to automate such integrated testing through the entire logging pipeline).

app/controllers/application_controller.rb
stephencdaly: Yeh that's a good point, I'll make that change to make it safer

app/lib/application_logger.rb
```diff
@@ -0,0 +1,43 @@
+require "json"
+
+class ApplicationLogger < ActiveSupport::Logger
```
chao-xian, line 3: Minor thing but worth doing: you could split adding this class and the spec in a prior commit to make this one a smaller set of changes in flight

app/lib/application_logger.rb
```diff
@@ -0,0 +1,43 @@
...
+    message_to_hash(msg).merge(attribs || {}).merge(CurrentLoggingAttributes.attributes).compact
+  end
+
+  def message_to_hash(message)
+    return { message: } if message.is_a? String
+    return message if message.is_a? Hash
+
+    raise ArgumentError, "message must be a string or hash"
+  end
+end
```
chao-xian: Feels like this could be a LogMessage class?

config/environments/production.rb
chao-xian: My view is we should have local less noisy and if we want it to be like production, say for debugging, an individual can easily flip that on. Plus keeping things closer to Rails means when we do our next upgrade we've deviated less.

app/lib/application_logger.rb
stephencdaly: I've split this out in the runner PR as there are more changes so thought it was worthwhile. While I agree with the principle, can I gently push back on doing that here, because I think the changes make more sense when viewed all together and also while its not much effort, it is effort unpicking and redoing commits.

app/lib/application_logger.rb
stephencdaly: Not sure I follow exactly what you're suggesting here. The intention here is that this class follows the same interface as the standard rails logger, where the first argument passed to `logger.info` etc. is a message, which can be of any type. This code deals with converting this to a hash and merging the logging context onto that hash before logging it - not sure which of this would benefit from extracting it out.

config/environments/production.rb
stephencdaly: Ok sure - happy to go with whatever the general preference is, so would your suggestion be to have the default logger configured for development, and only have the logger that adds the context for production?

How about the Lograge request logging, should I keep that how it is in `application.rb` or move that too?

app/input_objects/account/organisation_input.rb
chao-xian: On that note, can we mark whatever fields Anne is relying on with some sort of comment pls?

app/lib/application_logger.rb
chao-xian: "Do the hard work to make it simple" - I understand the effort, and that's why I suggested that video 😄 

It's literally much harder for me to understand.

app/input_objects/account/organisation_input.rb
stephencdaly: I've checked with Anne and she doesn't use this log lines for any reporting.

What Me and Dan were thinking is that in forms-runner we could potentially move towards treating anything logged using the `EventLogger` as something we shouldn't be changed as it will be used for analytics purposes.
If we do this, we could then use `Rails.logger` for logging anything adhoc that isn't used for analytics and have a distinction there.

What do you reckon? Something we can maybe think of next time we're wanting to add more logging to runner.

config/environments/production.rb
chao-xian: Yes I think so on the first one. I think on the second one feels that we should move out for consistency.

app/lib/application_logger.rb
chao-xian: For further context and for others to understand where I'm coming from: my adhd couldn’t cope - and it’s not just for me but for any other folks who struggle as I do. Generally unless it’s a “find and replace all” type of change, if the number of files affected is more than ~5 or 6 files I collapse under the weight of the page scroll (see also very large diffs).

This is the video I mentioned: https://www.youtube.com/watch?v=GyUxe7PMD4A

stephencdaly: I've separated this out now. Thanks for the context on why it's useful to do this!

chao-xian: TYSM!

chao-xian (approved):
This is really awesome work and I deeply appreciate the changes you've made from my feedback. Nice one!

PR #1312, opened 2024-07-12
Ensure requests specs are testing behaviour when groups feature is enabled

lfdebrux:

We are now using the groups feature everywhere, and we want to make sure our specs are testing things with the groups feature enabled. For our request specs this means making sure that forms belong to a group (especially when testing code related to access control).

This does make our request specs a bit larger and slower though; it might be worth asking whether we should be mocking more things in these specs. Currently I think we treat them like integration tests though, and I wasn't sure that now was the time to open that question, and instead focused on making it so that we can remove our legacy non-groups code.

### DIFF
--- a/spec/requests/forms/archive_form_controller_spec.rb
+++ b/spec/requests/forms/archive_form_controller_spec.rb
@@ -5,6 +5,10 @@ RSpec.describe Forms::ArchiveFormController, type: :request do
   let(:form) { build(:form, :live, id:) }
 
   before do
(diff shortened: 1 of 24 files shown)


lfdebrux:
> I think there is quite a lot of overlap between this PR and #1282. Is it ok if we merge #1282 and then merge this over the top?

Yep, whatever works best!

lfdebrux:
Closing in favour of #1282 and #1312.

PR #1324, opened 2024-07-17
Reduce authorisation session timeout to 8 hours

stephencdaly:

Trello card: https://trello.com/c/nIuSV6Fa/1563-rtp-ref-7-reduce-forms-admin-user-session-timeout-to-8-hours

Change the length of time the authorisation sesssion is valid for to 8 hours. This means that we will require the user to login again after this time has elapsed.

For Auth0, this will force a check with Auth0 that the session with Auth0 is still valid, if it is the user won't be required to login again. We will separately adjust the Auth0 application configuration to reduce the length of time the session is valid for.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -47,7 +47,11 @@ sentry:
 # How we authenticate users
 auth_provider: # use default auth_provider from environment
 
-auth_valid_for: 86400 # the time after which re-authentication is required, in seconds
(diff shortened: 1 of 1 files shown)


thomasiles (approved):
looks good

jamie-o-wilkinson (approved):

PR #1327, opened 2024-07-18
Remove homepage_controller

stephencdaly:

Trello card: https://trello.com/c/tRzLMO7a/1612-remove-homepagecontroller-from-forms-admin

Now that the groups feature has been enabled, and the code that checks the feature flag has been removed, we can remove the HomepageController and set the app root to the groups index page.

This means that visiting `/` directly renders the groups index page. `/groups` also renders the groups index page, and this is where the back links point to. We don't see any issues with this behaviour and it just makes testing slightly easier.

Keeping the back links pointing to `/groups` also means that if we update the homepage in future, there will be fewer changes to make.


jamie-o-wilkinson (approved):
LGTM and works locally 🥇

PR #1331, opened 2024-07-19
Remove form policy scope

stephencdaly:

Remove a redundant usage of FormPolicy::Scope, and then remove FormPolicy::Scope as it is no longer used anywhere since enabling the groups feature.


SamJamCul (approved):
Looks good, get rid of it!

PR #1337, opened 2024-07-23
Add standard user role

jamie-o-wilkinson:

This PR adds a new Standard role to the User model. This is intended to supersede both trial and editor roles, and this is the first step in migrating all trial and editor users to standard users


stephencdaly (approved):
I paired on this, but I am confident in these changes. No behaviour will change as a result of this PR yet as no users will be standard users.

I have tested locally that acting as a `standard` user works as expected.

PR #1338, opened 2024-07-23
Update user name and org validation

jamie-o-wilkinson:

Previously, when checking whether we needed to validate a user's name or organisation, we checked whether that user's role had changed from trial. This is no longer needed, as we require all users to set their name and organisation regardless of role. It is enough to just prevent name and organisation from being unset.

### DIFF
--- a/spec/requests/users_controller_spec.rb
+++ b/spec/requests/users_controller_spec.rb
@@ -191,8 +191,8 @@ RSpec.describe UsersController, type: :request do
           expect(user.reload.organisation).to be_nil
         end
 
-        it "returns an error if organisation is not chosen and role is changed to editor" do
-          put user_path(user), params: { user: { role: "editor", organisation_id: nil } }
+        it "returns an error if organisation is not chosen and role is changed to organisation_admin" do
+          put user_path(user), params: { user: { role: "organisation_admin", organisation_id: nil } }
           expect(response).to have_http_status(:unprocessable_entity)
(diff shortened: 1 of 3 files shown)


spec/requests/users_controller_spec.rb
```diff
@@ -175,17 +175,11 @@
       context "with a trial user with no name set" do
         let(:user) { create(:user, :with_trial_role, name: nil) }
 
-        it "does not return error if name is not chosen and role is not changed" do
+        it "does not return error if name is not chosen" do
```
chao-xian, line 178: What are your thoughts on renaming this? Seems the actual thing it's doing is redirecting if a name is missing

lfdebrux (dismissed):
Nice!

chao-xian (approved):
One small comment but looks good otherwise looks good

spec/requests/users_controller_spec.rb
stephencdaly: I think the name still makes sense. It successfully updates the user if the name is not chosen, so long as there wasn't an existing name - rather than returning an error, which is what this is testing.

I guess it could be renamed to "successfully updates the user when a name is not set"

jamie-o-wilkinson: We've renamed it to "successfully updates the user when a name is not set"

lfdebrux (approved):

PR #1339, opened 2024-07-23
Remove user missing organisation error

jamie-o-wilkinson:

We longer need to raise UserMissingOrganisationError in the form policy, as the policies appropriately handle a user's organisation being missing. This PR removes the error itself, raising of it, handling it, and the template shown when raised. 


thomasiles (approved):
This LGTM and is nicely all red 👍

PR #1341, opened 2024-07-23
Move authorization for delete_confirmtion_controller#destroy to start

lfdebrux:

Because we were authorizing the delete in the `delete_page` and `delete_form` helpers, we weren't checking the user was authorized when they decided not to delete or did not send a valid request.

This PR fixes things by making sure that checking the user's authorization is the first things we do in the action.

We also add a spec for when user chooses not to delete form. This bug was introduced because we weren't testing the path when a user presses the delete button for a form but then chooses not to delete it at the confirmation page.

### DIFF
--- a/app/controllers/forms/delete_confirmation_controller.rb
+++ b/app/controllers/forms/delete_confirmation_controller.rb
@@ -9,6 +9,8 @@ module Forms
     end
 
     def destroy
+      authorize current_form
+
       load_page_variables
       @delete_confirmation_input = DeleteConfirmationInput.new(delete_confirmation_input_params)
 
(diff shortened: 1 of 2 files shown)


stephencdaly (commented):
The changes look good. I wonder whether there should be tests for each controller action for when the user is not authorised to make sure it shows the appropriate error.

Also this controller is a bit strange in that it is used for deleting forms and also pages within forms. Should the tests be added to to add cases for deleting a page rather than a form? That might go a bit beyond the scope of this change though..

lfdebrux:
> The changes look good. I wonder whether there should be tests for each controller action for when the user is not authorised to make sure it returns the appropriate error.

Good shout on the tests for the errors, I've added those now.

> Also this controller is a bit strange in that it is used for deleting forms and also pages within forms. Should the tests be added to to add cases for deleting a page rather than a form? That might go a bit beyond the scope of this change though..

Yeah this controller is very weird. Personally I'd like to see it refactored, I've put it down as a candidate for one of the mob refactoring sessions.

jamie-o-wilkinson (approved):

stephencdaly (approved):
Nice one

PR #1342, opened 2024-07-23
Update the creation of default groups to remove reference to trial users

jamie-o-wilkinson:

This PR updates the creation of default groups to remove reference to trial users. We've also removed the associated rake task to create default groups for trial users, as this should no longer be necessary.


chao-xian (approved):
Nice work

PR #1345, opened 2024-07-24
Make sure groups feature is enabled everywhere

lfdebrux:

Trello card: https://trello.com/c/k3WbZBZw/36-remove-groups-feature-flag

We're confident now that we want to use the groups feature going forward, and have started removing the legacy code, so we want it enabled everywhere.

This commit removes the groups feature flag override from settings/*.yml files so that the global value is used and we have the groups feature enabled for everyone everywhere.

Note that the feature flag code is still around, we're just making sure that it's always set to true.

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -1,7 +1,3 @@
-features:
-  groups:
-    enabled: true
-
 forms_api:
   # URL to form-api endpoints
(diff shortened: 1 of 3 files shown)


chao-xian (approved):

cadmiumcat (approved):
🎏

PR #1346, opened 2024-07-24
Remove policy check for submission email task

stephencdaly:

There was a redundant check that the user could view the form when deciding whether to allow the user to complete the submission email task. This existed because previously we were checking that the user wasn't a trial user.

There was no need for this policy check now, as a user that doesn't have permission to view the form would not be able to visit the task list page.


thomasiles (approved):
good spot

lfdebrux (approved):

PR #1347, opened 2024-07-24
Attempt to fix flaky tests

stephencdaly:

These tests were checking that 2 arrays of groups/forms were exactly equal. However, as the code under test was retreiving the groups/forms from the database and not applying sorting, the order was unpredictable. This was causing the tests to occasionally fail.

Replace `eq` assertions with `match_array`, which ignores the ordering


lfdebrux:
Wow, that's really interesting, I had no idea this test was failing, I guess I got lucky 😅 How often were you seeing failures?

stephencdaly:
> Wow, that's really interesting, I had no idea this test was failing, I guess I got lucky 😅 How often were you seeing failures?

Only once so far, so it's probably pretty rare

chao-xian (approved):

stephencdaly:
We've talked on Slack about this and we want to apply an ordering to groups/forms and there is a trello ticket for this. I'll merge this PR anyway as it should prevent these tests sporadically failing before we do this.

PR #1350, opened 2024-07-25
Change default user role to standard

stephencdaly:


lib/tasks/users.rake
```diff
@@ -0,0 +1,27 @@
...
+  Rails.logger.info("Number of trial users: #{User.where(role: 'trial').count}")
+  Rails.logger.info("Number of editor users: #{User.where(role: 'editor').count}")
+  Rails.logger.info("Number of standard users: #{User.where(role: 'standard').count}")
+
+  User.where(role: %w[trial editor]).update_all(role: "standard")
+
+  Rails.logger.info("Number of trial users: #{User.where(role: 'trial').count}")
+  Rails.logger.info("Number of editor users: #{User.where(role: 'editor').count}")
+  Rails.logger.info("Number of standard users: #{User.where(role: 'standard').count}")
+end
```
thomasiles, line 27: small thing might not be needed. Is it helpful to log super admin and org admin counts as well just to be sure they haven't changed?

thomasiles (dismissed):
Amazing work!

This was a huge task and it's really nicely broken down and documented 🏅

lib/tasks/users.rake
jamie-o-wilkinson: Good suggestion. I've added that 👍

thomasiles (approved):

PR #1352, opened 2024-07-26
Add features report

DavidBiddle:

Trello card: https://trello.com/c/mrozr2YJ/1652-report-for-the-features-okr

Adds a new report which lists:
- the number of live forms
- the number of live forms that contain a given answer type
- the total number of live form pages that use each answer type
- the number of live forms with a payment URL.
- the number of live forms with at least one condition.

Additionally adds:
- a new reports index where we can link to this and future report pages
- a link to the reports index in the header.

The link and all of the new reports pages can only be accessed by super admins.

#### Header
GOV.UK Forms header, with navigation links for MOUs, Users, Reports, and Support

#### Index
GOV.UK Forms page with a heading 'Reports' and a single link, 'Feature usage on live forms'

#### Features report
GOV.UK Forms page with a heading 'Feature usage on live forms'. There is a summary list with the heading 'Features' and entries for 'Total live forms', 'Live forms with routes' and 'Live forms with payments'. There is a table with the heading 'Answer types', and column headings 'Answer type', 'Number of live forms which ask for this answer type', and 'Number of pages of this answer type on live forms'. There are rows for all of the answer types

### DIFF
--- a/spec/views/reports/features.html.erb_spec.rb
+++ b/spec/views/reports/features.html.erb_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe "users/edit.html.erb" do
+describe "reports/features.html.erb" do
   let(:report) do
     Report.new({ total_live_forms: 3,
                  live_forms_with_answer_type: { address: 1,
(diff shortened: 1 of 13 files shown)


spec/views/reports/features.html.erb_spec.rb
```diff
@@ -47,15 +47,33 @@
...
+      expect(rendered).to have_css("[data-live-forms-with-answer-type-#{answer_type.to_s.dasherize}]", text: report.live_forms_with_answer_type.attributes[answer_type].to_s)
     end
 
     it "includes the number of live pages with #{answer_type}" do
-      expect(response.body).to include "Number of pages of this answer type: #{report.live_pages_with_answer_type.attributes[answer_type]}"
+      expect(rendered).to have_css("[data-live-pages-with-answer-type-#{answer_type.to_s.dasherize}]", text: report.live_pages_with_answer_type.attributes[answer_type].to_s)
+    end
+  end
+
+  context "when an answer type is missing from the data" do
```
thomasiles, line 66: nice! I was wondering if this would make it harder to add new answer types.

This test covers that case 👏

thomasiles (approved):
great PR, works well locally and is really nicely structured 👏

PR #1353, opened 2024-07-26
Remove trial and editor roles

jamie-o-wilkinson:

Now that we have migrated all users with the trial and editor roles to the standard role, we can safely delete any and all code that references those trial and editor roles. This means we've deleted these roles themselves, along with any remaining usages of them in code and in tests.

We can also delete some rake tasks used to migrate users to the standard role and to summarise trial users, which no longer exist.


app/models/user.rb
```diff
@@ -24,8 +24,6 @@ class UserAuthenticationException < StandardError; end
   enum :role, {
     super_admin: "super_admin",
     organisation_admin: "organisation_admin",
-    editor: "editor",
-    trial: "trial",
```
thomasiles, line 28: When I tested the code I tried to access the /users route and an exception was thrown because I hadn't run the rake task to convert trial users into standard users.

I'm not sure any code needs to change but maybe a reminder in slack telling people to run in their local environments would be handy?

It did make me wonder if it's safer to always keep old values of this enum in for safety. I think by this point we are confident that trial and editors should not exist and be treated as errors if they do? We also use strings instead of numbers so no users would suddenly change roles if the indexes change.

thomasiles (approved):
Nicely sausage made!

Removing trial and editors is super satisfying, I can't find any issues with it. It looks comprehensive and I guess any sneaky bits left over can be removed later if we come across them.

🏅

app/models/user.rb
jamie-o-wilkinson: Good catch @thomasiles! Yes we've gone through each environment and converted all trial and editor users to the standard role, so any trial or editors should be treated as errors from now on.

Hmm we've deleted the migration rake task...do you think we should retain this, or message that people should drop and reseed their local database?

PR #1357, opened 2024-07-30
Fix creation of standard users in db/seeds.rb

jamie-o-wilkinson:

During the replacement of trial and editors roles with standard, we replaced the `editor_user` factory with the default user factory with `:standard` shorthand syntax. We made a mistake doing this in the database seed, which this PR fixes.


thomasiles (approved):
good catch!

PR #1359, opened 2024-07-30
Remove trial and editor roles

jamie-o-wilkinson:

This PR reapplies the removal of trial and editor roles, but with the users rake task retained so that we can verify there are no trial or editor users remaining in production after this PR gets deployed.


DavidBiddle (approved):

jamie-o-wilkinson:
I've done a dry run of the role migration task which shows there are no trial or editor users remaining in any environment. I'll run again after this is merged.

PR #1362, opened 2024-07-31
Push pageview and external link events to GA

DavidBiddle:

Trello card: https://trello.com/c/qAojecL7/1602-enable-google-analytics-in-admin-and-runner

We previously added and enabled google analytics in admin and runner, but events aren't coming through in GA. THis is because we weren't actually pushing the events we care about into the dataLayer.

This PR does that. The events we're pushing are:
- page views - every time a user views a page, we push an event containing some basic data about the page to GA
- external link clicks - every time a user clicks a link without a relative URL, we log some information about that URL to GA

### How to test
Firstly:
- Set `analytics_enabled: true` in your `development.yml` file
- login as a non-super admin user
- open the broswer console

To test page views: 
- load any page on your local forms-admin
- enter `dataLayer` in the console and inspect the array that comes back
- you should see one object in the array that looks like this:
```js
{
  event:"page_view",
  page_view:{
    location:{
      href:"http://localhost:3000/",
      origin:"http://localhost:3000/",
      protocol:"http:",
      host:"localhost:3000",
      hostname:"localhost",
      port:"3000",
      pathname:"/",
      search:"",
      hash:""
    },
    referrer:"",
    schema_name:"simple_schema",
    status_code:200,
    title:"Your groups – GOV.UK Forms"
  },
  gtm.uniqueEventId":2
}
```

To test external link clicks:
- make your life easier by stopping links taking you to another page - run this snippet in your browser console:
  - `document.querySelectorAll('a').forEach((link)=>{link.addEventListener('click',(event)=>{event.preventDefault()})})`
- click an external link (e.g. the Open Government Licence link in the footer)
- enter `dataLayer` in the console and inspect the array that comes back
- you should see one object in the array that looks like this:
```js
{
  event:"event_data",
  event_data:{
    event_name:"navigation",
    external:true,
    method:"primary click",
    text:"Open Government Licence v3.0",
    type:"generic link",
    url:"https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/"
  },
  gtm.uniqueEventId":5
}
```

### DIFF
--- a/app/frontend/entrypoints/application.js
+++ b/app/frontend/entrypoints/application.js
@@ -45,6 +45,7 @@ document
 if (document.body.dataset.googleAnalyticsEnabled === 'true') {
   installAnalyticsScript(window)
   sendPageViewEvent()
+  attachExternalLinkTracker()
 }
 
 initAll()
--- a/app/frontend/entrypoints/application.js
+++ b/app/frontend/entrypoints/application.js
@@ -3,7 +3,11 @@ import dfeAutocomplete from 'dfe-autocomplete'
 import copyToClipboard from '../javascript/copy-to-clipboard'
 import markdownEditorToolbar from '../javascript/markdown-editor-toolbar'
 import { pasteListener } from '../javascript/paste-html-to-markdown'
-import { installAnalyticsScript } from '../javascript/google-tag'
+import {
+  installAnalyticsScript,
+  sendPageViewEvent,
+  attachExternalLinkTracker
+} from '../javascript/google-tag'
 import ajaxMarkdownPreview from '../javascript/ajax-markdown-preview'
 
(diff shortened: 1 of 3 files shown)


thomasiles (approved):
This looks really good and neat to me and I really like the trick to disable navigation on external links.

I tried to test it with the google analytics debugger which spits out events sent to GA to the developer console. It doesn't show any events being sent. I do see them in dataLayer though. Is that expected?

DavidBiddle:
@thomasiles - interesting, that's not intentional. I can see some (impossible to parse) requests being sent to GTM when I click on external links so I think _something_ is being sent. I think I'll merge this and see if the performance folks are getting anything useful back from it

PR #1364, opened 2024-08-01
Make DraftQuestion forwards compatible with Page response

lfdebrux:

Trello card: https://trello.com/c/UpuZ4LCB/1674-3-add-add-another-answer-attribute-to-questions-in-forms-api-3 

Currently if an attribute is added to the Page model in forms-api trying to edit the question for that page will result in an error:

     Failure/Error: edit_draft_question.attributes = page.attributes.except(:id, :position, :next_page, :has_routing_errors, :routing_conditions, :question_with_text)

     ActiveModel::UnknownAttributeError:
       unknown attribute 'newly_added_to_api' for DraftQuestion.
     # ./app/controllers/pages_controller.rb:65:in `setup_draft_question_for_existing_page'
     # ./app/controllers/pages_controller.rb:47:in `draft_question'
     # ./app/controllers/pages/questions_controller.rb:3:in `block in <class:QuestionsController>'
     # ./spec/requests/pages/questions_controller_spec.rb:171:in `block (4 levels) in <top (required)>'
     # ./spec/support/features.rb:19:in `block (2 levels) in <top (required)>'

This is happening because the attributes passed to `edit_draft_question` includes the new attributes, but these attributes are not in the schema for the `draft_questions` table.

This PR fixes things by filtering the attributes set in `edit_draft_question` to only those known to the `DraftQuestion` model.

This makes it a lot easier to add attributes to pages in forms-api, otherwise we would have to worry about adding any new attribute to forms-admin first.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -62,7 +62,10 @@ private
     edit_draft_question = DraftQuestion.find_or_initialize_by(form_id: current_form.id, user_id: current_user.id, page_id: page.id)
 
     if edit_draft_question.new_record?
-      edit_draft_question.attributes = page.attributes.except(:id, :position, :next_page, :has_routing_errors, :routing_conditions, :question_with_text)
+      attributes = page.attributes
+        .slice(*edit_draft_question.attribute_names)
+        .except(:id)
+      edit_draft_question.attributes = attributes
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
works well locally and doesn't break if a new attribute is added to DraftQuestion and isn't in the API response yet. 👏

PR #1366, opened 2024-08-02
Add new report to show the organisation count of users

thomasiles:
### Create a report showing simple stats about users in organisations

Trello card: https://trello.com/c/lbm0G3EX/1659-create-report-to-show-number-of-users-per-organisation

This PR adds a simple report to show organisations and the number of users which belong to those organisations.

It's really likely that we will want to iterate this:
- improving he content
- changing the stats reported, e.g. to add user types
- add export as CSV or other formats

This PR is only the MVP.

The reports index page:
image

The user report:
image

There is no need to display the table caption and heading as they are the same. I'll change it when the content is checked to see if there are any changes needed.

### DIFF
--- a/app/service/users_report_service.rb
+++ b/app/service/users_report_service.rb
@@ -1,0 +1,34 @@
+class UsersReportService
+  def user_data
+    {
+      caption: I18n.t("reports.users.heading"),
+      head: [
+        { text: I18n.t("reports.users.table_headings.organisation_name") },
+        { text: I18n.t("reports.users.table_headings.user_count"), numeric: true },
+      ],
+      rows:,
+    }
+  end
+
+private
+
+  def rows
+    as_data_rows(org_name_user_count)
+  end
+
+  def as_data_rows(raw_data)
+    raw_data.map do |org_name, count|
+      [{ text: org_name || I18n.t("users.index.organisation_blank") },
+       { text: count, numeric: true }]
+    end
+  end
+
+  def org_name_user_count
+    User.left_joins(:organisation)
+      .group("organisations.id")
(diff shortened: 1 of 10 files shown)


config/locales/en.yml
```diff
@@ -1028,14 +1028,14 @@ en:
     heading: Provide a link to privacy information for this form
     submit_button: Save and continue
   reports:
+    back_link: Back to reports
```
DavidBiddle, line 1031: Good catch!

app/service/users_report_service.rb
```diff
@@ -0,0 +1,34 @@
+class UsersReportService
+  def user_data
+    {
+      caption: I18n.t("reports.users.heading"),
+      head: [
+        { text: I18n.t("reports.users.table_headings.organisation_name") },
+        { text: I18n.t("reports.users.table_headings.user_count"), numeric: true },
+      ],
+      rows:,
```
DavidBiddle, line 9: I think it would make sense to make the first cell a row header, as in the design system examples: https://design-system.service.gov.uk/components/table/

```suggestion
      rows:,
      first_cell_is_header: true
```

spec/views/reports/users.html.erb_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "rails_helper"
+
+describe "reports/useres.html.erb" do
```
DavidBiddle, line 3: Tiny typo:

```suggestion
describe "reports/users.html.erb" do
```

DavidBiddle (approved):
Raised a couple of tiny suggestions but looks great overall :tada:

Really like how you've wrangled the data in UsersReportService to keep the view simple!

PR #1369, opened 2024-08-05
Remove organisation hint from edit user page

hannahkc:

An edit of content on the 'edit user' page that was missed. Removing the 'organisation' hint as it is not really needed and is potentially misleading now that most users will only see forms in groups they are in (rather than all their organisation's forms). 

Trello card:https://trello.com/c/nXOuz0qN/1609-migration-change-trial-and-editor-users-to-standard-users

I have not run this locally. I have just guessed that this will work. Please check my changes carefully! 


DavidBiddle (approved):
Works locally and code change looks good 🎉

PR #1370, opened 2024-08-05
Update ActiveRecord query to use includes

chao-xian:
...instead of joins as the latter is lazy, and that causes the loop to make additional calls to the database to get data for the associated organisation per user.

See:
- https://dev.to/jetthoughts/what-is-the-difference-between-joins-and-includes-in-rails-activerecord-23jk
- https://guides.rubyonrails.org/active_record_querying.html#includes

### DIFF
--- a/app/service/users_report_service.rb
+++ b/app/service/users_report_service.rb
@@ -25,7 +25,7 @@ private
   end
 
   def org_name_user_count
-    User.left_joins(:organisation)
+    User.includes(:organisation)
       .group("organisations.id")
       .select("organisations.name, COUNT(users.id) AS user_count")
       .order(Arel.sql("COUNT(users.id) DESC"))


thomasiles:
> ...instead of joins as the latter is lazy, and that causes the loop to make additional calls to the database to get data for the associated organisation per user.
> 
> See:
> 
> * https://dev.to/jetthoughts/what-is-the-difference-between-joins-and-includes-in-rails-activerecord-23jk
> * https://guides.rubyonrails.org/active_record_querying.html#includes

I don't think this is right in this case is it? They both produce the same query - a left outer join?

I think what you are referring to is right when loading other models but it isn't the case here, where we are purposefully selecting two columns. Includes is kinda complicated and does different things in different circumstances, I think.

Can you double check the output?

Is there an argument that `includes` is clearer than `left_ joins`?

DavidBiddle (approved):

chao-xian:
I've been running some `Explains` to double check but having a bit of trouble with it and not 100% conclusive on it being better/faster/efficient. Happy to leave as is.

PR #1373, opened 2024-08-06
Add default Google Analytics consent

DavidBiddle:

Trello card: https://trello.com/c/qAojecL7/1602-enable-google-analytics-in-admin-and-runner

In forms-admin, we don't ask the user for analytics consent because it's an internal tool and we're not required to. However, our Google tag code and google analytics still expect to be told whether the user has consented.

This PR does this by:
- setting the `analytics_consent` cookie to `true` on page load
- passing the default consent item to the dataLayer: 
```js
[
  'consent',
  'default',
  {
    ad_storage: 'denied',
    analytics_storage: 'granted'
  }
]
```

### Testing instructions
Firstly:

- Set analytics_enabled: true in your development.yml file
- login as a non-super admin user
- open the broswer console

Then test cookies:
- go to 'Storage' in Firefox devtools or  'Application' in the chrome
- check that there is an `analytics_enabled` cookie with value `true`  

And test the dataLayer consent item:
- go to the console and type 'dataLayer'
- check that one of the items in the array looks like the default consent item above

### DIFF
--- a/app/frontend/entrypoints/application.js
+++ b/app/frontend/entrypoints/application.js
@@ -8,6 +8,7 @@ import {
   sendPageViewEvent,
   attachExternalLinkTracker
 } from '../javascript/google-tag'
+import { saveConsentStatus } from '../javascript/utils/cookie-consent'
 import ajaxMarkdownPreview from '../javascript/ajax-markdown-preview'
 
 document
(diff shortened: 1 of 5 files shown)


thomasiles (approved):
works locally 👍

PR #1376, opened 2024-08-09
Handle mailchimp unsubscribers

DavidBiddle:

Trello card: https://trello.com/c/tr2zhSTd/1668-we-need-to-handle-mailchimp-unsubscribes-better

If users manually unsubscribe form a MailChimp list (e.g. by clicking the unsubscribe link in an email) MailChimp assigns them the 'unsubscribed' status. Users with this status cannot be resubscribed or archived via the API, and attempting to do so results in an error.

This PR updates the task to avoid trying to resubscribe these users. In the process there were also a few other changes:
- fixed a bug where users who get archived (e.g. by having their Forms accounts disabled) aren't resubscribed if they become active again
- fixed a bug with the pagination logic (we were looping through the wrong number of items)
- various renames to make it more obvious what the script is doing in MailChimp
- rewrote the tests to make them more obvious and more explicit about how the different MailChimp statuses should be handled

### DIFF
--- a/spec/lib/mailchimp_list_synchronizer_spec.rb
+++ b/spec/lib/mailchimp_list_synchronizer_spec.rb
@@ -21,6 +21,7 @@ RSpec.describe MailchimpListSynchronizer do
         "name" => "List 1",
         "stats" => {
           "member_count" => 3,
+          "total_contacts" => 3,
         },
       }
     end
@@ -117,7 +118,8 @@ RSpec.describe MailchimpListSynchronizer do
         {
           "name" => "List 1",
           "stats" => {
-            "member_count" => 1001,
+            "member_count" => 995,
+            "total_contacts" => 1001,
           },
         }
       end
--- a/spec/lib/mailchimp_list_synchronizer_spec.rb
+++ b/spec/lib/mailchimp_list_synchronizer_spec.rb
@@ -20,31 +20,10 @@ RSpec.describe MailchimpListSynchronizer do
       {
         "name" => "List 1",
         "stats" => {
-          "member_count" => 3,
-          "total_contacts" => 3,
+          "member_count" => 1,
+          "total_contacts" => 1,
         },
       }
-    end
-
-    let(:list_1_members_info) do
-      {
-        "members" => [
-          { "email_address" => "keep@domain.org", "status" => "subscribed" },
-          { "email_address" => "archive@domain.org", "status" => "subscribed" },
-          { "email_address" => "retireduser@domain.org", "status" => "subscribed" },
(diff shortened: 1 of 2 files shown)


spec/lib/mailchimp_list_synchronizer_spec.rb
```diff
@@ -77,91 +56,227 @@
...
+      end
     end
 
-    it "archives users from MailChimp who do not appear in the database, but do appear in the mailing list" do
-      archived_email_hash = Digest::MD5.hexdigest "archive@domain.org"
+    RSpec.shared_examples "it does not subscribe or archive the user" do
+      it "does not subscribe or archive the user" do
+        archived_email_hash = Digest::MD5.hexdigest "user@domain.org"
+
+        expect(mailchimp_client_lists).not_to receive(:set_list_member).with(
```
stephencdaly, line 88: I wonder whether this expectation is quite brittle. Because we're asserting on the method not receiving a call with very specific arguments, this might continue to pass if the logic under test doesn't work as expected but the hash we pass to the method changes slightly.

Could we instead either assert that `:set_list_member` is either not called with any arguments, or if it is called do an expectation on the arguments that they don't match the ones we don't want it to be called with.

Something like
```
expect(mailchimp_client_lists).to receive(:set_list_member) do |list_id, subscriber_hash, body|
  expect(body[:email_address]).not_to eq("user@domain.org")
end
```

I'm not convinced what I've suggested is great, there might be a better way to do it.

stephencdaly (approved):
Changes look good and are well broken down into commits with good explanations.

I have a slight concern about the use of `not_to receive` in the specs, but not entirely sure my concerns are valid. You can consider what I've written, but I'm happy for this to be merged. Happy to talk it through either way.

spec/lib/mailchimp_list_synchronizer_spec.rb
stephencdaly: Actually it's probably better if it is called to check it is called exactly once and with the expected arguments. Something like this?

```
expect(mailchimp_client_lists).to receive(:set_list_member).once do |list_id, subscriber_hash, body|
  expect(body["email_address"]).to eq("some_other_user@domain.org")
end
```

DavidBiddle: Yeah, I think this is valid and that seems like a good approach - thanks for the suggestion!

DavidBiddle: Have updated!

stephencdaly (approved):
Looks good to me!

PR #1385, opened 2024-08-14
Increase the number of options allowed in selections to 30

AP-Hunt:

Trello card: https://trello.com/c/ZMi9kq7P

Increases the maximum number of items in a selection type question to 30, from 20.

Did I catch all the places?

### DIFF
--- a/app/input_objects/pages/selections_settings_input.rb
+++ b/app/input_objects/pages/selections_settings_input.rb
@@ -35,7 +35,7 @@ class Pages::SelectionsSettingsInput < BaseInput
     filter_out_blank_options
 
     return errors.add(:selection_options, :minimum) if selection_options.length < 2
-    return errors.add(:selection_options, :maximum) if selection_options.length > 20
+    return errors.add(:selection_options, :maximum) if selection_options.length > 30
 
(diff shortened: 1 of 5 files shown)


cadmiumcat (approved):
Looks good to me 👍 , maybe needs a dev reviewing?

thomasiles:
Nice work!

Should users in the admin be able to add up to 30 through the interface? I think given the ticket and the changes you have made, they should.

Currently I still see the message about not being able to add any more answers when i get to 20. This is because the view [limits the amount of options which can be added](https://github.com/alphagov/forms-admin/blob/a5fc9515232de1c6e64337be81c81fd9958b66e9/app/views/pages/selections_settings.html.erb#L46).

Obviously there are better ways we could do this, not least sticking that number in a constant and referencing it in the different places it's accessed.

I'd be happy just changing that line in the view to a 30 for now, given that this ticket is for a support request.

I'd use github change suggestions but I don't know how to do it.

I've tested it locally and everything works with that change. If you want me to push a commit, slack me and I'll stick it in.

AP-Hunt:
Nice catch @thomasiles. I missed that in the enormous output of `git grep 20` I was sifting through 😂

PR #1389, opened 2024-08-19
Fix vitest silent failures

DavidBiddle:

Trello card: None

Error messages in our vitest specs weren't causing the tests to fail - for example: https://github.com/alphagov/forms-admin/actions/runs/10419073126/job/28856405937?pr=1383#step:11:24

In this case the error turned out to be specific to the test, but these silent failures are risky as they mean we might accidentally break our javascript and not be alerted to it.

This PR:
- updates our vitest config to throw an error when a test calls `console.error`
- fixes the jsdom error that caused the error we saw by stubbing out the navigation behaviour.

See commit messages for complete context.

### DIFF
--- a/app/frontend/test/setup.js
+++ b/app/frontend/test/setup.js
@@ -1,0 +1,14 @@
+import { beforeEach, vi } from 'vitest'
+
+beforeEach(() => {
+  // in some cases the console.error is called immediately after the test is
+  // run, so we have to do this here instead of an afterEach block
+  vi.restoreAllMocks()
+
+  // mock console.error and make it throw a real error in the test suite context
+  vi.spyOn(console, 'error').mockImplementation(message => {
+    throw new Error(
+      `Failing due to test calling console.error with message: ${message}`
+    )
+  })
+})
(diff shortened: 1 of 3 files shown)


app/frontend/test/setup.js
```diff
@@ -0,0 +1,5 @@
+console.error = message => {
+  throw new Error(
+    `Failing due to console.error while running test!\n\n${message}`
```
chao-xian, line 3: So this overrides what console.error does and throws a real error in the test suite's context? Can you point to the docs that this is what to do pls in the PR?

DavidBiddle: > So this overrides what console.error does and throws a real error in the test suite's context?

Correct

> Can you point to the docs that this is what to do pls in the PR?

There's nothing in the docs about this. I've seen a few people do workarounds for this but they're all variations on this theme.

Thinking about it I think it might make more sense to do this in a beforeEach block, use a spy instead, and restore it in an afterEach block. I'll have a go and see if that works.

lfdebrux: If yes it might be worth raising an issue with Vitest developers requesting this functionality (if there isn't already an issue), seeing as they are saying it should be a drop-in replacement for Jest.

lfdebrux (approved):

chao-xian (approved):

PR #1392, opened 2024-08-21
Add repeatable page radio input

thomasiles:
### Add a new input for creating/editing pages to allow multiple answers

Trello card: https://trello.com/c/J74xCxkW/1700-5-5-implement-add-another-answer-in-forms-admin-for-single-question-only

This work is behind the repeating_page_enabled feature flag.

This PR adds a new radio option when creating or editing the page of a form. This option sets the `is_repeating` boolean field on pages in the API.

The feature flag is only used to guard the control in the view. This means that it's possible using developer tools to set `is_repeating` on pages regardless of the value of the feature flag.

Given that:
- the feature flag is only there to make release easier
- the API and runner both support the `is_repeating` field

It shouldn't be a problem. It should also make removing the feature flag easier.

image

### DIFF
--- a/app/views/pages/_form.html.erb
+++ b/app/views/pages/_form.html.erb
@@ -34,6 +34,11 @@
     <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible" />
   <% end %>
 
+  <% if Settings.features.repeatable_page_enabled %>
+    <%= f.govuk_collection_radio_buttons :is_repeatable, question_input.repeatable_options, :id, :name, :description, legend: { size: 'm', tag: 'h2' }, bold_labels: false %>
+    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible" />
+  <% end %>
+
   <h2 class="govuk-heading-m"><%= t("pages.answer_settings") %></h2>
   <%= render PageSettingsSummaryComponent::View.new(draft_question) %>
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -39,6 +39,8 @@ en:
               too_long: Hint text must be %{count} characters or less
             is_optional:
               inclusion: Select ‘mandatory’ if people have to answer this question
+            is_repeatable:
+              inclusion: Select ‘Yes’ if someone can add more than one answer
             question_text:
               blank: Enter a question
               too_long: Question text must be %{count} characters or less
@@ -512,6 +514,7 @@ en:
       pages_question_input:
         is_optional_options:
           'true': We’ll add ‘(optional)’ to the end of the question text.
+        is_repeatable: Select ‘Yes’ if you want to let someone add more than one answer (up to 10) — for example, listing all their company contacts.
     label:
       account_name_input:
         name: Enter your full name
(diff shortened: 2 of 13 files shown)


app/views/pages/_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= form_with model: [form_object, question_input], url: action_path do |f| %>
   <%= f.govuk_error_summary %>
 
+
```
lfdebrux, line 4:
```suggestion
```

config/locales/en.yml
```diff
@@ -616,12 +619,16 @@ en:
         is_optional_options:
           'false': Mandatory
           'true': Optional
+        is_repeatable_options:
+          'false': No — this question can only be answered once
```
jamie-o-wilkinson, line 623: I wondered whether "this question can only be answered once" should be in hint text, matching the hint for Optional?

lfdebrux (approved):
Works for me; very nice!

spec/views/pages/_forms.html.erb_spec.rb
```diff
@@ -40,6 +40,18 @@
...
   end
 
+  context "when feature repeatable page is enabled", :feature_repeatable_page_enabled do
+    it "has an radio input for repeatable" do
+      expect(rendered).to have_field("pages_question_input[is_repeatable]", type: :radio)
+    end
+  end
+
+  context "when feature repeatable page is not enabled", feature_repeatable_page_enabled: false do
+    it "has an radio input for repeatable" do
```
thomasiles, line 50: "does not have a radio input for repeatable" 🤦

jamie-o-wilkinson, line 50:
```suggestion
    it "does not have a radio input for repeatable" do
```

spec/views/pages/_forms.html.erb_spec.rb
```diff
@@ -40,6 +40,18 @@
     expect(rendered).to have_unchecked_field("pages_question_input[is_optional]")
   end
 
+  context "when feature repeatable page is enabled", :feature_repeatable_page_enabled do
+    it "has an radio input for repeatable" do
```
jamie-o-wilkinson, line 44:
```suggestion
    it "has a radio input for repeatable" do
```

jamie-o-wilkinson (commented):
A couple of minor comments but otherwise looks good to me!

config/locales/en.yml
lfdebrux: @jamie-o-wilkinson this content is what was agreed with our content designer, see https://app.mural.co/t/gaap0347/m/gaap0347/1687946839058/9e1b89cd76af287c05c82df1cdb307a830324315?wid=0-1723651460259

thomasiles: This is a good spot.

It's hard to link to, but this is one of the options Charlotte considered for the content of this page. You can see all the options in [the middle, bottom of this mural page](https://app.mural.co/t/gaap0347/m/gaap0347/1687946839058/9e1b89cd76af287c05c82df1cdb307a830324315?wid=237-1717167377834).

Basically, hint text for the option was considered and rejected. It's possible that could change in the future though as this will be tested with users soon.

lfdebrux (approved):

PR #1395, opened 2024-08-22
Add rake task for changing groups organisation

thomasiles:
### WIP: changing groups organisations

Sometimes groups need to move from on organisation to another.

Currently, this would have to be done through the rails console, or in
production through the DB.

This commit adds a new rake tasks to change the organisation of one or
more groups.

A new namespace, groups, has been added. This contains the tasks
change_organisation and change_organisation_dry_run.

This layout is supposed to match the other tasks we have.

To invoke the task from the command line, use:

```
rake groups:change_organisation_dry_run[VcyzrC3R,2]
```

The task takes one or more groups external_ids and must end with
the id of the target organisation.

Trello card: 

### DIFF
--- a/lib/tasks/groups.rake
+++ b/lib/tasks/groups.rake
@@ -1,0 +1,61 @@
+namespace :groups do
+  desc "change the organisation of one or more groups"
+  task :change_organisation, [] => :environment do |_, args|
+    run_task("groups:change_organisation", args, rollback: false)
+  end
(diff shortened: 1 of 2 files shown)


thomasiles:
I think this will be done better here: https://github.com/alphagov/forms-admin/pull/1394

I'm throwing this over the wall in case it's useful.

PR #1399, opened 2024-08-23
Remove Form changes when changing group org

thomasiles:
# Fix the change group organisation task

The rake task to move groups between organisations included updating each form in the group to change the organisation_id held in the API.

Running the task in production failed when the Form did not have an organisation set. This happened because trial groups don't have an organisation_id set.

The task could either be changed to allow for Forms which don't have an organisation_id, or we could not change the Forms themselves.

We no longer use the organisation_id field. The organisation of a form is based only on the group it belongs to.

Because of this, I think its better to remove the code and tests, which are simpler as a result.

This PR changes the task to not alter the Form organisation_id.

### DIFF
--- a/lib/tasks/groups.rake
+++ b/lib/tasks/groups.rake
@@ -35,7 +35,7 @@ def run_task(task_name, args, rollback:)
   abort usage_message if group_ids.blank? || org_id.blank?
 
   ActiveRecord::Base.transaction do
-    change_organisation(group_ids, org_id, task_name:, rollback:)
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):
Agreed that this seems fine to do as long as we mark the `organisation_id` attribute on forms in forms-api as deprecated.

We're already not keeping this field in sync for forms that were previously made by trial users but have since been put into groups belonging to organisations, so can't rely on this field anyway.

PR #1402, opened 2024-08-28
Update payment_link.yml

StephenGill:
Small microcopy change: updates typo in example for payment links (currently example link wouldn't validate, because it's missing the "www.").

Replaces example in error message, so it's the same as that used in the on-page example - for consistency.

### DIFF
--- a/config/locales/input_objects/payment_link.yml
+++ b/config/locales/input_objects/payment_link.yml
@@ -6,11 +6,11 @@ en:
     hint:
       forms_payment_link_input:
         payment_url: |-
-          For example, https://gov.uk/payments/your-payment-link
+          For example, https://www.gov.uk/payments/your-payment-link
   activemodel:
     errors:
       models:
         forms/payment_link_input:
           attributes:
             payment_url:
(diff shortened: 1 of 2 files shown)


lfdebrux:
> We should change the validation of the field to not require `www.`, URLs don't have to have that part and the example on the GOV.UK Pay website don't have it.

Ah, turns out GOV.UK Pay does show the `www.` when a user creates a payment link:

Screenshot of web browser  Set payment link information - GOV.UK Pay  Set payment link information Title Briefly describe what the user is paying for. For example, “Pay for a parking permit”. This will also be your website address. Do not include names or personal information. laurence-test The website address for this payment link will look like: https://www.gov.uk/payments/laurence-test-service/laurence-test  Details (optional) Give your users more information. For example, you could tell them how long it takes for their application to be processed.

I still think we should change the validation, but we might as well update the hint text to match the example on Pay.

The other microcopy change looks good too.

hannahkc (approved):

lfdebrux:
@SamJamCul has fixed the validation in https://github.com/alphagov/forms-admin/pull/1402 🙌

PR #1406, opened 2024-08-29
Add terms of use accepted page

jamie-o-wilkinson:

Trello card: https://trello.com/c/xdOq34ui

This PR adds a terms of use input object, view, and controller. It doesn't add the full terms of use content itself, as this is yet to be finalised. 

It also doesn't extend the account check before_action to prompt a user to agree to the terms of use before continuing use of forms-admin. This will be added in a subsequent PR.

The reason to merge this before content is finalised and before enforcing agreement, is to allow the E2E test users to agree to the terms of use so that the E2E aren't broken when the enforcement is introduced.

### DIFF
--- a/spec/input_objects/account/terms_of_use_input_spec.rb
+++ b/spec/input_objects/account/terms_of_use_input_spec.rb
@@ -1,0 +1,56 @@
+require "rails_helper"
+
+describe Account::TermsOfUseInput do
+  include ActiveSupport::Testing::TimeHelpers
+
+  subject(:terms_of_use_input) { described_class.new(user:) }
+
+  let(:user) { create(:user) }
+
+  describe "validations" do
+    it "is valid when the user has agreed" do
+      terms_of_use_input.agreed = true
+      expect(terms_of_use_input).to be_valid
+    end
+
+    it "is invalid when the user has not agreed" do
+      terms_of_use_input.agreed = false
+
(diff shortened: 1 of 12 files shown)


spec/input_objects/account/terms_of_use_input_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  let(:user) { create(:user) }
+
+  describe "validations" do
+    it "is valid with an agreed attribute" do
+      terms_of_use_input.agreed = true
+      expect(terms_of_use_input).to be_valid
+    end
+
+    it "is invalid without an agreed attribute" do
```
DavidBiddle, line 16: I don't think this test name quite fits what it does - 'without an agreed attribute' makes me expect `terms_of_use_input.agreed` to be `nil` or `undefined`.

```suggestion
    it "is invalid when the user has not agreed" do
```

```suggestion
    it "is invalid when the agreed attribute is false" do
```

(This is nitpicky so don't let it block you if you don't agree)

DavidBiddle (dismissed):
Looks good and works locally 🎉 just one minor comment from me about a test name.

spec/input_objects/account/terms_of_use_input_spec.rb
jamie-o-wilkinson: Good suggestion, I've pushed up a change for this.

Thanks!

PR #1423, opened 2024-09-06
Remove usages of form.organisation_id

stephencdaly:

Trello card: https://trello.com/c/ZHGNp5Lk/1774-remove-organisationid-from-forms-api

We no longer rely on the organisation_id of the form in any production code in admin. We have also stopped setting this column for forms in some cases since moving to the groups model, meaning the data in the column cannot be relied upon. This PR cleans up any remaining usages in the code to allow for the column to be dropped from the forms table in forms-api.


thomasiles (approved):
great work 👏

PR #1453, opened 2024-09-19
Ensure single option selection questions can't be repeatable

SamJamCul:

Trello card: https://trello.com/c/SXihoLCc/1745-5-stop-form-creators-from-trying-to-add-routing-to-a-repeatable-question

In order to ensure that an invalid route can't be set up with a repeatable selection question, this change means that if a question is changed to a selection/only-one-option combination, it will be forced to not be repeatable. The option for changing one of these questions to repeatable will not be available either.

Are you able to make a route based on a repeatable question? You shouldn't be able to if this change is working!

### DIFF
--- a/app/views/pages/_form.html.erb
+++ b/app/views/pages/_form.html.erb
@@ -35,8 +35,12 @@
   <% end %>
 
   <% if Settings.features.repeatable_page_enabled %>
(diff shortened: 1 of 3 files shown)


thomasiles (approved):
neat idea to do it as part the template 👍

PR #1473, opened 2024-09-26
Change 'optional tasks' to 'optional task'

hannahkc:
This changes the 'Optional tasks' headings in the task lists to 'Optional task'. Because there is only one optional task in both places this is used so it should not be plural.

Trello card: None

- have I missed any tests or anything else I might not have considered?


DavidBiddle (approved):
Good change and no issues locally 🎉

PR #1479, opened 2024-10-01
Add another answer feature report

SamJamCul:
Adds the live_forms_with_add_another_answer report to the list of feature reports. This counts the number of live forms that have at least one question that is repeatable (which at the moment means a single repeatable question)

Trello card: https://trello.com/c/a2bMFMxU/1827-5-report-when-forms-are-using-add-another-answer-single-qn

Adds the 'Live forms with add another answer' row to the features report page. 

image

This is the first of two parts to this, and only counts the feature usage within live forms.


DavidBiddle (approved):
Code looks good and can see the new item in my local features report 🎉

PR #1483, opened 2024-10-04
Make tweaks to frontend build

DavidBiddle:

Trello card: None

Makes a few smallish changes to the frontend build:

- Enable vite-ruby autobuild in development - will automatically build the frontend if any of the files in app/frontend change. We already have this on in test, it'll make developers' lives easier to have it on for local dev too.
- Remove custom linting plugin - I don't think we were really getting much from this and remoiving it speeds up the dev server startup and hot module reloads. If we find we have a need for linting on every update, we should probably just use [vite-plugin-eslint](https://github.com/gxmari007/vite-plugin-eslint) which is a more sophisticated implementation.
- Move markdown editor icons into app/frontend icons folder - these are tied to a JS enhancement so it makes sense for them to be in the frontend folder.


config/vite.json
```diff
@@ -15,7 +15,8 @@
     "autoBuild": false
   },
   "development": {
-    "port": 3036
+    "port": 3036,
+    "autoBuild": true
```
lfdebrux, line 19: 🙌

lfdebrux (approved):
Looks good to me! Nice changes 👍

PR #1485, opened 2024-10-07
Exclude CSV task from status count

DavidBiddle:

Trello card: 

Excludes the 'Get completed forms as CSV files' task from the count of completed/uncompleted tasks, since we do that for the other optional task.


stephencdaly (approved):
Looks good to me!

PR #1486, opened 2024-10-07
Temporarily exclude task to share preview from task count

stephencdaly:

We are adding a new mandatory task to share a preview of a draft form.

We need to add a task status to the form in forms-api before we add/ update the pages for this in forms-admin. Temporarily exclude this task from the counts for the `You've completed X out of X tasks` content on the task list page.

We can stop excluding this task from the counts at the same time we add the page for the new task and add it to the task list.

There is a draft PR to add the new task status to forms-api, which will only be merged after this PR has been merged https://github.com/alphagov/forms-api/pull/592


DavidBiddle (approved):
Looks good to me and the count is unchanged on my local machine 🎉

PR #1491, opened 2024-10-08
Fix Sass 'mixed declarations' warning

DavidBiddle:

Trello card: https://trello.com/c/m5hB6sVM/2088-resolve-sass-mixed-declarations-deprecation-warning

Fixes a [deprecation warning](https://sass-lang.com/documentation/breaking-changes/mixed-decls/) related to the way that the sass declarations are ordered. In v2, the way these  declarations are turned into CSS will change and if they're left as they are they'll generate more CSS than is necessary.

There shouldn't be any visual differences as a result of these changes. 

### DIFF
--- a/app/components/metrics_summary_component/_index.scss
+++ b/app/components/metrics_summary_component/_index.scss
@@ -9,11 +9,11 @@
 }
 
 .app-metrics__big-number {
-  @include govuk-font-size(19);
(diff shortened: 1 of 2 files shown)


lfdebrux (approved):
Yay for no warnings! 🙌 

And thank you so much for figuring this out.

PR #1494, opened 2024-10-09
Update MOU agreed banner

hannahkc:

I've changed the use of 'signature' in this banner to 'agreed' for consistency with how we're referring to agreeing to the MOU elsewhere. 

I have also changed the date format for style - I'm hoping this will show the date in the format: 'day number with no leading 0' 'full month as a word' 'year'. For example: 9 October 2024

I very much might have got this wrong! 

Trello card: none

- Did Hannah break anything?
- Does the date formatting work as Hannah expected? 


DavidBiddle (approved):
Looks good to me :tada:

Screenshot 2024-10-09 at 16 55 49

PR #1498, opened 2024-10-10
Use curly quotes and apostrophes in translations

stephencdaly:

Fix places in the translations where we were using straight quotes/apostrophes rather than the curly ones.


hannahkc (approved):

hannahkc:
You're a hero.

PR #1500, opened 2024-10-10
Refactor MarkCompleteInput for re-use

stephencdaly:

We're adding a new task to share a preview of a draft form, which will have radio buttons to mark the task as complete. This will need to re-use the MarkCompleteComponent.

Extract the common fields and methods from MarkCompleteInput into a base class, and add MarkPagesSectionCompleteInput which extends this base class and has the attributes and methods that apply specifically  to the task to mark the pages section as complete.

Refactor DeclarationInput to extend this new base class, as we also pass this input object into MarkCompleteController.


app/controllers/forms_controller.rb
```diff
@@ -30,6 +30,6 @@ def mark_pages_section_completed
 private
 
   def mark_complete_input_params
-    params.require(:forms_mark_complete_input).permit(:mark_complete).merge(form: current_form)
+    params.require(:forms_mark_pages_section_complete_input).permit(:mark_complete).merge(form: current_form)
```
DavidBiddle, line 33: This line wasn't updated initially, so marking the pages section complete threw an error

DavidBiddle (approved):

PR #1503, opened 2024-10-11
Prevent browser default behaviour when copy button is clicked

DavidBiddle:

Trello card: 

In the copy to clipboard javascript, we create a button and attach a click event handler to it which writes to the user's clipboard. However, this button also has default behaviour defined by the browser.

One of those default button behaviours is that when the button is inside a form, clicking it will submit the form. We don't want the clipboard button to do this, so this PR adds `event.preventDefault` to the click event handler to turn this behaviour off.


thomasiles (approved):
nice test

PR #1504, opened 2024-10-11
Add page for share preview task

stephencdaly:

Add the page for the the new "Share a preview of your draft form" task.

This page uses the `MarkCompleteComponent` for the radio buttons to mark the task as complete. A new input object, `SharePreviewInput`, which extends the `MarkCompleteInput` base class has been added to pass the HTML form data between the view and controller.

We've refactored the `FormURLComponent` so this can be re-used to display the link to the preview with a button to copy it to the clipboard.

This has not yet been added to the task list, and will be added in a subsequent PR.

Screenshot 2024-10-11 at 12 10 21

Screenshot 2024-10-11 at 12 10 54

Screenshot 2024-10-11 at 13 08 06

### DIFF
--- a/app/components/form_url_component/view.html.erb
+++ b/app/components/form_url_component/view.html.erb
@@ -1,6 +1,6 @@
-<%= tag.div "data-module": "copy-to-clipboard", "data-copy-button-text": t("form_url_component.copy_to_clipboard") do %>
+<%= tag.div "data-module": "copy-to-clipboard", "data-copy-button-text": @button_text do %>
   <h2 class="govuk-heading-m">
-    <%= t("form_url_component.form_url") %>
+    <%= @heading_text %>
   </h2>
 
   <p data-copy-target>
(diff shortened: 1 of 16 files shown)


app/components/mark_complete_component/view.html.erb
```diff
@@ -1,4 +1,5 @@
 <% if @generate_form %>
+  <!-- TODO: find out which test is passing in nil -->
```
DavidBiddle, line 2: https://github.com/alphagov/forms-admin/blob/a2a64e3ffc494f069b0429bdddef831e90947b36/spec/components/mark_complete_component/view_spec.rb#L12-L26

stephencdaly: Hmm I can't remember why this was a problem

DavidBiddle (dismissed):

DavidBiddle (approved):
Code looks good and can see it working locally 🎉

PR #1506, opened 2024-10-11
Show an ordered list of forms from a form creator's groups

chao-xian:
## What problem does this pull request solve?

Trello card: https://trello.com/c/0ObcUSir

The list of forms was appearing as they are returned by the API which doesn't make sense to the user. This PR applies alphabetical sorting.

## Before
localhost_3000_groups_NBgwE3D4

## After
localhost_3000_groups_NBgwE3D4 (1)

### DIFF
--- a/app/presenters/form_list_presenter.rb
+++ b/app/presenters/form_list_presenter.rb
@@ -43,7 +43,7 @@ private
   end
 
   def rows
-    forms.sort_by(&:name).map do |form|
+    forms.sort_by { |form| [form.name.downcase, form.created_at] }.map do |form|
       [{ text: form_name_link(form) },
        { text: find_creator_name(form) },
        { text: form_status_tags(form), numeric: true }].compact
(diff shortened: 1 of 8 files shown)


app/presenters/form_list_presenter.rb
```diff
@@ -43,7 +43,7 @@ def head
   end
 
   def rows
-    forms.map do |form|
+    forms.sort_by(&:name).map do |form|
```
lfdebrux, line 46: What happens if there is more than one form with the same name?

lfdebrux (approved):
Thanks for doing this, this is a really nice example of what idiomatic Rails code should look like; I especially appreciate the detailed commit messages.

I have one comment/question, but otherwise I think this is good to go after a rebase.

app/presenters/form_list_presenter.rb
alice-carr: I've changed this so that now if two forms have the same name, we then sort on `created_at`

lfdebrux (commented):
Thanks for sorting the sorting @alice-carr! 😁 

One more thing, I can see there's a merge commit in the branch now, I think we prefer to rebase branches on main before merging so each PR has only one merge commit in the main branch.

Are you happy to do the rebase?

lfdebrux (approved):

PR #1513, opened 2024-10-14
Add make live validation for share preview status

DavidBiddle:

Trello card: https://trello.com/c/QuqK5jmG/1867-add-new-task-to-share-a-preview-of-your-draft-form-for-drafts-of-new-live-and-archived-forms#comment-670921da24e1e3431e7b696c

Adds a validation message when the user tries to make the form live without completing the share preview task first.

This should be merged before https://github.com/alphagov/forms-api/pull/599 or users will potentially see an unfriendly error message.


stephencdaly (approved):
Have tested locally, and this is compatible with the changes in https://github.com/alphagov/forms-api/pull/599

PR #1517, opened 2024-10-15
Update govuk-frontend ruby/node packages

thomasiles:
A new version of govuk-frontend has been released, and we need to update our ruby/node packages to match.

It's better to update the packages in one go, rather than updating them one by one. This will ensure that we don't have any issues with missing assets or styles.

Update to the latest:
- govuk-frontend 5.7.1,
- govuk-components 5.7.0,
- govuk_design_system_formbuilder 5.7.0

This includes a new crest logo in the footer.

When this is merged, the dependabot prs can be closed:
- https://github.com/alphagov/forms-admin/pull/1507
- https://github.com/alphagov/forms-admin/pull/1510
- https://github.com/alphagov/forms-admin/pull/1511

Trello card: 


DavidBiddle (approved):
Code looks good, and I can see the new crest when running locally 🎉

PR #1522, opened 2024-10-17
Log actual user when super admin is acting as another user

lfdebrux:

We want to make sure that when the "Act as user" feature is used that our logs don't confuse the actions taken by the actual user with something done by the user they are acting as.

This is especially important for some operations such as signing an MoU, as it could have legal ramifications if our logs make it look like a person has done something that was actually done by a super admin.

This commit changes our logging so that if a super admin is acting as another user we log the super admin's details.


stephencdaly (approved):
Great work! Well tested, and I've tested locally and it works as expected.

PR #1530, opened 2024-10-21
Add rake task to remove an empty group

thomasiles:
# Delete two trial groups made in error

Trello card: https://trello.com/c/uwpvcpyz/1889-delete-2-trial-groups-from-dvsa-made-in-error

When users create groups they don't want, we need a way to remove them
from the organisation and remove them. This can happen if a user selects
the wrong organisation at login.

This commit adds a simple rake task to remove a group when given a group
external id.

The group must not contain any forms. This avoids any complications
about what should happen to those forms.

To remove a group which has forms, remove the forms first.

### DIFF
--- a/lib/tasks/groups.rake
+++ b/lib/tasks/groups.rake
@@ -26,6 +26,24 @@ namespace :groups do
 
     run_bulk_task(task_name:, source_organisation_id:, target_organisation_id:, rollback: true)
   end
+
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):
Looks good! Good test coverage and I've tested locally too.

lfdebrux (approved):
Nice!

PR #1534, opened 2024-10-23
Add report page for all forms with add another answer

SamJamCul:
This adds a new route to the reports controller, which leads to the full report for all forms that have a repeatable question. This report shows the form with a link to that form, and has a row for each question within the form that's currently set to be repeatable.

Trello card: https://trello.com/c/a2bMFMxU/1827-5-report-when-forms-are-using-add-another-answer-single-qn

This adds a new report to the reports controller, which leads to the full list of all forms with add another answer. This is contingent on https://github.com/alphagov/forms-api/pull/615, which sets up the report from the API's side. 

The report lists all form names with a link to that forms' edit path - and the question text for each question in that form as a separate row. 

image

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -20,4 +20,12 @@ class ReportsController < ApplicationController
 
     render locals: { data: }
   end
+
+  def add_another_answer
+    authorize Report, :can_view_reports?
+
+    data = Report.find("features")
+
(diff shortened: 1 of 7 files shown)


hannahkc:
I was going to suggest changing the capitalisation of the labels to sentence case ('Form name' and 'Question text') but looks like that's already been done! So content looks grand. 👍

thomasiles (approved):
🏅

PR #1539, opened 2024-10-25
Add form state column to add another answer report

SamJamCul:
Adding the state column to the report containing all forms with add another answer questions. This is to make navigating the report a bit easier. The forms are ordered by state, so they're grouped up more sensibly.

Trello card: https://trello.com/c/a2bMFMxU/1827-report-when-forms-are-using-add-another-answer-single-qn

Adds the state column to the add another answer report. It's ordered by state, so the rows are now grouped by states. 

image


stephencdaly (approved):
Looks good and works as expected

PR #1545, opened 2024-10-29
Add text box for selection options

DavidBiddle:

Trello card: https://trello.com/c/0P1FpI7w/1934-productionizizzle-the-create-a-long-list-autocomplete-page

Adds:
- a new page with a text box which allows users to configure multiple selection options at once.
- a link to that page from the selection options page (this page is only visible to people in groups with the feature enabled)

### Screenshots
Screenshot of the selection options page, showing a new 'Enter all options into a textbox' link next to the 'Add another' button.

Screenshot of the new bulk selection options page. It has explanatory text at the top, a text box for entering the options, and a yes/no radio button question for whether the page should include 'none of the above'

### Testing notes
To test this locally, you'll need to turn the `long_lists_enabled` setting on for a group, and edit a form within that group. The easiest way to do this is with the [new rake task](https://github.com/alphagov/forms-admin/pull/1556):
```zsh
bin/rake "groups:enable_long_lists[group_id]"
```


app/input_objects/pages/long_lists_selection/bulk_options_input.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  attr_accessor :include_none_of_the_above, :draft_question, :bulk_selection_options
+
+  validates :draft_question, presence: true
+  validates :include_none_of_the_above, inclusion: { in: %w[true false] }
+  validate :bulk_selection_options, :validate_selection_options
+
+  def assign_form_values
+    self.bulk_selection_options = draft_question.answer_settings[:selection_options].map { |option| option[:name] }.join("\n")
+    self.include_none_of_the_above = draft_question.is_optional
```
stephencdaly, line 14: Nice, this is a good way of initialising the input. I'll update the corresponding `options_input` controller to use this pattern.

app/input_objects/pages/long_lists_selection/bulk_options_input.rb
```diff
@@ -0,0 +1,49 @@
+class Pages::LongListsSelection::BulkOptionsInput < BaseInput
+  DEFAULT_OPTIONS = { selection_options: [],
```
stephencdaly, line 2: I don't think this is used anywhere, we could get rid of it.

app/input_objects/pages/long_lists_selection/bulk_options_input.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+private
+
+  def validate_selection_options
+    unique_options = unique_selection_options
+
+    return errors.add(:bulk_selection_options, :minimum) if unique_options.length < 2
+
+    errors.add(:bulk_selection_options, :maximum) if unique_options.length > 1000
```
stephencdaly, line 39: I think we'll need to apply the same rules as in `options_input` to decide the maximum based on whether `only_one_option` is true or not. This could be done in a subsequent PR.

config/locales/en.yml
```diff
@@ -179,6 +179,11 @@ en:
       route_deleted: Question %{question_position}’s route has been deleted
       route_updated: Question %{question_position}’s route has been updated
       title: Success
+  bulk_options:
+    hint: List your options, with each option on a new line.
+    label: Add options to your list
+    longer_than_30_options: If your list is longer than 30 options people will be able to search and find the options by typing in an autocomplete box.
+    select_one_option: People will be able to select one option from a list.
```
stephencdaly, line 186: I went with "your" list on the other page - the designs were unclear which it should be, but this can be picked up in the content sweep.

I wonder whether we should centralise the shared content in one place, but we can consider that later.

spec/input_objects/pages/long_lists_selection/bulk_options_input_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    it "is invalid if more than 1000 Bulk selection options are provided" do
+      bulk_options_input.bulk_selection_options = (1..1001).to_a.join("\n")
+      error_message = I18n.t("activemodel.errors.models.pages/long_lists_selection/bulk_options_input.attributes.bulk_selection_options.maximum")
+      expect(bulk_options_input).not_to be_valid
+
+      expect(bulk_options_input.errors.full_messages_for(:bulk_selection_options)).to include("Bulk selection options #{error_message}")
+    end
+
+    it "is valid if there are between 2 and 20 unique selection values" do
```
stephencdaly, line 28: `between 2 and 1000`

app/input_objects/pages/long_lists_selection/bulk_options_input.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def unique_selection_options
+    bulk_selection_options.split(/\n/).map(&:strip).compact_blank.uniq
+  end
+
+  def answer_settings
+    { only_one_option: "true", selection_options: unique_selection_options.map { |option| { name: option } } }
```
stephencdaly, line 47: This should make sure it uses the value of `only_one_option` that already exists on the `answer_settings`

spec/input_objects/pages/long_lists_selection/bulk_options_input_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      expect(bulk_options_input.submit).to be_falsey
+    end
+
+    it "sets draft_question answer_settings and is_optional" do
+      bulk_options_input.bulk_selection_options = (1..2).to_a.join("\n")
+      bulk_options_input.include_none_of_the_above = "true"
+      bulk_options_input.submit
+
+      expected_settings = {
+        only_one_option: "true",
```
stephencdaly, line 71: We should probably have a test that makes sure the value of `only_one_option` that already exists on the `answer_settings` is not overwritten.

app/input_objects/pages/long_lists_selection/bulk_options_input.rb
```diff
@@ -0,0 +1,49 @@
...
+  def validate_selection_options
+    unique_options = unique_selection_options
+
+    return errors.add(:bulk_selection_options, :minimum) if unique_options.length < 2
+
+    errors.add(:bulk_selection_options, :maximum) if unique_options.length > 1000
+  end
+
+  def unique_selection_options
+    bulk_selection_options.split(/\n/).map(&:strip).compact_blank.uniq
```
stephencdaly, line 43: Is this the solution we decided for this? To strip duplicates with no error?

stephencdaly (commented):
Nice work, looks really good. I've left some comments - some are just observations that we don't need to do anything about right now but there are a few that might need addressing or a discussion about.

app/input_objects/pages/long_lists_selection/bulk_options_input.rb
DavidBiddle: Done - copied your neat `answer_settings.merge` solution 😄

spec/input_objects/pages/long_lists_selection/bulk_options_input_spec.rb
DavidBiddle: Good spot! Have fixed

config/locales/en.yml
DavidBiddle: I've updated to 'your list' for consistency, let's see what the content sweep brings

app/input_objects/pages/long_lists_selection/bulk_options_input.rb
DavidBiddle: I've added this in the current PR now because it was fairly simple to do something based on your implementation. There's probably some refactoring we can do to share the logic but for now I think it's probably OK to have a bit of repetition since we'll be coming back to this later

app/input_objects/pages/long_lists_selection/bulk_options_input.rb
DavidBiddle: Done!

spec/input_objects/pages/long_lists_selection/bulk_options_input_spec.rb
DavidBiddle: Done!

app/input_objects/pages/long_lists_selection/bulk_options_input.rb
DavidBiddle: Really good spot - we decided on showing a validation message but I missed this and stuck with what was in the initial exploration branch. I've updated to validate for duplicates instead. 

Worth flagging that we don't appear to have content for this validation message so I've put in a placeholder one that we can update after the content sweep.

stephencdaly (approved):
All my comments have been addressed. I've tested locally and it all works as expected.

PR #1546, opened 2024-10-29
Add workflow for long lists

stephencdaly:

Trello card: https://trello.com/c/ufPgMvic

Add the new workflow for creating select from a list of options question with the ability to add more than 30 options when "People can choose only one option" is selected.

This new journey is only shown when the group the form is in has a new `long_lists_enabled` database flag set to true.

The new journey has 4 pages:
1. A page asking for the question text - this is unchanged from the existing journey
2. A page with radios to ask whether people can choose only one option or more than one option
3. A page to add the selection options and choose whether an option for "None of the above" is shown. The number of options the user can add depends on whether they chose "only one option" or "more than one option". For the former, the limit is 1000 and for the latter it is 30.

The journey doesn't currently consider differences in content between when you are adding a new question or editing an existing question. We think some titles will be different, but this will be updated in a subsequent PR to make any content updates.

### Screenshots of the new journey

Screenshot 2024-10-31 at 15 34 45

When "People can choose only one option is picked":

Screenshot 2024-10-31 at 15 45 32

When "People can choose more than one option is picked":

Screenshot 2024-10-31 at 15 45 24

### DIFF
--- a/app/components/page_settings_summary_component/view.rb
+++ b/app/components/page_settings_summary_component/view.rb
@@ -3,9 +3,10 @@
 module PageSettingsSummaryComponent
   class View < ViewComponent::Base
     include Rails.application.routes.url_helpers
-    def initialize(draft_question)
+    def initialize(draft_question:, long_lists_enabled: false)
       super
       @draft_question = draft_question
+      @long_lists_enabled = long_lists_enabled
     end
 
     def before_render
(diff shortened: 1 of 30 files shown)


app/components/page_settings_summary_component/view.rb
```diff
@@ -3,9 +3,10 @@
 module PageSettingsSummaryComponent
   class View < ViewComponent::Base
     include Rails.application.routes.url_helpers
-    def initialize(draft_question)
+    def initialize(draft_question:, long_lists_enabled:)
```
DavidBiddle, line 6: We could default this to false to save needing to edit the calls to this component in `page_settings_summary_component_preview.rb`:

```suggestion
    def initialize(draft_question:, long_lists_enabled: false)
```

DavidBiddle (dismissed):
Made one suggestion but looks great and works on my machine 🎉

app/components/page_settings_summary_component/view.rb
stephencdaly: Good idea, this will make the tidy-up easier when we remove the feature flag. Have updated

DavidBiddle (approved):

PR #1552, opened 2024-10-31
Record time when users last signed in

lfdebrux:

Trello card: https://trello.com/c/8WNltVGz/1917-find-out-which-accounts-have-been-inactive-for-over-a-year 

We want to be able to find out when users haven't signed in to our service for a while. This PR will make it easier in future, by starting to save the time when users sign in to the database.

I think that we don’t need the information to be that granular, so I propose we update the time `last_signed_in_at` on the user record whenever they authenticate. As our session policy requires form creators to sign in every 20 hours and we're most likely interested in users who have been inactive for 12 months or more this seems like a reasonable level of resolution to me, and requires much fewer writes to the database.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -105,6 +105,10 @@ class User < ApplicationRecord
     Settings.analytics_enabled && !super_admin?
   end
 
+  def signed_in!
+    update!(last_signed_in_at: Time.zone.now)
+  end
+
 private
 
   def requires_name?
--- a/config/initializers/authentication.rb
+++ b/config/initializers/authentication.rb
@@ -51,4 +51,6 @@ Warden::Manager.after_authentication do |user, auth, _opts|
   if user.provider == "auth0"
     auth.session["auth0_connection_strategy"] = auth.env["omniauth.auth"][:extra][:raw_info][:auth0_connection_strategy]
   end
+
+  user.signed_in!
 end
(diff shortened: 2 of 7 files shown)


app/models/user.rb
```diff
@@ -105,6 +105,10 @@ def collect_analytics?
     Settings.analytics_enabled && !super_admin?
   end
 
+  def signed_in!
+    update(last_signed_in_at: Time.zone.now)
```
stephencdaly, line 109: Does this need to be `update!` or is there some magic that means this gets saved?

stephencdaly (commented):
Is there a test we can add somewhere to make sure the `last_signed_in_at` is updated during authentication?

app/models/user.rb
lfdebrux: Ah, good point, it should be `update!`. Both `update` and `update!` save the changes, the difference is that `update!` raises an exception if there is a failure in saving while `update` just returns false if there is an issue.

lfdebrux: (from https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update)

lfdebrux: Fixed now

stephencdaly: Is there a test we can add to for this? Maybe `authentication_controller_spec`?

lfdebrux:
> Is there a test we can add somewhere to make sure the `last_signed_in_at` is updated during authentication?

Good point about tests; I've added some now.

stephencdaly (approved):
Nice work! I'm not sure I can test this locally as I'm not sure I can test with athu0, but I assume you've already tested this or have a plan to verify it works as expected.

lfdebrux:
> Nice work! I'm not sure I can test this locally as I'm not sure I can test with athu0, but I assume you've already tested this or have a plan to verify it works as expected.

I've tested this in dev and it worked 👍

PR #1556, opened 2024-11-01
Add rake tasks to enable/disable long lists enabled

stephencdaly:

Adds rake tasks to enable/disable the `long_lists_enabled` flag for a group.

Tests haven't been added for these tasks as the logic is simple, and this will only need to be used a few times.


DavidBiddle (approved):
Code looks good and both new tasks work for me locally 🎉

PR #1557, opened 2024-11-01
Ensure backwards compatibility for only_one_option

stephencdaly:

The only_one_option setting for selection questions was previously set by a checkbox where the unchecked value was "0". Either "0" or "false" work with runner, but we think it's neater for the radio buttons for the new journey to create a selection question to have values of "true" or "false".

This change ensures that existing questions have the correct option selected on the question to decide whether "People can choose only one option" for the new journey to create a selection question.

This code to do the conversion will need to be kept unless we migrate existing questions after enabling the new journey to create selection questions for all forms.


DavidBiddle (approved):
Works for me locally and code looks good 🎉

PR #1559, opened 2024-11-04
Include subfolder locales in checks

DavidBiddle:

Trello card: 

We had the `i18n-tasks` gem configured so that it only looked at files in the top level of `config/locales`. We have quite a lot of locale files in subfolders of `config/locales`, which weren't covered by these checks.

This PR:
- updates the config to include the subfolders in the checks
- normalizes the files with `i18n-tasks normalize`

### DIFF
--- a/config/locales/input_objects/confirm_action.yml
+++ b/config/locales/input_objects/confirm_action.yml
@@ -1,7 +1,8 @@
+---
 en:
   helpers:
     label:
       confirm_action_input:
         options:
-          yes: "Yes"
-          no: "No"
+          'no': 'No'
+          'yes': 'Yes'
--- a/config/locales/pages/guidance.yml
+++ b/config/locales/pages/guidance.yml
@@ -1,27 +1,35 @@
+---
 en:
-  helpers:
-    hint:
-      pages_guidance_input:
-        page_heading: When you add guidance your question text will no longer be the main page heading, so you need to use a different one. Use a heading that’s a statement rather than a question - for example, ‘Interview needs’.
-        guidance_markdown: Use Markdown if you need to format your guidance content. Formatting help can be found below.
-    label:
-      pages_guidance_input:
-        page_heading: Give your page a heading
-        guidance_markdown: Add guidance text
   activemodel:
     errors:
       models:
         pages/guidance_input:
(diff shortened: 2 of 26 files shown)


config/locales/input_objects/confirm_action.yml
```diff
@@ -1,7 +1,8 @@
+---
 en:
   helpers:
     label:
       confirm_action_input:
         options:
-          yes: "Yes"
-          no: "No"
+          'false': 'No'
+          'true': 'Yes'
```
lfdebrux, line 8: I'm not sure about this change 🤔

config/locales/pages/guidance.yml
```diff
@@ -1,27 +1,35 @@
...
               too_long: Guidance text must be 5,000 characters or less
+              unsupported_markdown_syntax: Guidance text can only contain formatting for links, subheadings (##), bulleted lists (*), or numbered lists (1.)
+            page_heading:
+              blank: Enter a page heading
+              too_long: Page heading must be %{count} characters or less
   activerecord:
     errors:
       models:
         draft_question:
-          attributes: *guidance_validation
```
lfdebrux, line 27: Ooh, `i18n-tool` doesn't like YAML references?

lfdebrux:
Good spot 😱

config/locales/input_objects/confirm_action.yml
DavidBiddle: It did - I think because YAML treats `yes` and `no` as equivalent to `true` and `false`. I think this is easier to read and still OK by i18n-tasks:
```
'no': 'No'
'yes': 'Yes'
```
so I'll update it

config/locales/pages/guidance.yml
DavidBiddle: I guess not! I suppose there's a logic to favouring keeping the files simple over reducing repetition in the locale files

config/locales/input_objects/confirm_action.yml
DavidBiddle: Done

lfdebrux (approved):
Thanks for spotting this and fixing it!

PR #1560, opened 2024-11-04
1924 add repository service to forms admin for form condition model

SamJamCul:

Trello card: https://trello.com/b/9cKfKKdR/govuk-forms-sprint-board

Adds a `ConditionRepository`, which is a new service which acts as a buffer between the app and calls to the Condition model. This is intended to add an abstraction layer between our ActiveResource usage and the rest of the app. 

There's also a refactor for the name of a test var referencing the ConditionsInput object.

### DIFF
--- a/app/controllers/pages/conditions_controller.rb
+++ b/app/controllers/pages/conditions_controller.rb
@@ -34,7 +34,7 @@ class Pages::ConditionsController < PagesController
   end
 
   def edit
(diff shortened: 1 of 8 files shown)


lfdebrux (approved):

chao-xian:
@SamJamCul the trello link is to the board and not the card

PR #1562, opened 2024-11-04
Redirect to correct version of selection options page

stephencdaly:

Trello card: https://trello.com/c/ufPgMvic/

We want to decide whether to show the version of the selection options page with fields per option, or a textbox to bulk enter all options depending on how many existing options the draft question has.

If there are no existing options, or less than or equal to 30 options we want to show the page with indivudual fields. If greater than 30, we want to show the page with the textbox.

We need to redirect to the correct version of the page when either redirecting from the "How do you need to collect the answer?" page or from the edit question page. Add helper methods to share the code between these 2 places.


DavidBiddle (approved):
Nice - PagesHelper is a neat solution and it works for me locally 🎉

PR #1564, opened 2024-11-05
1924 add repository service to forms admin for form condition model

SamJamCul:

Trello card: https://trello.com/c/6AKTMAKE/1924-add-repository-service-to-forms-admin-for-form-condition-model

Refactor of https://github.com/alphagov/forms-admin/pull/1560. The `Condition` objects created as part of the `save` and `destroy` methods now include a step where they also build the prefix options. These are necessary for ActiveResource for its API calls. 

Also slightly refactored the `find` method so that the object ids are all passed in as key/value arguments.

### DIFF
--- a/app/controllers/pages/conditions_controller.rb
+++ b/app/controllers/pages/conditions_controller.rb
@@ -34,7 +34,7 @@ class Pages::ConditionsController < PagesController
   end
 
   def edit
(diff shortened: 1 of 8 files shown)


lfdebrux (approved):
Thanks for fixing this!

PR #1567, opened 2024-11-05
Add feature test for long lists

DavidBiddle:

Adds scenarios for the 'add a question for each type of answer' spec which cover:
- adding checkbox and selection questions under the new flow when long lists are enabled
- adding a selection question with more than 30 options

### DIFF
--- a/spec/features/form/add_or_edit_questions/add_a_question_for_each_type_of_answer_spec.rb
+++ b/spec/features/form/add_or_edit_questions/add_a_question_for_each_type_of_answer_spec.rb
@@ -35,6 +35,31 @@ feature "Add/editing a single question", type: :feature do
         and_i_add_another_question
       end
     end
+
+    context "when the form is in a group which has long lists enabled" do
+      let(:group) { create(:group, organisation: standard_user.organisation, long_lists_enabled: true) }
+      let(:options) { 10.times.map { Faker::Name.unique.name } }
+
+      scenario "add a selection question which the user can answer more than once" do
+        when_i_am_viewing_an_existing_form
+        and_i_want_to_create_or_edit_a_page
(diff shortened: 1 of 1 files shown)


spec/features/form/add_or_edit_questions/add_a_question_for_each_type_of_answer_spec.rb
```diff
@@ -35,6 +35,45 @@
...
+        when_i_am_viewing_an_existing_form
+        and_i_want_to_create_or_edit_a_page
+        and_i_select_selection_type_question
+        and_i_select_people_can_only_choose_one_option
+        and_i_configure_the_selection_options(options)
+        and_i_click_save
+        and_i_add_another_question
+      end
+
+      context "when there are more than 30 options" do
```
stephencdaly, line 63: I wonder whether its necessary to have a feature test for this? Feature tests are slow to run, so should only cover the minimum happy path to ensure everything works together. I think we have enough unit tests to cover that we can add more than 30 options, unless I'm missing something else that this tests that isn't covered by the other 2.

stephencdaly (commented):
Looks good, just a question

spec/features/form/add_or_edit_questions/add_a_question_for_each_type_of_answer_spec.rb
DavidBiddle: Yeah I debated this and erred on the side of more tests. But you're right that this is covered elsewhere so I'll get rid!

DavidBiddle: Done

stephencdaly (approved):

PR #1571, opened 2024-11-06
Rename route card service

thomasiles:
# Rename RouteSummaryCardData Service to Presenter

Change the name of the `RouteSummaryCardData` service to `Presenter`.

This is to make the name more consistent with the other presenters and better reflects its purpose.


lfdebrux:
Looking at examples from blog posts and the like, it seems to me that a presenter is more commonly a class that wraps one or more models (see for instance [[1](https://www.rubyguides.com/2019/09/rails-patterns-presenter-service/), [2](https://medium.com/@steve.v.brewer/rails-presenters-for-readable-modular-and-testable-code-e63431360c66), [3](https://github.com/drapergem/draper)), which this class doesn't do... could we refactor it to be a [decorator](https://refactoring.guru/design-patterns/decorator/ruby/example) like those examples?

chao-xian:
Decorator pattern isn't quite what we want here (and often requires the [Draper gem](https://github.com/drapergem/draper)) and are usually for calling in the controller on models. We're using the presenter pattern for example directly calling in the view. This also follows how GOV.UK apply it widely.

chao-xian (approved):

PR #1572, opened 2024-11-06
Group vite and govuk-frontend dependencies

DavidBiddle:

Trello card: https://trello.com/c/3Z5nfY2a/56-investigate-dependabot-groups

Creates dependency groups for vite and govuk-frontend related dependencies.

These group together dependencies that are closely coupled to each other, where we want to avoid sequencing issues.

May also speed up updating our dependencies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,11 @@ updates:
     allow:
       - dependency-type: "direct"
       - dependency-name: "vite_ruby"
+    groups:
+      govuk_frontend:
+        patterns:
+          - "govuk-components"
+          - "govuk_design_system_formbuilder"
 
   # Maintain dependencies for npm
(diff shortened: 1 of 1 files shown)


DavidBiddle:
Tested in a fork and this doesn't work as originally intended - Dependabot doesn't allow you to group across different package managers.

That means that it isn't currently possible to group:

- `govuk-components` & `govuk_design_system_formbuilder` (bundler) with govuk-frontend (npm)
- `vite_rails` (bundler) with `vite` & `vite-plugin-ruby` (npm).

I think that makes this grouping approach less valuable. I'm happy to look at different grouping approaches, I think these would work:
- grouping dev dependencies for bundler and npm respectively
- grouping all patch versions for packages that use semantic versioning (i.e. not Rails)

PR #1574, opened 2024-11-06
Content changes to edit question page

stephencdaly:

Trello card: https://trello.com/c/aVhQN4uC/1935-update-the-edit-your-question-page-for-long-list-autocomplete

Update the content displayed in the summary table on the "Edit your question" for selection questions.

When the `long_lists_enabled` flag is turned on for the group, use the new content for displaying the answer to the question asking how many options people should be able to select on the edit question page. Keep the old content for the old journey when the `long_lists_enabled` flag is turned off for the group.

Also update the content on the page to answer this question with the latest iteration of content from the designs.

List the selection options as a bulleted list in the summary table on the "Edit your question" page with the number of options that were entered displayed. This change will apply regardless of whether the `long_lists_enabled` flag  is enabled.

Screenshot 2024-11-07 at 08 54 29

### DIFF
--- a/app/components/page_settings_summary_component/view.html.erb
+++ b/app/components/page_settings_summary_component/view.html.erb
@@ -1,73 +1,81 @@
 <%= govuk_summary_list do |summary_list| %>
-  <%= summary_list.with_row do |row|
-      row.with_key(text: t("page_settings_summary.answer_type"))
-      row.with_value(text: t("helpers.label.page.answer_type_options.names.#{@draft_question.answer_type}"))
-      row.with_action(text: t("page_settings_summary.change"),
-                 href: @change_answer_type_path,
-                 visually_hidden_text: t("page_settings_summary.answer_type")
-      ) end %>
+  <%= summary_list.with_row do |row| %>
+    <%= row.with_key(text: t("page_settings_summary.answer_type")) %>
+    <%= row.with_value(text: t("helpers.label.page.answer_type_options.names.#{@draft_question.answer_type}")) %>
+    <%= row.with_action(text: t("page_settings_summary.change"),
+                    href: @change_answer_type_path,
+                    visually_hidden_text: t("page_settings_summary.answer_type")) %>
+  <% end %>
 
   <% if show_selection_settings_summary %>
-    <%= summary_list.with_row do |row|
-          row.with_key(text: t("page_settings_summary.selection.options"))
-          row.with_value(text: show_selection_options)
-          row.with_action(text: t("page_settings_summary.change"),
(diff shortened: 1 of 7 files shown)


app/components/page_settings_summary_component/view.html.erb
```diff
@@ -1,73 +1,102 @@
...
-    <%= summary_list.with_row do |row|
-          row.with_key(text: t("page_settings_summary.selection.include_none_of_the_above"))
-          row.with_value(text: @draft_question.is_optional ? t("selections_settings.yes") : t("selections_settings.no"))
-          row.with_action(text: t("page_settings_summary.change"),
+                    visually_hidden_text: t("page_settings_summary.selection.options")) %>
+    <% end %>
+    <% if @long_lists_enabled %>
+      <%= summary_list.with_row do |row| %>
+        <%= row.with_key(text: t("page_settings_summary.selection.how_many_selections")) %>
+        <%= row.with_value(text: answer_settings[:only_one_option] == "true" ? t("helpers.label.pages_long_lists_selection_type_input.only_one_option_options.true") : t("helpers.label.pages_long_lists_selection_type_input.only_one_option_options.false")) %>
```
DavidBiddle, line 32: Up to you whether you think this is better but I think this is equivalent:
```suggestion
        <%= row.with_value(text: t("helpers.label.pages_long_lists_selection_type_input.only_one_option_options.#{answer_settings[:only_one_option]}")) %>
```

DavidBiddle: Ah, I fogot about that!

spec/components/page_settings_summary_component/view_spec.rb
```diff
@@ -95,6 +99,59 @@
...
+
+          it "says people can select one or more options in the summary table" do
+            rows = page.find_all(".govuk-summary-list__row")
+            expect(rows[2].find(".govuk-summary-list__value")).to have_text I18n.t("helpers.label.pages_long_lists_selection_type_input.only_one_option_options.true")
+          end
+        end
+
+        # This ensures there is backwards compatibility for existing questions as we previously set "only_one_option" to
+        # "0" rather than "false"
+        context "when only_one_option has value '0'" do
```
stephencdaly, line 138: New test to ensure backwards compatibility with existing questions

app/components/page_settings_summary_component/view.html.erb
stephencdaly: Added the test

DavidBiddle (approved):

PR #1580, opened 2024-11-07
Route to long_lists_selection_type_edit_path when changing answer type

DavidBiddle:
When updating the flow for long list questions, we missed this reference to selections_settings_edit_path. This introduced a bug/inconsistency in groups with long lists enabled:
- if a user creates a new question with the selection answer type, they're asked for their question title and then sent to long_lists_selection_type_edit_path.
- if the user changes an existing question into a selection, they skip the question type question (correctly, since the question already has question text), but they're taken to selections_settings_edit_path rather than long_lists_selection_type_edit_path.

This commit adds logic to direct users to
long_lists_selection_type_edit_path if long_lists_enabled is set on the group.

Trello card: 


stephencdaly (approved):
Good spot! Tested locally and it works as expected.

PR #1582, opened 2024-11-08
Change text for conditions in pagelist component

thomasiles:
# Change the condition text used in the "Your questions" list

Trello card: https://trello.com/c/ifg2UBf7/1964-change-the-content-describing-routes-in-the-question-list-pag

Conditions are displayed in the page list component with a text description.

This commit changes the way conditions are described.

It:
- condenses the text into a single line paragraph
- shortens the content
- includes page numbers for the question a condition goes to

To support these changes, I've changed the tests. The tests now explicitly check the result of getting a description against a string.

I chose this over referencing the translations because it's so much easier to read and understand.

I chose to test the condition_description method over the testing the sub-methods to try and test the behaviour rather than the implementation.

I considered checking the rendered component for the text. This makes working out why a test failed difficult. Instead I've chosen to check the result of the method and then check the text is included in the view.

image


config/locales/en.yml
```diff
@@ -971,10 +971,11 @@ en:
     check_your_answers: Check your answers before submitting
     condition_answer_value_text: is answered as “%{answer_value}”
     condition_answer_value_text_with_errors: is answered as [Answer not selected]
-    condition_check_page_text: If the question “%{check_page_text}”
+    condition_check_page_text: If “%{check_page_text}”
     condition_compact_html: 'If the answer is “%{answer_value}”<br> then skip to %{goto_page_number}: “%{goto_page_text}”'
     condition_compact_html_end_of_form: If the answer is “%{answer_value}”<br> then skip to “Check your answers before submitting”
-    condition_goto_page_text: take the person to “%{goto_page_text}”
+    condition_goto_page_check_your_answers: take the person to “Check your answers before submitting”
```
lfdebrux, line 977:
```suggestion
    condition_goto_page_check_your_answers: take the person to “Check your answers before submitting”.
```

lfdebrux (commented):
Looks great! Thank you especially for the detailed commit message 🙏 

I tested this locally, it matches the designs in Mural:

Screenshot of question route description for route skipping to another question

Screenshot of question route description for route skipping to the check your answers page

(except for one missing full stop when skipping to the check your answers page, which I've highlighted)

So I think this is done, but might be worth checking with @hannahkc first.

config/locales/en.yml
thomasiles: well spotted! 👍

hannahkc:
Looks good to me! Thanks both! :)

hannahkc (approved):

lfdebrux (approved):

PR #1583, opened 2024-11-08
Handle switching selection type

stephencdaly:

Trello card: https://trello.com/c/vdNsBNFy

#### Warning on the "How many options should people be able to select?" page

Display a warning on the "How many options should people be able to select?" page if there are more than 30 existing selection options.

There is a limit of 30 options when the user selects "People can select one or more options", so if the user is changing their answer to this question we want to warn them that they will need to reduce the number of options.

We already display a warning if the question has routes, as changing to "People can select one or more options" will cause the routes to be deleted. Reword this error to make sense with the new way we are asking this question.

If we would display both warnings, insead display a single notification banner with content about both these things.

#### Error message on the "Edit your question" page

Add validation on the "Edit your question" page to ensure that the form creator hasn't managed to switch the selection type to "People can select one or more options" when there are more than 30 options. This is possible if the form creator changes this setting and then clicks the back button after they are taken to the page to edit their selection options.

Display the error as an error in the summary list table, similar to how we display routing errors on the list questions page.

Warning when there are more than 30 options:

Screenshot 2024-11-12 at 10 17 30

Warning when there is a route from the question:

Screenshot 2024-11-12 at 10 16 42

Warning when there is both more than 30 options and a route from the question:

Screenshot 2024-11-12 at 10 17 18

Error on the edit your question page, if the user clicks back before reducing the number of options to 30 or less.

Screenshot 2024-11-12 at 14 29 54

### DIFF
--- a/app/components/page_settings_summary_component/view.rb
+++ b/app/components/page_settings_summary_component/view.rb
@@ -5,10 +5,11 @@ module PageSettingsSummaryComponent
     include Rails.application.routes.url_helpers
     include PagesHelper
 
-    def initialize(draft_question:, long_lists_enabled: false)
+    def initialize(draft_question:, errors: nil, long_lists_enabled: false)
       super
       @draft_question = draft_question
       @long_lists_enabled = long_lists_enabled
+      @errors = errors
     end
 
     def before_render
--- a/app/input_objects/pages/question_input.rb
+++ b/app/input_objects/pages/question_input.rb
@@ -6,10 +6,13 @@ class Pages::QuestionInput < BaseInput
   # TODO: We could lose these attributes once we have an Check your answers page
   attr_accessor :answer_settings, :page_heading, :guidance_markdown
 
+  attr_reader :selection_options # only used for displaying error
+
   validates :draft_question, presence: true
   validates :hint_text, length: { maximum: 500 }
   validates :is_optional, inclusion: { in: %w[false true] }
   validates :is_repeatable, inclusion: { in: %w[false true] }, if: -> { Settings.features.repeatable_page_enabled }
+  validate :validate_number_of_selection_options
 
   def submit
     return false if invalid?
@@ -30,5 +33,14 @@ class Pages::QuestionInput < BaseInput
 
   def repeatable_options
     [OpenStruct.new(id: "true"), OpenStruct.new(id: "false")]
+  end
+
+  def validate_number_of_selection_options
+    return if draft_question.nil?
+    return unless @draft_question.answer_type == "selection"
+    return unless @draft_question.answer_settings[:only_one_option] == "false" &&
+      @draft_question.answer_settings[:selection_options].length > 30
+
+    errors.add(:selection_options, :too_many_selection_options)
   end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -104,6 +104,8 @@ en:
             question_text:
               blank: Enter a question
               too_long: Question text must be %{count} characters or less
+            selection_options:
+              too_many_selection_options: You cannot have more than 30 options in a list when people can select more than one option
   activerecord:
     errors:
       models:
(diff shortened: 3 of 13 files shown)


app/input_objects/pages/question_input.rb
```diff
@@ -6,10 +6,14 @@ class Pages::QuestionInput < BaseInput
   # TODO: We could lose these attributes once we have an Check your answers page
   attr_accessor :answer_settings, :page_heading, :guidance_markdown
 
+  # Only used for error summary
+  attr_accessor :selection_options
```
DavidBiddle, line 10: I think if we don't need to be able to assign this attribute we can use `attr_reader` to stop ruby generating a method to assign it:
```suggestion
  attr_reader :selection_options
```

DavidBiddle: (I don't think we've been hugely consisten about this elsewhere though)

app/components/page_settings_summary_component/view.rb
```diff
@@ -5,10 +5,11 @@ class View < ViewComponent::Base
     include Rails.application.routes.url_helpers
     include PagesHelper
 
-    def initialize(draft_question:, long_lists_enabled: false)
+    def initialize(draft_question:, question_input:, long_lists_enabled: false)
```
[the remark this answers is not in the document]
stephencdaly: Nice idea, yeh that probably makes more sense

config/locales/en.yml
```diff
@@ -1276,6 +1279,14 @@ en:
     remove_html: Remove <span class="govuk-visually-hidden">option %{option_number}</span>
     select_more_than_one_option: People will be able to select more than one option from your list.
     select_one_option: People will be able to select one option from your list.
+  selection_type:
+    reduce_your_options_warning:
+      body: You can only have up to 30 options in a list where people can select one or more options. If you make this change, you'll be able to edit your list on the next page.
```
hannahkc, line 1284:
```suggestion
      body: You can only have up to 30 options in a list where people can select one or more options. If you make this change, you’ll be able to edit your list on the next page.
```

config/locales/en.yml
```diff
@@ -1276,6 +1279,14 @@ en:
...
+      heading: If you change this to ‘people can select one or more options’, you'll need to edit your list
```
hannahkc, line 1285:
```suggestion
      heading: If you change this to ‘one or more options’, you’ll need to edit your list
```

config/locales/en.yml
```diff
@@ -1276,6 +1279,14 @@ en:
...
+    routing_and_reduce_your_options_combined_warning:
+      body: You can only have up to 30 options in a list where people can select one or more options. If you make this change, you'll be able to edit your list on the next page.
```
hannahkc, line 1287:
```suggestion
      body: You can only have up to 30 options in a list where people can select one or more options. If you make this change, you’ll be able to edit your list on the next page.
```

config/locales/en.yml
```diff
@@ -1276,6 +1279,14 @@ en:
...
+      heading: If you change this to ‘people can select one or more options’, the route from this question will be deleted and you'll need to edit the list
```
hannahkc, line 1288:
```suggestion
      heading: If you change this to ‘one or more options’, the route from this question will be deleted and you’ll need to edit the list
```

config/locales/en.yml
```diff
@@ -1276,6 +1279,14 @@ en:
...
+    routing_warning: If you change this to ‘people can select one or more options’, the route from this question will be deleted
```
hannahkc, line 1289:
```suggestion
    routing_warning: If you change this to ‘one or more options’, the route from this question will be deleted
```

hannahkc (commented):
We're shortening the wording of the options on this page so have updated that in these warnings too. And corrected a couple of straight quotes.

config/locales/en.yml
stephencdaly: Thanks, have committed these changes.

DavidBiddle (approved):
Code looks good and when testing locally I can see the warning message on the type page and the error on the edit question page :tada: styling looks good too!

PR #1587, opened 2024-11-11
Add a PageRepository and reroute Page model usage through it

SamJamCul:
The PageRepository being introduced is used as an interface between the application and any Page method calls that result in an API call through ActiveResource. The intention is to create a layer of separation, so that database requests can be rerouted more easily as part of the Admin/API responsibilities shifting.

Trello card: https://trello.com/c/QuEeu8uI/1939-add-repository-service-to-forms-admin-for-form-page-model

This PR adds the PageRepository, which is intended to act as a buffer between the app and any ActiveResource API usage when using the Page model. I've tried to go through and find all the spots where we use Page model requests, it's a little fiddly! 

As well as CRUD ops, it also implements the `move_page`, as this also hooks into the API.

If you know of anywhere else that we use the API's Page model, please leave a comment and I'll reroute it through the PageRepository.

### DIFF
--- a/app/controllers/forms/delete_confirmation_controller.rb
+++ b/app/controllers/forms/delete_confirmation_controller.rb
@@ -53,11 +53,11 @@ module Forms
       @delete_confirmation_input = DeleteConfirmationInput.new
 
       if params[:page_id].present?
-        @page = Page.find(params[:page_id], params: { form_id: current_form.id })
-        @url = destroy_page_path(current_form, @page)
+        @page = PageRepository.find(page_id: params[:page_id], form_id: current_form.id)
+        @url = destroy_page_path(current_form, @page.id)
         @confirm_deletion_legend = t("forms_delete_confirmation_input.confirm_deletion_page")
         @item_name = @page.question_text
-        @back_url = edit_question_path(current_form, @page)
+        @back_url = edit_question_path(current_form, @page.id)
       else
         @url = destroy_form_path(current_form)
         @confirm_deletion_legend = t("forms_delete_confirmation_input.confirm_deletion")
@@ -80,7 +80,7 @@ module Forms
     def delete_page(form, page)
       success_url = form_pages_path(form)
 
-      if page.destroy
+      if PageRepository.destroy(page)
         redirect_to success_url, status: :see_other, success: "Successfully deleted ‘#{page.question_text}’"
       else
         raise StandardError, "Deletion unsuccessful"
--- a/app/controllers/pages/questions_controller.rb
+++ b/app/controllers/pages/questions_controller.rb
@@ -19,9 +19,8 @@ class Pages::QuestionsController < PagesController
   def create
     @question_input = Pages::QuestionInput.new(page_params_for_form_object)
 
-    # TODO: Move Page creation to be part of the form submit method
-    if @question_input.submit
-      @page = PageRepository.create!(page_params_for_forms_api)
+    if @question_input.valid?
+      @page = @question_input.submit
       clear_draft_questions_data
       redirect_to edit_question_path(current_form, @page.id), success: "Your changes have been saved"
     else
@@ -44,8 +43,7 @@ class Pages::QuestionsController < PagesController
 
     @question_input = Pages::QuestionInput.new(page_params_for_form_object)
 
-    # TODO: Move Page creation to be part of the form submit method
-    if @question_input.submit && (@page = PageRepository.save!(@page))
+    if @question_input.update_page(@page)
       clear_draft_questions_data
       redirect_to edit_question_path(current_form, @page.id), success: "Your changes have been saved"
     else
@@ -61,6 +59,7 @@ private
 
   def page_params_for_form_object
     page_params.merge(draft_question:,
+                      form_id: current_form.id,
                       answer_type: draft_question.answer_type,
                       answer_settings: draft_question.answer_settings,
                       page_heading: draft_question.page_heading,
--- a/app/controllers/pages/questions_controller.rb
+++ b/app/controllers/pages/questions_controller.rb
@@ -19,12 +19,11 @@ class Pages::QuestionsController < PagesController
   def create
     @question_input = Pages::QuestionInput.new(page_params_for_form_object)
 
-    @page = Page.new(page_params_for_forms_api)
-
     # TODO: Move Page creation to be part of the form submit method
-    if @question_input.submit && @page.save
+    if @question_input.submit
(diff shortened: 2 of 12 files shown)


app/models/page.rb
```diff
@@ -46,10 +46,6 @@ def show_selection_options
     answer_settings.selection_options.map { |option| option.attributes[:name] }.join(", ")
   end
 
-  def submit
```
SamJamCul, line 49: This is no longer used (and possibly never was) but it seemed like a good time to junk it.

app/services/page_repository.rb
```diff
@@ -0,0 +1,33 @@
+class PageRepository
+  class << self
+    def find(page_id:, form_id:)
+      Page.find(page_id, params: { form_id: })
+    end
+
+    def create!(record)
+      page = Page.new(record.attributes, false)
+      page.prefix_options = record.prefix_options
+      page.save # rubocop:disable Rails/SaveBang
```
lfdebrux, line 10: I think we can change this to `save!`

```suggestion
      page.save!
```

app/services/page_repository.rb
```diff
@@ -0,0 +1,33 @@
...
+      page
+    end
+
+    def save!(record)
+      page = Page.new(record.attributes, true)
+      page.prefix_options = record.prefix_options
+      page.save # rubocop:disable Rails/SaveBang
```
lfdebrux, line 17:
```suggestion
      page.save!
```

spec/requests/forms/delete_confirmation_controller_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe Forms::DeleteConfirmationController, type: :request do
   let(:form) { build(:form, :with_active_resource, id: 2) }
+  let(:page) { build(:page, form_id: form.id) }
```
lfdebrux, line 5: Thanks for adding tests to the delete confirmation controller! Would you mind putting them in a separate commit?

spec/requests/pages/questions_controller_spec.rb
```diff
@@ -350,63 +310,6 @@
...
-        end
-
-        it "does not set is_repeatable" do
-          matched_request = ActiveResource::HttpMock.requests.find do |request|
-            request.method == :put && request.path == "/api/v1/forms/2/pages/1"
-          end
-
-          expect(JSON.parse(matched_request.body)).to include("is_repeatable" => true)
-        end
-      end
```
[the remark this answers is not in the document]
lfdebrux: I think these tests are from a time when the `is_repeatable` attribute wasn't always being exposed to users, so we wanted to make sure that if a controller updated a page without giving a value for that attribute its value wouldn't be changed.

That probably isn't an issue any more.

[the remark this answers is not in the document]
lfdebrux: So I think these changes are good. Would you mind putting them into a different commit?

app/controllers/forms/delete_confirmation_controller.rb
```diff
@@ -53,11 +53,11 @@ def load_page_variables
...
 
       if params[:page_id].present?
-        @page = Page.find(params[:page_id], params: { form_id: current_form.id })
-        @url = destroy_page_path(current_form, @page)
+        @page = PageRepository.find(page_id: params[:page_id], form_id: current_form.id)
+        @url = destroy_page_path(current_form, @page.id)
         @confirm_deletion_legend = t("forms_delete_confirmation_input.confirm_deletion_page")
         @item_name = @page.question_text
-        @back_url = edit_question_path(current_form, @page)
+        @back_url = edit_question_path(current_form, @page.id)
```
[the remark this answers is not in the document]
SamJamCul: The url builder complains about being unable to find the `page_id` for the page object that's constructed using the repository. It was unable to find it, and so failed to build the link.

app/controllers/pages/questions_controller.rb
```diff
@@ -22,9 +22,9 @@ def create
     @page = Page.new(page_params_for_forms_api)
 
     # TODO: Move Page creation to be part of the form submit method
-    if @question_input.submit && @page.save
+    if @question_input.submit && (@page = PageRepository.create!(@page))
```
lfdebrux, line 25: I wonder if we can refactor this whole bit to put the `create!` inside the if statement. Then we can have the signature of `PageRepository#create!` be more similar to the signature of `ConditionRepository#create!`.

app/services/page_repository.rb
SamJamCul: Sorted

app/services/page_repository.rb
SamJamCul: 👍🏻

spec/requests/forms/delete_confirmation_controller_spec.rb
SamJamCul: Sorted!

spec/requests/pages/questions_controller_spec.rb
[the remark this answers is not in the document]
SamJamCul: Sorted

app/controllers/pages/questions_controller.rb
SamJamCul: I'm not sure I follow - do you think we should 
 - move the `.create!` call within the condition
 - refactor the repo's `create!` so that we can pass in the new page params
 as opposed to what we're doing now, constructing the new Page on l22 (which we sort of throw away pretty quickly)?

SamJamCul: Sorted, it now takes in the params that are set in the `page_params_for_forms_api` and hands them to `Page.create` in the page repository.

app/controllers/pages/questions_controller.rb
```diff
@@ -19,9 +19,7 @@ def new
   def create
     @question_input = Pages::QuestionInput.new(page_params_for_form_object)
 
-    # TODO: Move Page creation to be part of the form submit method
-    if @question_input.submit
-      @page = PageRepository.create!(page_params_for_forms_api)
+    if @question_input.valid? && (@page = @question_input.submit)
```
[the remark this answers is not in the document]
lfdebrux: If `@question_input.submit` will always return a page object if the question input is valid, then can we move the call and assignment inside the if block?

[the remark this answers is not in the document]
SamJamCul: Yes! sorting it, good thinking 🤔

lfdebrux (approved):

PR #1595, opened 2024-11-12
Remove code hiding groups controller behind feature flag

lfdebrux:

We don't ever want to switch off the groups feature flag, so we don't need this code.

Unfortunately I don't have time to remove all the related code, but we can at least chip away at it slowly.


DavidBiddle (approved):
Change looks good and no issues with the groups feature when running locally 🎉

PR #1596, opened 2024-11-12
Show secondary skip conditions as cards and in the "your questions" list

thomasiles:

Trello card: https://trello.com/c/jmTMiyBI/1965-show-conditions-without-answer-value-as-part-of-branch-route-in-forms-admin

Change page list for a form and route summary cards on the edit question routes page to work with the secondary skip route for a branch.

image

image

image


app/components/page_list_component/view.rb
```diff
@@ -57,5 +57,31 @@ def condition_page_position(condition)
...
     end
+
+    def conditions_for_page_with_index(page_id)
+      routing_conditions_with_index.fetch(page_id, [])
+    end
+
+    def routing_conditions_with_index
+      @routing_conditions_with_index ||= process_routing_conditions
+    end
+
```
thomasiles, line 68: process_routing_conditions is a fun method!

We are labelling the conditions with order of their check_page_id but displaying them in order of routing_page_id. We use routing_conditions_with_index as an index to conditions, memoising it so that we only do it once for the page list.

It's a hash. The keys are page_id's, the values are arrays of [condition, index].

When we display conditions and we need to create names like "Route n of question m's route" the index is the n.

I'm very open to suggestions for better ways to do this.

config/locales/en.yml
```diff
@@ -969,15 +969,16 @@ en:
...
+    condition_answer_value_text: "“%{answer_value}”"
+    condition_answer_value_text_with_errors: "[Answer not selected]"
+    condition_check_page_text: "“%{check_page_text}”"
     condition_compact_html: 'If the answer is “%{answer_value}”<br> then skip to %{goto_page_number}: “%{goto_page_text}”'
     condition_compact_html_end_of_form: If the answer is “%{answer_value}”<br> then skip to “Check your answers before submitting”
-    condition_goto_page_check_your_answers: take the person to “Check your answers before submitting”.
-    condition_goto_page_text: take the person to %{goto_page_position}, “%{goto_page_text}”
-    condition_goto_page_text_with_errors: take the person to [Question not selected]
-    condition_name: Question %{page_index}’s route
+    condition_description: If %{check_page_text} is answered as %{answer_value} take the person to %{goto_page_text}
```
lfdebrux, line 977: In the [latest designs](https://app.mural.co/t/gaap0347/m/gaap0347/1728478914347/6a7e0b709f50f00f81cab37414fdf44ec8601a70?wid=221-1730300100613) we've changed "take the person to" to "go to"

```suggestion
    condition_description: If %{check_page_text} is answered as %{answer_value} go to %{goto_page_text}
```

config/locales/en.yml
```diff
@@ -969,15 +969,16 @@ en:
...
+    condition_goto_page_check_your_answers: "“Check your answers before submitting”."
+    condition_goto_page_text: "%{goto_page_position}, “%{goto_page_text}”"
+    condition_goto_page_text_with_errors: "[Question not selected]"
+    condition_name: Route %{route_index} of question %{page_index}’s route
```
lfdebrux, line 981: Questions always have more than one route, so this should be plural:

```suggestion
    condition_name: Route %{route_index} of question %{page_index}’s routes
```

config/locales/en.yml
```diff
@@ -991,6 +992,7 @@ en:
         goto_page_doesnt_exist: The question you’re taking the person to no longer exists - edit question %{page_index}’s route
     none_of_the_above: None of the above
     route: Route
+    secondary_skip_description: After %{check_page_text} go to %{goto_page_text}
```
lfdebrux, line 995: I think we want the question number of the check page in here as well, see https://app.mural.co/t/gaap0347/m/gaap0347/1728478914347/6a7e0b709f50f00f81cab37414fdf44ec8601a70?wid=0-1730911072926

lfdebrux (commented):
Some small design tweaks

app/presenters/route_summary_card_data_presenter.rb
```diff
@@ -25,7 +25,7 @@ def summary_card_data
 
   def all_routes
     all_form_routing_conditions = pages.flat_map(&:routing_conditions).compact_blank
-    all_form_routing_conditions.reject { |rc| rc.routing_page_id != page.id || rc.check_page_id != page.id }
+    all_form_routing_conditions.select { |rc| (rc.routing_page_id == page.id) || (rc.check_page_id == page.id) }
```
lfdebrux: Interesting that this needed changing! Is it possible to add a test for it? 🤔

spec/presenters/route_summary_card_data_presenter_spec.rb
```diff
@@ -58,8 +60,24 @@
       end
     end
 
+    context "with a route with no answer_value" do
+      let(:routing_conditions) do
+        [
+          build(:condition, id: 1, routing_page_id: 1, check_page_id: 1, answer_value: "Yes", goto_page_id: 2, skip_to_end: false),
+          build(:condition, id: 2, routing_page_id: 2, check_page_id: 1, answer_value: nil, goto_page_id: nil, skip_to_end: true),
+        ]
+      end
```
lfdebrux, line 69: This test doesn't seem quite right; with the two conditions built here, they wouldn't both end up in page 1's `routing_conditions` attribute, would they?

app/components/page_list_component/view.rb
```diff
@@ -17,7 +17,11 @@ def show_down_button(index)
     end
 
     def condition_description(condition)
-      I18n.t("page_conditions.condition_description", check_page_text: condition_check_page_text(condition), goto_page_text: goto_page_text_for_condition(condition), answer_value: answer_value_text_for_condition(condition))
+      if condition.secondary_skip?
+        I18n.t("page_conditions.secondary_skip_description", check_page_text: condition_check_page_text(condition), goto_page_text: goto_page_text_for_condition(condition))
+      else
+        I18n.t("page_conditions.condition_description", check_page_text: condition_check_page_text(condition), goto_page_text: goto_page_text_for_condition(condition), answer_value: answer_value_text_for_condition(condition))
+      end
```
lfdebrux: If we wanted to reduce lines I think we could make the logic change just the translation key?

app/components/page_list_component/view.rb
lfdebrux: Thanks for the detailed explanation of what this is doing and the challenges.

I think this is a good approach in general. If we wanted to be able to test it more it might be worth splitting it out into its own class somewhere, but if we're not expecting to reuse this anywhere that might be overkill.

spec/presenters/route_summary_card_data_presenter_spec.rb
thomasiles: Thats a good point!

app/components/page_list_component/view.rb
thomasiles: Yes, I think this code might end up in it's own service so that the numbering is in one place. I think that passing in the conditions for the page might simplify this code.

I just want to check that this is used elsewhere, probably in making new routes.

lfdebrux (approved):

PR #1603, opened 2024-11-14
Use details to show selection options

DavidBiddle:

Trello card: https://trello.com/c/djqTBCoO/1983-add-show-x-options-details-component-for-lists-on-edit-question-page?filter=label:Feature%20team%20one,label:Tech%20Support,label:X%20feature%20teams

Renders the list of selection options in a details component if there are 10 of more entries in the list.

#### Closed (the default):
A GOV.UK summary list item with key 'Options', and a change link. In the value area it has a closed GOV.UK details component with summary value '10 options'.

#### Open:
The same summary list item shown in the image above, but the details component is now open, revealing 10 bullet points named 'Option 1', 'Option 2', Option 3' and so on until 'Option 10'.


stephencdaly (approved):
Looks good to me! Have checked locally and it works as expected.

PR #1608, opened 2024-11-18
Add username and password auth for user research environment

lfdebrux:

Trello card: https://trello.com/c/mwfuU1o0/1989-fix-the-ur-environment-so-that-participants-can-log-in 

We want to stop using HTTP Basic Auth in the user research environment, because it seems to work less and less with the browsers of our research participants. We still don't want to be using the full-fat Auth0 authentication though, as we want to be able to set up test user accounts in advance and stop the general public from using this environment.

This commit adds a HTML form for signing in, with the same level of security as the HTTP Basic Auth we were previously using, but more likely to be compatible with modern browsers.

We use OmniAuth to create the form and compare the credentials, which is a bit of an awkward fit, but is simple to start with. If the credentials don't match, we just bounce the user out to our even less pretty auth failure page, with a link to retry. None of this is using the GOV.UK Design System styles and patterns like the rest of the site, but again I think this is fine for now.

If you want to test this locally, you'll need to set the forms_env to `user-research`, as well as setting the `auth_provider` and a username and password:

```
SETTINGS__FORMS_ENV=user-research SETTINGS__AUTH_PROVIDER=user-research SETTINGS__USER_RESEARCH__AUTH__USERNAME=user1 SETTINGS__USER_RESEARCH__AUTH__PASSWORD=test bin/rails s
```

I've tested these changes in the user research environment.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,4 +19,7 @@ inherit_mode:
 # See https://github.com/alphagov/rubocop-govuk/blob/main/CONTRIBUTING.md
 # **************************************************************
 
+RSpec/SpecFilePathFormat:
+  CustomTransform:
+    OmniAuth: omniauth
 
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -72,6 +72,8 @@ private
     root_path
   end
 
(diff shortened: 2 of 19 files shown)


alice-carr (approved):
This all makes sense, and I can see this working at https://admin.research.forms.service.gov.uk/

PR #1612, opened 2024-11-19
Add a form for setting the "secondary skip" route of a question

thomasiles:
### Add a form for setting the "secondary skip" route of a question

Trello card: https://trello.com/c/Ubj0HU4s/1966-allow-form-creators-to-add-a-secondary-route-to-a-question-route-for-up-to-2-branches

image

This PR is adds the input model, view, controller and route for creating new "secondary_skip" conditions.

This feature is behind the flag "branch_routing".

The following features haven't been done yet:
- redirecting the user if the user is trying to create a skip for question which doesn't have a route
- preventing the user adding more than one skip using /new(!)

These will be added soon.

Also not included from the ticket is JS code to manipulate the select options of the "take them to" field based on the answer to the first select. That will be done separately too. 

There are more tests to be added too for example request tests for the edit/create routes.

I think a few other things can also be cleaned up and refactored but I think it's safe to do those later.

### DIFF
--- a/app/controllers/pages/secondary_skip_controller.rb
+++ b/app/controllers/pages/secondary_skip_controller.rb
@@ -22,6 +22,28 @@ class Pages::SecondarySkipController < PagesController
     end
   end
 
+  def edit
+    secondary_skip_input = Pages::SecondarySkipInput.new(form: current_form, page:, record: secondary_skip_condition).assign_values
+
+    render template: "pages/secondary_skip/edit", locals: {
+      secondary_skip_input:,
+      back_link_url: show_routes_path(form_id: current_form.id, page_id: page.id),
+    }
+  end
+
+  def update
+    secondary_skip_input = Pages::SecondarySkipInput.new(secondary_skip_input_params.merge(record: secondary_skip_condition)).assign_values
+
+    if secondary_skip_input.submit
+      redirect_to show_routes_path(form_id: current_form.id, page_id: page.id)
+    else
+      render template: "pages/secondary_skip/edit", locals: {
+        secondary_skip_input:,
+        back_link_url: show_routes_path(form_id: current_form.id, page_id: page.id),
+      }, status: :unprocessable_entity
+    end
+  end
+
 private
 
   def secondary_skip_input_params
@@ -30,5 +52,9 @@ private
 
   def ensure_branch_routing_feature_enabled
     raise ActionController::RoutingError, "branch_routing feature not enabled" unless Settings.features.branch_routing
+  end
+
+  def secondary_skip_condition
+    @secondary_skip_condition ||= current_form.pages.flat_map(&:conditions).compact_blank.find(&:secondary_skip?)
   end
 end
--- a/app/controllers/pages/secondary_skip_controller.rb
+++ b/app/controllers/pages/secondary_skip_controller.rb
@@ -8,7 +8,25 @@ class Pages::SecondarySkipController < PagesController
       back_link_url: show_routes_path(form_id: current_form.id, page_id: page.id),
     }
   end
+
+  def create
+    secondary_skip_input = Pages::SecondarySkipInput.new(secondary_skip_input_params)
+
+    if secondary_skip_input.submit
+      redirect_to show_routes_path(form_id: current_form.id, page_id: page.id)
+    else
+      render template: "pages/secondary_skip/new", locals: {
+        secondary_skip_input:,
+        back_link_url: show_routes_path(form_id: current_form.id, page_id: page.id),
+      }, status: :unprocessable_entity
+    end
+  end
+
 private
+
+  def secondary_skip_input_params
+    params.require(:pages_secondary_skip_input).permit(:routing_page_id, :goto_page_id).merge(form: current_form, page:)
+  end
 
   def ensure_branch_routing_feature_enabled
     raise ActionController::RoutingError, "branch_routing feature not enabled" unless Settings.features.branch_routing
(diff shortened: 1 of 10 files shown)


app/controllers/pages/secondary_skip_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def create
+    secondary_skip_input = Pages::SecondarySkipInput.new(secondary_skip_input_params)
+
+    if secondary_skip_input.submit
+      redirect_to show_routes_path(form_id: current_form.id, page_id: page.id)
+    else
+      render template: "pages/secondary_skip/new", locals: {
+        secondary_skip_input:,
+        back_link_url: show_routes_path(form_id: current_form.id, page_id: page.id),
```
thomasiles, line 20: I think back links can now be directly in the views and we don't need to pass locals at all. Secondary_skip_input can be an instance variable instead.

app/views/pages/secondary_skip/edit.html.erb
```diff
@@ -0,0 +1,11 @@
+<% set_page_title(title_with_error_prefix(t('page_titles.new_secondary_skip', route_index: 2), false)) %>
```
thomasiles, line 1: I've left the heading the same for now.

app/input_objects/pages/secondary_skip_input.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  def answer_value
+    page.conditions.find { |rc| rc.answer_value.present? }.answer_value
+  end
+
+  def continue_to
+    page.has_next_page? ? page_name(page.next_page) : end_page_name
+  end
+
```
thomasiles, line 78: the above methods are all for displaying the form rather than related directly to the form model.

I've put them here for now because it's direct. there maybe a better location for them

config/locales/en.yml
```diff
@@ -1308,6 +1323,17 @@ en:
         You can only add one route from each question.
       </p>
   save_and_continue: Save and continue
+  secondary_skip:
+    new:
+      back: Back to question %{page_position}'s routes
+      caption: Question %{page_position}'s routes
+      details_html: |
+        <p>People who select an answer that does not have a route assigned to it will continue to the next question.</p>
+        <p>If you need to, you can make these people skip one or more questions later in the form. For example, you might want them to skip route 1's questions. This allows you to create 2 different ‘branches’ of questions.</p>
```
lfdebrux, line 1332:
```suggestion
        <p>If you need to, you can make these people skip one or more questions later in the form. For example, you might want them to skip route 1’s questions. This allows you to create 2 different ‘branches’ of questions.</p>
```

config/locales/en.yml
```diff
@@ -1308,6 +1323,17 @@ en:
         You can only add one route from each question.
       </p>
   save_and_continue: Save and continue
+  secondary_skip:
+    new:
+      back: Back to question %{page_position}'s routes
+      caption: Question %{page_position}'s routes
```
lfdebrux, line 1329:
```suggestion
      back: Back to question %{page_position}’s routes
      caption: Question %{page_position}’s routes
```

lfdebrux (commented):
Spotted some quotes that could be curly

lfdebrux (commented):
This works well on my machine, nice one getting this done!

I'm still going through the code, but so far what I'm seeing looks great.

app/views/pages/secondary_skip/_form.html.erb
```diff
@@ -0,0 +1,36 @@
+<%= form_with(model: secondary_skip_input, url: submit_path) do |f| %>
+  <% if  secondary_skip_input&.errors.any? %>
+    <%= f.govuk_error_summary %>
+  <% end %>
+
+  <h1 class="govuk-heading-l">
+    <span class="govuk-caption-l"><%= t("secondary_skip.new.caption", page_position: secondary_skip_input.page.position) %></span>
+    <%= t("page_titles.new_secondary_skip", route_index: 2) %>
```
lfdebrux, line 8: Let's put a TODO comment here to come back to this magic number when we start supporting multiple branches...

app/views/pages/secondary_skip/_form.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  </h1>
```
lfdebrux, line 9: It probably makes sense to have the page heading in the page template rather in this partial, so that it and the page title can be changed at the same time easily

lfdebrux (approved):
:shipit:

config/locales/en.yml
```diff
@@ -106,6 +106,15 @@ en:
               too_long: Question text must be %{count} characters or less
             selection_options:
               too_many_selection_options: You cannot have more than 30 options in a list when people can select more than one option
+        pages/secondary_skip_input:
+          attributes:
+            goto_page_id:
+              already_consecutive: The question you want to skip to is already straight after the question you want to skip from
+              blank: Select the question or page you want to take the person to
+              equal: The question you want to skip from  cannot be the same question you want to take them to
```
hannahkc, line 114:
```suggestion
              equal: The question you want to skip from cannot be the same question you want to take them to
```

thomasiles: well spotted! I've updated it

config/locales/en.yml
thomasiles: Thanks, I should really be better at checking these before asking for a review

lfdebrux (approved):

PR #1613, opened 2024-11-19
Log entry method for selection options

stephencdaly:

Trello card: https://trello.com/c/SVZMGF6m

We want to know how many people are using the text box or individual fields to enter their options when creating a "Select from a list of options" question. We want to know whether this depends on how many options they are entering.

Log when options are submitted, with:
- The input method (is_bulk_entry)
- The number of options (options_count)
- Whether people can select more than one option (only_one_option)

We will use this log line to create a report in Splunk.

Example log line:

```json
{
  "level": "INFO",
  "time": "2024-11-19T17:19:54.191+00:00",
  "message": "Submitted selection options for a selection question",
  "is_bulk_entry": true,
  "options_count": 5,
  "only_one_option": false,
  "host": "localhost",
  "request_id": "b2754d29-0ecf-41ba-b08d-38387f9a01e3",
  "session_id_hash": "1d58c47c593c22b10014f124ada9d0de46e4f1b2386c2fb5fe5ab63e6597c6d5",
  "user_id": 1,
  "user_email": "example@example.com",
  "user_organisation_slug": "government-digital-service",
  "form_id": "8",
  "page_id": "32"
}
```

### DIFF
--- a/app/helpers/logging_helper.rb
+++ b/app/helpers/logging_helper.rb
@@ -1,0 +1,9 @@
+module LoggingHelper
+  def log_selection_question_options_submitted(is_bulk_entry:, options_count:, only_one_option:)
+    Rails.logger.info("Submitted selection options for a selection question", {
+      is_bulk_entry:,
+      options_count:,
+      only_one_option:,
+    })
+  end
+end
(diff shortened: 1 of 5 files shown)


DavidBiddle (approved):
Neat solution, code looks good and works for me locally :tada:

PR #1614, opened 2024-11-20
Fix task description

lfdebrux:

I noticed a task description that looks incorrect, I've changed it to what I think it should be.


DavidBiddle (approved):
Good spot! 🎉

PR #1615, opened 2024-11-20
Update links in selection questions reports

stephencdaly:

Trello card: https://trello.com/c/SVZMGF6m/1985-measure-and-report-how-form-creators-use-long-lists-of-options

Update the links to forms in the reports that list the selection questions with each input component to link to the readonly page that lists the questions for the live form, rather than the draft questions page.


DavidBiddle (approved):
Looks good and works locally 🎉

PR #1616, opened 2024-11-21
Fix secondary_skip #update and add before_actions to prevent errors

thomasiles:
# Fix errors in editing secondary skip routes

This PR combines two things:
- fixes an issue which prevents secondary skips being updated correctly
- adds before actions to prevent secondary skips being added when they shouldn't be

They could be separate PRs but are combined because they cover very similar parts of the code and the updates to the tests reveal them.

### DIFF
--- a/app/controllers/pages/secondary_skip_controller.rb
+++ b/app/controllers/pages/secondary_skip_controller.rb
@@ -1,5 +1,5 @@
 class Pages::SecondarySkipController < PagesController
-  before_action :ensure_branch_routing_feature_enabled
+  before_action :ensure_branch_routing_feature_enabled, :ensure_page_has_skip_condition
 
   def new
     secondary_skip_input = Pages::SecondarySkipInput.new(form: current_form, page:)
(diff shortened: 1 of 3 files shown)


app/controllers/pages/secondary_skip_controller.rb
```diff
@@ -54,6 +56,20 @@ def ensure_branch_routing_feature_enabled
...
+    end
+  end
+
+  def ensure_secondary_skip_exists
+    redirect_to show_routes_path(form_id: current_form.id, page_id: page.id) if secondary_skip_condition.blank?
+  end
+
+  def ensure_secondary_skip_blank
+    redirect_to show_routes_path(form_id: current_form.id, page_id: page.id) if secondary_skip_condition.present?
+  end
```
lfdebrux, line 71: I think rather than doing a redirect here we should just return a 404, or even raise an exception; we shouldn't be directing users to these pages in these scenarios right?

lfdebrux (approved):
Really nice! I like how you've split the commits up a lot. Sorry I missed this bug during my initial review, good call adding more request specs.

I've left one comment about the before actions, but it's not a blocker.

app/controllers/pages/secondary_skip_controller.rb
thomasiles: hmm, thats a good point. I was thinking the only time this would show up would be when a user had multiple tabs open. In that case a link might still point to new and I figured redirecting them back to the routes page would make sense rather than showing an error.

That might pretty confusing in use though so perhaps showing an error is better.

I'm going to merge this for now but I'm happy to change it.

lfdebrux: Ooh, that's a good point about multiple tabs, I hadn't thought about that. Something to think more about then.

PR #1618, opened 2024-11-21
1967 add link to set questions to skip page behind feature flag in forms admin

SamJamCul:

Trello card: https://trello.com/c/pNKOqQ1j/1967-add-link-to-set-questions-to-skip-page-behind-feature-flag-in-forms-admin

Adds links to the routing summary cards for secondary routes. These show up only when the `branch_routing` flag is on. 

When creating a route, you can see a link to create a secondary skip on the secondary route card. 

image

Once a secondary route has been created, you will get an `Edit` action at the top of the summary card.

image

### DIFF
--- a/app/presenters/route_summary_card_data_presenter.rb
+++ b/app/presenters/route_summary_card_data_presenter.rb
@@ -45,7 +45,7 @@ private
         title: I18n.t("page_route_card.route_title", index:),
         classes: "app-summary-card",
         actions: [
-          govuk_link_to("Edit", edit_condition_path(form_id: form.id, page_id: page.id, condition_id: routing_condition.id)),
+          govuk_link_to(I18n.t("page_route_card.edit"), edit_condition_path(form_id: form.id, page_id: page.id, condition_id: routing_condition.id)),
         ],
       },
       rows: [
@@ -68,6 +68,9 @@ private
       card: {
         title: I18n.t("page_route_card.route_title", index:),
         classes: "app-summary-card",
+        actions: [
+          edit_secondary_skip_link,
+        ],
       },
       rows: [
         {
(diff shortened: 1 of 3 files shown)


app/presenters/route_summary_card_data_presenter.rb
```diff
@@ -79,10 +82,27 @@ def default_route_card(index)
...
+      govuk_link_to("Edit", edit_secondary_skip_path(form_id: form.id, page_id: page.id))
+    end
+  end
+
   def secondary_skip_rows
     secondary_skip = all_routes.find(&:secondary_skip?)
 
-    return [] if secondary_skip.blank?
+    if secondary_skip.blank?
+      if Settings.features.branch_routing
```
lfdebrux, line 95: I think we prefer to use the `FeatureService` to check if a feature is enabled:

```suggestion
      if FeatureService.enabled?(:branch_routing)
```

lfdebrux (approved):
Amazing stuff!

app/presenters/route_summary_card_data_presenter.rb
```diff
@@ -79,10 +82,27 @@ def default_route_card(index)
     }
   end
 
+  def edit_secondary_skip_link
+    if Settings.features.branch_routing && all_routes.find(&:secondary_skip?).present?
+      govuk_link_to("Edit", edit_secondary_skip_path(form_id: form.id, page_id: page.id))
```
lfdebrux, line 87: This should use a translation string, it would be nice if we had a generic "Edit" key 🤔 

```suggestion
      govuk_link_to(t(".edit"), edit_secondary_skip_path(form_id: form.id, page_id: page.id))
```

app/presenters/route_summary_card_data_presenter.rb
SamJamCul: Good point, I think there's another spot I can refactor that as well

lfdebrux (approved):

PR #1619, opened 2024-11-21
Fix secondary skip display pagelist

thomasiles:
### Fix the text for secondary skip's in the pagelist component

Trello card: https://trello.com/c/faPNxjxY/2517-update-routing-pages-to-match-designs-snags-ticket

The text used when showing secondary skips in the pagelist is referencing the wrong question.

This commit fixes it.

## Wrong, before this commit
image

## Right, after this commit
image

See the designs for more info.

### DIFF
--- a/app/components/page_list_component/view.rb
+++ b/app/components/page_list_component/view.rb
@@ -18,7 +18,7 @@ module PageListComponent
 
     def condition_description(condition)
       if condition.secondary_skip?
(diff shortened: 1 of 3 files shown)


lfdebrux (approved):

lfdebrux:
Oh good spot! It's hard keeping all this stuff together in one's head 😅

PR #1623, opened 2024-11-22
Turn on long lists feature for all users

DavidBiddle:

Trello card: https://trello.com/c/v9BZe6or/1998-release-long-lists-with-autocomplete-for-all-departments?filter=label:Feature%20team%20one,label:Tech%20Support,label:X%20feature%20teams

Includes:
- removing any use of the long_lists_enabled flag on groups
- updating and cleaning up the tests
- A tiny content fix for some visually hidden text that had the wrong translation in it

There are some cleanup tasks that this PR does not include:
- removing the `long_lists_enabled` database column and the task for setting it. That will be done separately, once we're happy that this PR hasn't broken anything and won't need to be reverted.
- removing the old selection type files and translations, since we don't need them any more
- renaming references to 'long list selection' (since we now only have one type of selection)

These will need to be done in one or more subsequent PRs once we're confident in the changes.

### DIFF
--- a/app/components/page_settings_summary_component/view.html.erb
+++ b/app/components/page_settings_summary_component/view.html.erb
@@ -39,22 +39,12 @@
                     href: @change_selections_options_path,
                     visually_hidden_text: t("page_settings_summary.selection.options")) %>
     <% end %>
-    <% if @long_lists_enabled %>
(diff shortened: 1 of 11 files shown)


stephencdaly (approved):
Changes all look good for me and it works as expected when running locally.

PR #1627, opened 2024-11-25
SPIKE: Delete users with no name/organisation

stephencdaly:

Trello card: https://trello.com/c/B7dPRWtc/

This tests deleting users that have not set their organisation or name along with deleting their forms.

Users are skipped if they are a member of a group - as we do not expect any users who have not set their organisation or name to be a member of a group, as groups were introduced after we started asking for the organisation and name, and we only put people's forms in a default group once they have entered their organisation.

If users do have a group, it probably makes sense to double check that they are the only member of the group and run the dedicated task to remove that group first.

We delete the user's account in auth0 - which requires the "Auth0 Management API" to be enabled for the application in the Auth0 tenant with the `delete:users` permission enabled.

### Testing

You can test this locally by connecting to the dev tenant in Auth0 using the credentials for the `forms-admin-local-dev` auth0 application. Start forms-admin using 
```
SETTINGS__AUTH_PROVIDER=auth0 SETTINGS__AUTH0__CLIENT_ID=<app-client-id> SETTINGS__AUTH0__CLIENT_SECRET=<client-secret> SETTINGS__AUTH0__DOMAIN=<auth-domain> bundle exec rails s
```

You can then use plus addressing to create additional users with your email as this application doesn't have Google Workspaces enabled. You can't do this on a remote environment because our email addresses will always use Google Workplaces to log in.

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -1,4 +1,6 @@
 namespace :users do
+  require "auth0"
+
   desc "Update all trial and editor users to be standard users - dry run"
   task update_user_roles_to_standard_dry_run: :environment do
     ActiveRecord::Base.transaction do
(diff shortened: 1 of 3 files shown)


lib/tasks/users.rake
```diff
@@ -12,6 +14,49 @@ namespace :users do
...
+
+      if user.memberships.present?
+        group_ids = user.memberships.map { |membership| membership.group.external_id }
+        Rails.logger.info("Skipping deleting user #{user.email} as they are a member of a group. Group IDs: #{group_ids}")
+        skipped_count += 1
+        next
+      end
+
+      Form.where(creator_id: user.id).find_all do |form|
+        Rails.logger.info("Deleting form with id #{form.id} for user #{user.email}")
```
stephencdaly, line 40: Check whether the user has any live forms? If so, skip the user?

lib/tasks/users.rake
```diff
@@ -12,6 +14,49 @@ namespace :users do
   task update_user_roles_to_standard: :environment do
     update_user_roles
   end
+
+  desc "Delete user accounts with no name or organisation"
+  task delete_users_with_no_name_or_org: :environment do
+    auth0_client = Auth0Client.new(
```
stephencdaly, line 20: After talking with @james-sheppard-gov-uk and @AP-Hunt, we don't want to grant access to the application in auth0 to do this deletion, as we don't currently have a way of rotating the client secret - so if it was leaked we want to limit the damage that can be done with it. So this task shouldn't attempt to delete data from auth0.

PR #1629, opened 2024-11-26
Fix error in routes#show for missing goto_page

thomasiles:
When a user has created a "basic route" and then deletes the page the condition points to with it's "goto_page_id", the show route throws an exception when the question_text cannot be accessed.

This commit fixes this error by using placeholder text when the goto_page cannot be found.

This is a short term fix intended to stop an error being raised, rather than a considered design fix.

## To recreate the original issue

1. create a selection question suitable for routing
2. create two questions of any type for the route to point to
3. add a route from the selection question to a page
4. delete that page
5. use the edit link on the route (which should be highlighted with an error "The question you’re taking the person to no longer exists")

The replacement text within the routes#show page:

image

Trello card: 

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1038,6 +1038,7 @@ en:
     edit: Edit
     if_answer_is: If the answer is
     page_name: "%{page_position}. %{page_name}"
+    page_name_not_exist: Edit this route to select a question or page you want to take the person to, or delete the route
     question_title: Question %{position}
     route_title: Route %{index}
     secondary_skip_after: Then after the person answers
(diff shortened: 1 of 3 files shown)


lfdebrux (approved):

config/locales/en.yml
```diff
@@ -1038,6 +1038,7 @@ en:
     edit: Edit
     if_answer_is: If the answer is
     page_name: "%{page_position}. %{page_name}"
+    page_name_not_exist: Edit this route to select a question or page you want to take the person to, or delete the route
```
hannahkc, line 1041:
```suggestion
    page_name_not_exist: Edit this route to select the question or page you want to take the person to, or delete the route
```

lfdebrux (approved):

PR #1631, opened 2024-11-26
Clean up after long lists

DavidBiddle:

Trello card: https://trello.com/c/BFWKD1ZG/2541-clean-up-after-long-lists-work

Includes:
- removing the `long_lists_enabled` database column and the task for setting it.
- removing the old selection type files and translations, since we don't need them any more
- renaming references to `long list selection` (since the distinction doesn't make sense now that the old selection files have been removed). We can just say `selection` instead
- renaming routes to remove references to `long_list_selection`
- renaming the `with_selections_settings` trait to the singular `with_selection_settings` for consistency

Needs doing elsewhere:
- Removing the logic that handles the old selection journey in forms-e2e-tests

**Don't merge this** until:
- We've sent out the comms for the long lists feature
- We're happy that we aren't going to revert the long lists feature

### Note on how big this PR is
This PR is big in terms of the number of files touched and the number of commits. The changes themselves are fairly straightforward (largely deletions and renames), but there are a lot of them. 

It would be easiest to go through it commit-by-commit.

If you reckon it's too unwieldy, let me know and I'll try to break it into a few PRs.


alice-carr (approved):
I've been through each commit and this all makes sense, + the commit messages are clear 👍

thomasiles (approved):
thanks for taking the time to rename things and tidy up and also chop things into sensible commits 🏅

PR #1636, opened 2024-11-28
Add delete actions to secondary skip controller

SamJamCul:
This adds delete and destroy actions to the secondary skip controller. These are used to remove a secondary skip from a question's routing, to complete the full CRUD functionality.

A separate delete_secondary_skip_input object has also been created, to handle the validations and submission for the delete page.

Trello card: https://trello.com/c/xMToaz6H/1992-add-page-to-delete-route-2-route-in-forms-admin

This activates the delete link that we added to the secondary route card. It now goes to a delete confirmation page, which has its own input and controller actions etc.

image


thomasiles (approved):
Works well for me and looks really neat 👍

PR #1638, opened 2024-11-29
Change conditions controller to use PageRepository everywhere

lfdebrux:

This was missed from PR #1587.


thomasiles (approved):
looks good

PR #1639, opened 2024-11-29
Remove unnecessary call to Page.new from questions controller

lfdebrux:

We want to abstract away as many calls to the Page model as possible using the page repository added in PR #1587.

While working on a ticket I noticed that we still have a call to `Page.new` in the `QuestionsController#new` action. I was about to add a `PageRepository.new` method, when I noticed that we don't actually need the page object in the view for that action.

This PR adds some tests for the new question page, as there weren't any already, and so we can check that replacing the page object with `nil` does not change the behaviour. As seen from the commits, that is the case, so we finally can delete the call from the controller.

### DIFF
--- a/app/controllers/pages/questions_controller.rb
+++ b/app/controllers/pages/questions_controller.rb
@@ -11,8 +11,6 @@ class Pages::QuestionsController < PagesController
                                                is_repeatable: draft_question.is_repeatable,
                                                draft_question:)
 
-    # TODO: Remove this once we have a check your question view. The new view should also pull data directly from draft_question instead of through page model
-    @page = Page.new(form_id: current_form.id)
     render :new, locals: { current_form:, draft_question: }
   end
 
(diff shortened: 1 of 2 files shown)


app/views/pages/new.html.erb
```diff
@@ -1,20 +1,20 @@
...
-      <span class="govuk-caption-l"><%= t("pages.question") %> <%= current_form.page_number(@page) %></span>
+      <span class="govuk-caption-l"><%= t("pages.question") %> <%= current_form.page_number(nil) %></span>
       <span class="govuk-visually-hidden"> - </span>
       <%= t("pages.heading") %>
     </h1>
 
     <%= render partial: 'form', locals: { is_new_page: true,
                                           form_object: current_form,
-                                          page_object: @page,
+                                          page_object: nil,
```
alice-carr, line 14: I may be misunderstanding how this works, but do we need to pass in the page_object at all here? removing it entirely the tests still pass

lfdebrux: Oh, you're right! I was thinking of these as function arguments and so needing to be defined, but they're not, so it's fine... I'll update the commit to remove that line.

alice-carr (approved):
makes sense, and works locally for me 👍

PR #1646, opened 2024-12-03
Add a dry run version of the org fetch task

SamJamCul:
Add a variant of the organisations:fetch rake task, which rolls the transaction back once it's been completed.

Trello card: https://trello.com/c/M81WBlnp/2021-update-organisations-table-in-forms-admin-database

The support ticket involves rerunning our `organisations:fetch` to get an up to date list of organisations. We don't currently have a dry run variant of this task, and because it involves using an external source, it seemed sensible to make sure it can successfully do a dry run before actually updating the database. 

The second commit is to prevent this from happening: 

image

### DIFF
--- a/lib/tasks/organisations.rake
+++ b/lib/tasks/organisations.rake
@@ -3,9 +3,22 @@ namespace :organisations do
   task fetch: :environment do
     include AdvisoryLock::DSL
 
-    with_lock("forms-admin:organisations:fetch") do
-      OrganisationsFetcher.new.call
-    end
+    Rails.logger.info("Starting organisation fetch")
+
+    run_organisation_fetch(rollback: false)
+
+    Rails.logger.info("Organisation fetch complete")
+  end
+
+  desc "Dry run for update organisations table using data from GOV.UK"
+  task fetch_dry_run: :environment do
+    include AdvisoryLock::DSL
+
+    Rails.logger.info("Starting dry run of organisation fetch")
+
+    run_organisation_fetch(rollback: true)
+
+    Rails.logger.info("Dry run of organisation fetch complete")
   end
 
   desc "Add organisation that is not in GOV.UK organisation database"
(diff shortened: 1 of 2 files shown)


lib/tasks/organisations.rake
```diff
@@ -8,6 +8,22 @@ namespace :organisations do
...
+  desc "Dry run for update organisations table using data from GOV.UK"
+  task fetch_dry_run: :environment do
+    include AdvisoryLock::DSL
+
+    ActiveRecord::Base.transaction do
+      puts("Beginning dry run organisation fetch")
+
+      with_lock("forms-admin:organisations:fetch") do
+        OrganisationsFetcher.new.call
+      end
```
stephencdaly, line 20: Even though this is a very small bit of logic, I reckon it's probably worth extracting into a method - just to make extra clear that this is a dry run testing the functionality of the `fetch` task, and if we update the `fetch` task this should be updated too.

SamJamCul: I've refactored it to abstract the task into a separate method - and the two tasks now share it. I've also added a bit of messaging output so that you can tell that something's happened, as before it was completely silent which I found a little unnerving.

lib/tasks/organisations.rake
```diff
@@ -3,9 +3,22 @@ namespace :organisations do
   task fetch: :environment do
     include AdvisoryLock::DSL
 
-    with_lock("forms-admin:organisations:fetch") do
-      OrganisationsFetcher.new.call
-    end
+    puts("Starting organisation fetch")
```
stephencdaly, line 6: I'm unsure whether there's a significant benefit to using `Rails.logger.info` rather than `puts`, but I think we've started using `Rails.logger.info` more in rake tasks. I'm not sure whether we actively decided this is the right thing to do though.
I think the main thing is just that if we use `puts` and don't add an assertion on the output in the tests, it will log to the console which is annoying - but that's not the case currently for the tests for this tasks so no major need to change.

stephencdaly (approved):
Ah you did something for the thing I was about to comment about before I could submit the review! Thanks for fixing the other test!

I left a comment about using `Rails.logger.info` for this specific reason - but this is fine to be merged as is.

lib/tasks/organisations.rake
SamJamCul: I'll swap it over, it does look more appropriate

SamJamCul: It's now using the logger, and the test has been updated to no longer worry about it

PR #1653, opened 2024-12-05
Use PageRepository in specs where appropriate

lfdebrux:

Trello card: https://trello.com/c/QuEeu8uI/2417-add-repository-service-to-forms-admin-for-form-page-model 

We want to mock repositories instead of forms-api requests in our specs as much as possible; this will make it easier to change the source of the model data in future.

This commit changes specs that were mocking HTTP requests to `/api/v1/forms/:form_id/pages/:page_id` so that they mock the corresponding PageRepository methods instead.


thomasiles (approved):
nice!

SamJamCul (approved):
Great job finding all of these, it's very fiddly!

PR #1658, opened 2024-12-09
Add a test for GDS Transport font loading

DavidBiddle:

Trello card: None

It's possible for changes to our build to cause assets not to be built properly. One of the problems this causes is a failure to load the GDS Transport font. There's a case of this on the latest Vite dependabot update: https://github.com/alphagov/forms-admin/pull/1656/files#r1873545007

This commit adds a test which will fail if the GDS Transport font is referenced in a font-face declaration but fails to load. This should reduce the risk of us accidentally merging a change which breaks our assets.


stephencdaly (approved):
Nice solution! I've confirmed that the test fails with the Vite v6 update.

PR #1659, opened 2024-12-09
Refactor controller actions and views for page delete confirmation

lfdebrux:

Trello card: https://trello.com/c/UThSZ8ga/2037-fix-behaviour-of-deleting-page-with-secondary-skip-route 

We want to be able to add some extra warnings and information to the confirmation page shown to users when they try to delete a question in a form.

However, to do this, we need to do some refactoring first, because currently the code for deleting pages is tangled up with the code for deleting forms.

This PR refactors the forms delete confirmation controller to deal with forms only, and moves the code for deleting pages to the pages controller; it is more idiomatic Rails to have the destroy and delete actions for a model in the controller for that model.

While working on this change I also noticed that there's an extra bit of hint text which is supposed to be shown to users when deleting a live form. It currently is not due to a typo, but it's also inaccurate as we haven't yet fully implemented deleting a form draft. So this PR just removes that hint text completely.

The changes in this PR should be purely refactors and not result in any user facing changes.

### DIFF
--- a/app/controllers/forms/delete_confirmation_controller.rb
+++ b/app/controllers/forms/delete_confirmation_controller.rb
@@ -61,7 +61,6 @@ module Forms
       else
         @url = destroy_form_path(current_form)
         @confirm_deletion_legend = t("forms_delete_confirmation_input.confirm_deletion")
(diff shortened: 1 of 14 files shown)


stephencdaly (approved):
Changes all look good, and I've tested locally and it works as expected. Thanks for doing this!!

PR #1663, opened 2024-12-11
Add warnings when deleting questions that are associated with routes

lfdebrux:

Trello card: https://trello.com/c/UThSZ8ga/2037-fix-behaviour-of-deleting-page-with-secondary-skip-route 

We've decided that we want to delete routes when a question is deleted, rather than leave the form in an invalid state.

As part of this we will show a notification banner about routes associated with the question to form creators when they are on the page to confirm that they want to delete a question. The content of the banner varies depending on the route associated; see the commits for details.

Note that currently the routes are not always actually deleted just yet, there needs to be a change to forms-api for that which is in https://github.com/alphagov/forms-api/pull/654.

<details><summary><h3>Screenshots</h3></summary>

#### Warning if the question is the start of a "route 1" route

https://admin.dev.forms.service.gov.uk/forms/12788/pages/15016/delete

admin dev forms service gov uk_forms_12788_pages_15016_delete

#### Warning if the question is the end of a "route 1" route

https://admin.dev.forms.service.gov.uk/forms/12788/pages/15017/delete

admin dev forms service gov uk_forms_12788_pages_15017_delete

#### Warning if the question is the end of a "route 2" route

https://admin.dev.forms.service.gov.uk/forms/12788/pages/15018/delete

admin dev forms service gov uk_forms_12788_pages_15018_delete

#### Warning if the question is the start of a "route 2" route

https://admin.dev.forms.service.gov.uk/forms/12788/pages/15023/delete

admin dev forms service gov uk_forms_12788_pages_15023_delete

</details>

### DIFF
--- a/spec/views/pages/delete.html.erb_spec.rb
+++ b/spec/views/pages/delete.html.erb_spec.rb
@@ -53,7 +53,17 @@ RSpec.describe "pages/delete" do
   end
 
   describe "when page to delete is the start of one or more routes" do
-    let(:page) { build :page, id: 1, form_id: 1, position: 2, routing_conditions: true }
+    let(:page) do
+      build(
+        :page,
+        id: 1,
+        form_id: 1,
+        position: 2,
+        routing_conditions: [
+          build(:condition, routing_page_id: 1, check_page_id: 1),
+        ],
+      )
+    end
 
     it "renders a notification banner" do
       expect(rendered).to have_css ".govuk-notification-banner"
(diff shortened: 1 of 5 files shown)


lfdebrux:
@dependabot rebase

spec/views/pages/delete.html.erb_spec.rb
```diff
@@ -39,4 +45,239 @@
       expect(rendered).not_to have_css ".govuk-hint"
     end
   end
+
+  describe "when page to delete is not associated with any routes" do
+    it "does not render a notification banner" do
+      expect(rendered).not_to have_css "govuk-notification-banner"
```
DavidBiddle, line 51: I think this is missing a dot which means it'll always pass:

```suggestion
      expect(rendered).not_to have_css ".govuk-notification-banner"
```

lfdebrux: Ah, very good spot, thank you! This should be fixed now.

PR #1665, opened 2024-12-13
Add task for deleting a user

DavidBiddle:

Trello card: 

Adds a new task for deleting an individual user, and a dry run task for checking it in advance.

Doesn't delete any associated forms or groups - this complicates things quite a lot and it's less risky for live forms if we delete or reassign those form/group associations prior to running this task to delete the user.

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -12,6 +12,18 @@ namespace :users do
   task update_user_roles_to_standard: :environment do
     update_user_roles
   end
+
+  desc "Delete user (dry run)"
+  task :delete_user_dry_run, %i[user_id] => :environment do |_, args|
+    run_deletion_task("delete_user_dry_run", args, rollback: true)
+  end
+
+  desc "Delete user"
+  task :delete_user, %i[user_id] => :environment do |_, args|
+    # This task doesn't check whether the user is associated with any forms or groups.
+    # Before running the task, remove the user from any groups and reassign their forms to another user.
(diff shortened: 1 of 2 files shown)


stephencdaly (dismissed):
Looks good - minor suggestion.

I think it might also be useful to add a code comment that this does not check for forms or groups for the user so that our future selves know we should check this before we try to delete a user. 

I think even better, it would check if either of these exist and abort the task if they do. The check should be fairly easy - there's some similar logic that exists in this spike PR https://github.com/alphagov/forms-admin/pull/1627/files

But that's probably overkill for this support task. I think a comment should be good enough.

lib/tasks/users.rake
```diff
@@ -12,6 +12,31 @@ namespace :users do
...
     update_user_roles
   end
+
+  desc "Delete user (dry run)"
+  task :delete_user_dry_run, %i[user_id] => :environment do |_, args|
+    usage_message = "usage: rake delete_user_dry_run[<user_id>]".freeze
+    abort usage_message if args[:user_id].blank?
+
+    ActiveRecord::Base.transaction do
+      user = User.find(args[:user_id])
```
stephencdaly, line 22: Minor - but do you reckon it's worth extracting the small amount of main logic into a private function shared between the dry-run and non-dry run task. If we want to change or adapt this it's helpful to have the code shared.

DavidBiddle: Good suggestion, have updated!

PR #1672, opened 2024-12-17
Update to Rails 8.0.1

DavidBiddle:

Trello card: 

Updated by:

- taking the initial Dependabot update commit
- running `bin/rails app:update` and reviewing the changes against RailsDiff and the commit history to avoid overwriting our deliberate changes
- fixing an issue caused by an APi deprecation


stephencdaly (approved):
Changes all look sensible enough to me and can't spot any issues when running locally. Nice work!

config/environments/production.rb
```diff
@@ -63,36 +48,50 @@
...
   config.i18n.fallbacks = true
 
-  # Don't log any deprecations.
-  config.active_support.report_deprecations = false
-
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
+  # Only use :id for inspections in production.
+  config.active_record.attributes_for_inspect = [:id]
```
chao-xian, line 87: Where did this change come from? It's part of https://railsdiff.org/7.2.2/8.0.1

DavidBiddle: Good question - it definitely came from rails app:update but looks like it was introduced in the v7.2.1 -> v7.2.2 bump: https://railsdiff.org/7.2.1/7.2.2

DavidBiddle: We do use `inspect` in one of our rake tasks so I think I need to remove this line or set it to `:all`

chao-xian: so we must have taken it out in the past?

DavidBiddle: Or we accepted the 7.2.2 dependabot bump without running the script

DavidBiddle: PR for removing it here anyway: https://github.com/alphagov/forms-admin/pull/1673

PR #1673, opened 2024-12-17
Remove inspect config added in error

DavidBiddle:

Trello card: 

Removes the `config.active_record.attributes_for_inspect` item set by the Rails app:update script - we use inspect in production for one of the organisation rake tasks, so this config change doesn't work for us.

I added this in error when making the update to Rails 8.0.1 - it was a change from an earlier Rails update.

Thanks @chao-xian for spotting this!


thomasiles (approved):
makes sense 👍

PR #1681, opened 2024-12-23
Firebreak - Add reports to download live form data as CSV files

stephencdaly:

Trello card: https://trello.com/c/mPtJhtWq/

Add 2 reports to get all data about live forms as a CSV, and to get all questions in all live forms as a CSV.

The intention of this is that it should make it easier to analyse data about forms and questions that exist on the platform without having to spend time to make dedicated reports. It should be quick and easy to add additional columns into these CSV reports as we add new features.

These reports call the forms-api "/api/v2/form-documents" endpoint added in https://github.com/alphagov/forms-api/pull/663 to get the form document objects for all live forms. We get the form documents rather than the form model as the form document represents the state of the live form that is served to form fillers, whereas the form model might have changes that have been made to a draft version of the form.

The forms-api endpoint supports pagination. Use this to avoid sending all form documents in a single large API response - default to 100 results per request page.

To make requests to forms-api, use Net:HTTP rather than ActiveResource. I attempted to use ActiveResource at first but this required an additional gem and some additional hacking at to be able to consume the headers for pagination.

Screenshot 2025-01-03 at 12 23 38

Screenshot 2025-01-03 at 15 02 06

Example CSVs:

[live_forms_report-2025-01-03 12_59_53 UTC.csv](https://github.com/user-attachments/files/18299886/live_forms_report-2025-01-03.12_59_53.UTC.csv)

[live_questions_report-2025-01-03 13_03_47 UTC.csv](https://github.com/user-attachments/files/18299897/live_questions_report-2025-01-03.13_03_47.UTC.csv)

Relies on forms-api pull request: https://github.com/alphagov/forms-api/pull/663


hannahkc:
Looks good from a content perspective. This will be so useful - thank you!

DavidBiddle (approved):
Code changes make sense to me, the commits are well structured & documented, and the report download works on my local machine 🎉

PR #1692, opened 2025-01-07
Remove dead code from DeleteConfirmationController

thomasiles:
There are two methods in the DeleteConfirmationController which don't seem to be used within the controller or elsewhere.

This commit removes them.


DavidBiddle (approved):
Deletions look sensible and the delete form logic stillk works for me locally 🎉

chao-xian (approved):

PR #1697, opened 2025-01-13
Reduce the max number of file upload questions to 4

stephencdaly:

Trello card: https://trello.com/c/7qrLP1gH/2067-restrict-file-uploads-to-5-questions-of-7mb-each

File attachments sent using SES are base64 encoded. The size limit for an email is 40MB after base64 encoding. Base64 encoding causes an overhead of 33–37% relative to the size of the original binary data. Therefore we are reducing the maximum number of file upload questions allowed in a form to 4. If there are 4 file upload questions with files uploaded with the maximum 7MB size, this will have a maximum total size after encoding of 1.37 x 4 x 7MB = 38.36MB.

Also update the content with a small tweak.


DavidBiddle (approved):
Changes look good and can see the new limit and error message working on my machine :tada:

PR #1705, opened 2025-01-15
Fix deleting users when they have one or more draft questions

lfdebrux:

Trello card: https://trello.com/c/qmQcoMsO/2061-7-jan-delete-the-accounts-with-no-name-or-organisation 

When running the task to delete users with no name or organisation set we saw errors when a user has some draft questions, because of a database constrain, see Sentry issue https://govuk-forms.sentry.io/issues/6220836647.

This commit fixes the error by making sure draft questions are deleted before the user.

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -103,6 +103,9 @@ def delete_users_with_no_name_or_org(dry_run: false)
       next
     end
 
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
I hadn't thought of this at all, good catch and write up

PR #1706, opened 2025-01-15
Switch to modern sass api

DavidBiddle:

Trello card: https://trello.com/c/pfzCTVnB/2066-investigate-asset-building-issue-with-modern-sass-api

We were seeing [an issue with assets failing to build properly using the 'modern' Sass API](https://trello.com/c/pfzCTVnB/2066-investigate-asset-building-issue-with-modern-sass-api). After investigating this appears to be caused by a bug in the Vite internal sass importer (https://github.com/vitejs/vite/issues/19196). 

This PR:
- uses the sass NodePackageImporter import govuk-frontend, which bypasses that bug and also offers us some other advantages (see commit message for more details)
- updates to use the modern API
- uses a more compact API for setting the asset paths


app/frontend/entrypoints/application.scss
```diff
@@ -1,8 +1,7 @@
-$govuk-images-path: "@govuk/assets/images/";
-$govuk-fonts-path: "@govuk/assets/fonts/";
+$govuk-assets-path: "@govuk/assets/";
 $govuk-global-styles: true;
 
-@import "@govuk/all";
+@import "pkg:govuk-frontend";
 @import "dfe-autocomplete/src/dfe-autocomplete";
```
lfdebrux, line 5: Can we also use NodePackageImporter with dfe-autocomplete?

lfdebrux (approved):

app/frontend/entrypoints/application.scss
DavidBiddle: Not as neatly (their package.json isn't currently configured to provide a Sass file, so we'll still have to configure the path to the file). I think it's worth doing anyway, for explicitness and consistency.

DavidBiddle: Not as neatly (their package.json isn't currently configured to provide a Sass file, so we'll still have to configure the path to the file). I think it's worth doing anyway, for explicitness and consistency.

DavidBiddle: Done!

PR #1710, opened 2025-01-16
Change delete button for routes to delete all routes for page

thomasiles:
## Let form creators delete all routes for a question

https://trello.com/c/KA2MdRKW/2036-change-delete-routes-button-to-delete-all-routes-for-a-page

https://app.mural.co/t/gaap0347/m/gaap0347/1728478914347/6a7e0b709f50f00f81cab37414fdf44ec8601a70

The delete button on the routes#show page now takes the user to a confirmation screen for removing all the routes associated with the page, not just the primary route.

# screenshots
image

image

image

### DIFF
--- a/app/controllers/pages/routes_controller.rb
+++ b/app/controllers/pages/routes_controller.rb
@@ -3,4 +3,29 @@ class Pages::RoutesController < PagesController
     back_link_url = form_pages_path(current_form.id)
     render locals: { current_form:, page:, pages: current_form.pages, back_link_url: }
   end
+
+  def delete
+    @delete_confirmation_input = Pages::Routes::DeleteConfirmationInput.new
+    render locals: { current_form:, page:, delete_confirmation_input: @delete_confirmation_input }
+  end
+
+  def destroy
+    @delete_confirmation_input = Pages::Routes::DeleteConfirmationInput.new(delete_confirmation_input_params)
+
+    if @delete_confirmation_input.submit
+      if @delete_confirmation_input.confirmed?
+        return redirect_to form_pages_path, success: t("banner.success.page_routes_deleted", question_position: current_form.page_number(page))
+      end
+    else
+      return render :delete, locals: { current_form:, page:, delete_confirmation_input: @delete_confirmation_input }, status: :unprocessable_entity
+    end
+
+    redirect_to show_routes_path(form_id: current_form.id, page_id: page.id)
+  end
+
+private
+
+  def delete_confirmation_input_params
+    params.require(:pages_routes_delete_confirmation_input).permit(:confirm).merge(form: current_form, page: page)
+  end
 end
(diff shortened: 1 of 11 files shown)


app/controllers/pages/routes_controller.rb
```diff
@@ -3,4 +3,40 @@ def show
...
+private
+
+  def delete_routes
+    all_form_routing_conditions = current_form.pages.flat_map(&:routing_conditions).compact_blank
+    page_routes = all_form_routing_conditions.select { |rc| rc.check_page_id == page.id }
+    page_routes.each do |rc|
+      rc.prefix_options[:form_id] = current_form.id
+      rc.prefix_options[:page_id] = rc.routing_page_id
+      ConditionRepository.destroy(rc)
+    end
```
thomasiles, line 36: This doesn't feel like the best place for this. I'm not sure where is though?

We need to get the form, then every page and then delete the conditions.

It could go in one of the repositories but I'm not 100% sure which one is the best choice. 

It could be a separate service or a new RouteRepository itself?

lfdebrux: The idea of a route repository is interesting!

I wonder about putting it in the form input object though, that might make sense? I guess we would need a new one, but it could inherit from `DeleteConfirmationInput` to avoid too much code duplication?

app/controllers/pages/routes_controller.rb
```diff
@@ -3,4 +3,40 @@ def show
...
+  def delete
+    @delete_confirmation_input = Pages::Routes::DeleteConfirmationInput.new
+    render locals: { current_form:, page:, delete_confirmation_input: @delete_confirmation_input }
+  end
+
+  def destroy
+    @delete_confirmation_input = Pages::Routes::DeleteConfirmationInput.new(delete_confirmation_input_params)
+
+    if @delete_confirmation_input.invalid?
+      return render :delete, locals: { current_form:, page:, delete_confirmation_input: @delete_confirmation_input }
```
lfdebrux, line 16: This should probably have `status: :unprocessable_entity`?

config/locales/en.yml
```diff
@@ -1145,6 +1150,11 @@ en:
       title: Add and edit your questions
     optional: "%{question_text} (optional)"
     question: Question
+    routes:
+      delete:
+        back: Back to question %{page_number}’s routes
+        caption: Question %{page_number}'s routes
```
hannahkc, line 1156:
```suggestion
        caption: Question %{page_number}’s routes
```

thomasiles:
I've moved the delete routes function and added a better test for it.

I've also corrected all the dodgy punctuation. I swear I check them every time and every time I get them wrong.

DavidBiddle (approved):
The code changes look good to me and I can see the deletion happening as expected when I run this locally 🎉

PR #1712, opened 2025-01-16
Update file upload content

jamie-o-wilkinson:

Trello card: https://trello.com/c/gg7uqjma/2069-update-content-in-production-for-muscle-version-v2

Adds file upload specific content to the edit question page when the answer type is "File upload". We still need to update the error messages for the question text field to match the new label, but this will be done in a subsequent PR.

Adds additional information about file upload questions at the top of the page.

Removes the ability to specify that the question can be answered more than once, as this is not currently supported for file upload questions.

Screenshot 2025-01-17 at 12 58 13

### DIFF
--- a/app/views/pages/_form.html.erb
+++ b/app/views/pages/_form.html.erb
@@ -5,7 +5,10 @@
     <p class="govuk-body"><%=t('helpers.label.pages_question_input.file_body_html')%></p>
   <% end %>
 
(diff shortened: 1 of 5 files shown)


app/views/pages/_form.html.erb
```diff
@@ -34,7 +41,7 @@
     <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible" />
   <% end %>
 
-  <% if question_input.answer_type == "selection" %>
+  <% if question_input.answer_type == "selection" || question_input.answer_type == "file" %>
```
DavidBiddle, line 44: If this gets any longer we might want to handle it in a method, but I think this is OK for now!

DavidBiddle (approved):
Looks good and I can see the changes working locally 🎉

PR #1725, opened 2025-01-22
Add FormRepository pages method

SamJamCul:

Trello card: https://trello.com/c/XJVoEeTw/2075-add-form-pages-to-form-repository

Adds a FormRepository `pages` method, which requests a form's pages via the form model. This is then used throughout the app to replace any requests for form pages that we're currently making.

It was a little fiddly to find all the places we're getting form pages, and the way activeresource works makes testing that we're actually stubbing out a request is also quite fiddly. 

There may be some tests where technically we could have removed the stubbing, and they still would have worked, as the way we factory build forms means they come with pages already, so a call for `.pages` wouldn't result in an API request. I've tried to still replace any possible circumstances where we could have made an API request with the FormRepository version. 

I'm also not sure if there'll be a performance hit with some of this, mainly around the conditions stuff. There are a few places where we're looking at a form's pages, and if each of those hits the API, we might (big might) see some slowdown. I've tried testing locally which was fine, but I'm very tempted to test on dev to see if it's fine.

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -11,7 +11,7 @@ class FormsController < ApplicationController
 
   def mark_pages_section_completed
     authorize current_form, :can_view_form?
-    @pages = current_form.pages
+    @pages = FormRepository.pages(current_form)
     @mark_complete_input = Forms::MarkPagesSectionCompleteInput.new(mark_complete_input_params)
 
     if @mark_complete_input.submit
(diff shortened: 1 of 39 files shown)


SamJamCul:
Just realised I've rebased on top of a slightly out of date PR, changing that now

lfdebrux (approved):
This works for me locally, thanks for putting it together.

PR #1727, opened 2025-01-24
Re-add linux-musl to Bundler platforms

lfdebrux:

In commit de10f10 we added *-linux-musl platforms to the Bundler lockfile, because the Alpine docker images we use in production are based on musl and a Nokogiri update started differentiating between linux-gnu and linux-musl platforms.

However, in commit 8f725b3 a Depdenabot update removed those lines.

This PR re-adds them, hopefully we can also find a way to stop Dependabot from removing them. I also tried to clean up the list of platforms by running `bundle lock --normalize-platforms`, which required first updating the version of Bundler used.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -276,11 +276,15 @@ GEM
     nio4r (2.7.4)
     nokogiri (1.18.1-aarch64-linux-gnu)
       racc (~> 1.4)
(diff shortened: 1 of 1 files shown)


SamJamCul (approved):
Will keep an eye on dependabot in case it does any more nefarious deletions

PR #1728, opened 2025-01-24
Move ActiveResource models to resources/api/v1

lfdebrux:

Trello card: https://trello.com/c/xotu6KNy/1995-add-tables-for-form-models-to-forms-admin 

We want to move tables with form, page, and condition records from the forms-api app to the forms-admin app. When we do this we will want the models in this app to use ActiveModel instead of ActiveResource. To make that future change easier, this PR moves the ActiveResource models classes to a different namespace.


DavidBiddle (approved):
Looks good and things still seem to work locally for me 🎉 

Fairly big PR so I really appreciated the small and well-explained commits!

PR #1645, opened 2024-12-03
Add file answer type

stephencdaly:
### What problem does this pull request solve?

Trello card: https://trello.com/c/icJPiPHn/

Adds a column `file_upload_enabled` to the `groups` table so that we can enable creating file upload questions just for specific groups while we are developing the feature.

Allow creating questions with answer type "file" if the flag is enabled.

Relies on https://github.com/alphagov/forms-api/pull/644

Screenshot 2024-12-03 at 10 18 14

### Things to consider when reviewing

- Ensure that you consider the wider context.
- Does it work when run on your machine?
- Is it clear what the code is doing?
- Do the commit messages explain why the changes were made?
- Are there all the unit tests needed?
- Has all relevant documentation been updated?

PR #1729, opened 2025-01-27
Add group scoped feature flag for branching

thomasiles:
### Add new Group scoped feature flag for branching

Trello card: https://trello.com/c/GgbE96M1/2083-enable-branch-routing-feature-in-production-for-an-internal-test-group

To begin with, we want to release the advanced branching features to a small number of users.

To do this, we are enabling some Groups to have the feature flag enabled, instead of enabling it for the whole environment. We still want the flag to be enabled in dev for all users and groups.

This PR is heavily based off the changes @stephencdaly and feature team one did for [file upload](https://github.com/alphagov/forms-admin/pull/1645).

It's more complicated because the branching feature flag is used in more places, mainly views.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -88,6 +88,10 @@ class PagesController < ApplicationController
     redirect_to form_pages_path, success: t("banner.success.form.page_moved", question_text: page_to_move.question_text, direction: move_params[:direction], position:)
   end
 
+  def branching_enabled
+    @branching_enabled ||= Settings.features.branch_routing || current_form.group.branching_enabled?
+  end
+
 private
 
   def clear_draft_questions_data
--- a/app/controllers/pages/routes_controller.rb
+++ b/app/controllers/pages/routes_controller.rb
@@ -1,7 +1,7 @@
 class Pages::RoutesController < PagesController
   def show
     back_link_url = form_pages_path(current_form.id)
-    render locals: { current_form:, page:, pages: FormRepository.pages(current_form), back_link_url: }
+    render locals: { current_form:, page:, pages: FormRepository.pages(current_form), back_link_url:, branching_enabled: }
   end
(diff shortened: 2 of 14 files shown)


app/controllers/pages_controller.rb
```diff
@@ -88,6 +88,10 @@ def move_page
     redirect_to form_pages_path, success: t("banner.success.form.page_moved", question_text: page_to_move.question_text, direction: move_params[:direction], position:)
   end
 
+  def branching_enabled
+    @branching_enabled ||= Settings.features.branch_routing || current_form.group.branching_enabled?
+  end
+
```
thomasiles, line 94: This could be a concern to stop it cluttering the page controller.

I think for what will be a short lived feature flag, it's simpler to be explicit.

happy to hear other views though.

lfdebrux: Could we instead extend the `FeatureService#enabled?` method to optionally accept a group?

jamie-o-wilkinson: @lfdebrux I think we previously did something similar to that (though I can't actually remember for which feature 😅 ). 

I think it made things tricky across environments, as the ids obviously don't match up. This meant that to test in dev you'd be using an id that might match an entirely different form in production, for example. I think @stephencdaly and I concluded it was nicer storing whether the feature was enabled on the model itself, as then you can freely test in different environments without causing issues.

lfdebrux: Sorry, I meant that passing a group to `enabled?` would check whether `group.send("#{feature}_enabled")` was true, or similar. Or if a form was passed it would look at `form.group`.

thomasiles: I've replaced this PR with [another which tackles this issue](https://github.com/alphagov/forms-admin/pull/1737)

PR #1730, opened 2025-01-27
Rename host logging field

stephencdaly:

Rename the "host" logging field to "hostname". An additional "host" field is added to the logs that go to Splunk for Kenesis. This means we end up with 2 values for "host" making it difficult to filter searches by it.


cadmiumcat (dismissed):
Thanks! 
I've deployed this to dev, the logs in splunk include the `hostname` and no `host` as expected

lfdebrux (approved):

PR #1737, opened 2025-01-29
Add fflag groups enabled

thomasiles:
### Add a feature flag for enabling advanced branching for specified groups

Trello card: https://trello.com/c/GgbE96M1/2083-enable-branch-routing-feature-in-production-for-an-internal-test-group

To begin with, we want to release the advanced branching features to a small number of users.

To do this, we are enabling some Groups to have the feature flag enabled, instead of enabling it for the whole environment. We still want the flag to be enabled in dev for all users and groups.

This PR adds a new way to specify that a feature flag value should be taken per group rather than set for the whole environment. See the commits for details.

This PR is heavily based off the changes @stephencdaly and feature team one did for [file upload](https://github.com/alphagov/forms-admin/pull/1645).

It's more complicated because the branching feature flag is used in more places, mainly views.

This PR replaces https://github.com/alphagov/forms-admin/pull/1729

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -114,7 +114,7 @@ class ApplicationController < ActionController::Base
   end
 
   def groups_enabled
-    @groups_enabled ||= current_user.present? && FeatureService.new(current_user).enabled?(:groups)
+    @groups_enabled ||= current_user.present? && FeatureService.new(user: current_user).enabled?(:groups)
   end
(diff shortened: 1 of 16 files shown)


lfdebrux:
Nice, thanks for making it work with the `FeatureService`.

I've tested this locally and it works for me 👍

lfdebrux (approved):

PR #1740, opened 2025-01-30
Add tables for forms models

lfdebrux:

Trello card: https://trello.com/c/xotu6KNy/1995-add-tables-for-form-models-to-forms-admin 

We want to migrate the records for forms conditions and pages to this app from forms-api.

This PR adds the tables we'll need in the database.

I've purposely not added models to use those tables, as for now we want the tables to be empty and unused.


DavidBiddle (approved):
Code looks good and commits are well documented. 

Have run the migration locally and confirmed that it runs OK and the resulting schema is the same as in forms-api for each of the new tables.

PR #1742, opened 2025-01-30
Make it easier to find where and how to add second route

thomasiles:
### Make it easier to find where and how to add second route

The description from the trello card explains what's changing well.

We want to make the following changes to the 'Add a route from a question' page:

- Change the guidance content, question, hint and error text to match the [new design in mural ](https://app.mural.co/t/gaap0347/m/gaap0347/1728478914347/6a7e0b709f50f00f81cab37414fdf44ec8601a70?wid=0-1737024454887)(also attached)

- Include all radio questions that a route can be added to in the radio option list (including questions that already have a route 1 added)

- Change the routing so that if a question that already has a route 1 is selected, the next page will be the ‘Question x’s routes' page (rather than the ‘Edit route 1’ page)

- Update the alternate [content that is shown if the form cannot have any routes added to any question yet](https://app.mural.co/t/gaap0347/m/gaap0347/1728478914347/6a7e0b709f50f00f81cab37414fdf44ec8601a70?wid=1-1737370812850) 

- Update the alternate [content that is shown if there are no more questions that can have a route added to them](https://app.mural.co/t/gaap0347/m/gaap0347/1728478914347/6a7e0b709f50f00f81cab37414fdf44ec8601a70?wid=25-1737370812850)

This is a bit complicated by the need to keep both existing and new versions going with the feature flag.

When the flag is enabled:
The new add route content:
image

..with error:
image

When there are no questions which can have a route:
image

When there are no more questions which can have routes added to:
image

### DIFF
--- a/app/controllers/pages/conditions_controller.rb
+++ b/app/controllers/pages/conditions_controller.rb
@@ -2,13 +2,13 @@ class Pages::ConditionsController < PagesController
   before_action :can_add_page_routing, only: %i[new create]
 
   def routing_page
-    routing_page_input = Pages::RoutingPageInput.new(routing_page_id: params[:routing_page_id])
+    routing_page_input = Pages::RoutingPageInput.new({ routing_page_id: params[:routing_page_id] }, branch_routing_enabled:)
     render template: "pages/conditions/routing_page", locals: { form: current_form, routing_page_input: }
   end
 
   def set_routing_page
     routing_page_id = params[:pages_routing_page_input][:routing_page_id]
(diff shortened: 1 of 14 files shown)


lfdebrux (approved):
These changes work for me.

I've tested locally with the feature flag off, comparing the pages between this branch and the main code in production, and it looks like the behaviour is the same, which is what we want.

Changes with the feature flag on look good too. 

I agree there are some places that could be improved by refactoring, but we can do that in subsequent PRs.

PR #1744, opened 2025-01-31
Get list of page routes in routes controller show action

lfdebrux:

Trello card: https://trello.com/c/6qRDmaOg/2095-changes-to-question-xs-routes-page-to-make-it-clearer 

Change the routes/show action, view and RouteSummaryCardDataPresenter to get page routes up front and pass down the chain to other objects. This lets us use the list of routes in places other than the presenter.

### DIFF
--- a/spec/services/page_conditions_service_spec.rb
+++ b/spec/services/page_conditions_service_spec.rb
@@ -1,0 +1,57 @@
+require "rails_helper"
+
+RSpec.describe PageConditionsService do
+  subject(:page_conditions_service) do
+    described_class.new(form:, pages:, page:)
+  end
+
+  let(:form) do
+    build(:form)
+  end
+
+  include_context "with pages with routing"
+
+  describe "#check_conditions" do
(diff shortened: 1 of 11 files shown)


spec/services/page_conditions_service_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+RSpec.describe PageConditionsService do
+  subject(:page_conditions_service) do
+    described_class.new(form:, pages:, page:)
+  end
+
+  let(:form) do
+    build(:form)
+  end
+
+  include_context "with pages with routing"
```
thomasiles, line 12: nice!

thomasiles (approved):
Looks much neater, works locally and has some really nice ideas 👏

spec/services/page_conditions_service_spec.rb
lfdebrux: Thanks!

PR #1745, opened 2025-01-31
Review apps terraform

AP-Hunt:

Trello card: https://trello.com/c/u0AOLJYF/579-terraform-for-deploying-forms-admin-to-integration-ecs-cluster

Introduces a small amount of Terraform into the repository for use in deploying a `forms-admin` review app. Alongside the Terraform is a GitHub Actions workflow for testing and validating it.

This is a large PR, and is expected to not be the final product.

Future PRs will introduce a GitHub Actions workflow for applying the Terraform in response to new pull requests.


lfdebrux (approved):
I think this makes sense to me broadly. I wonder if at some point we'll be able to abstract some of it into a shared module that lives outside this repo, but obviously for now while we're still in the early days it makes sense to have all the configuration living here.

PR #1746, opened 2025-02-03
Update add a route question

hannahkc:

Updated the question on the 'Add a route from a question' page to match the new designs. 


lfdebrux (approved):
Changes look good to me, thanks for fixing this

PR #1747, opened 2025-02-03
Change where add a question route goes if selected question already has a route

lfdebrux:

We've decided that if a question already has a route we should take the user to the 'Question x's routes' page instead of straight to the settings for the secondary skip, to better summarise the current situation of routes for the page - before the new one is added - and to future proof the journey when you can add multiple routes.

<h3>Screen recording</h3>

Screen recording of the new behaviour:

https://github.com/user-attachments/assets/3aa9c882-9736-4399-a332-af9beba60f4c


hannahkc:
This looks to be working correctly now from the screen recording - thanks for fixing this!

alice-carr (approved):

PR #1752, opened 2025-02-06
Add warning when user is about to delete condition with secondary skip

lfdebrux:

Trello card: https://trello.com/c/qcaW0Bv1/2107-add-a-warning-to-the-delete-route-page-that-shows-when-there-is-a-secondary-skip-route-for-the-same-question 

Currently (when branching is enabled) deleting a skip route for a question that has a secondary skip route puts the form in a state where there is a route with an error that can’t be deleted.

We do want users to be able to delete individual routes for a question.

So we’ve agreed that we will delete the secondary skip for a question when deleting any other routes for that question. And we’ll let the user know before the delete a route, with a warning.

This PR does some refactoring and adds the warning; note that the deletion will take place in forms-api, see alphagov/forms-api#686.


hannahkc:
Content in the banner looks good! thanks!

PR #1753, opened 2025-02-06
Change "Question x's routes" page

lfdebrux:

Trello card: https://trello.com/c/6qRDmaOg/2095-changes-to-question-xs-routes-page-to-make-it-clearer 

Tweak the design and layout of the "Question x's routes" page.

### Screen recording

Screen recording of adding a secondary skip route to a form via the edit routes page for an existing route.

https://github.com/user-attachments/assets/2c926a70-f07e-4f74-a739-88a522fbb35a


thomasiles (approved):
This is really nice and works locally 🏅

PR #1754, opened 2025-02-06
Change labels for routes in page list

lfdebrux:

Trello card: https://trello.com/c/GyEfMdbM/2122-change-labels-for-routes-on-add-and-edit-your-questions-page 

### Screenshots

Screenshot of "Add and edit your questions" page with these changes

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1021,7 +1021,7 @@ en:
     condition_goto_page_check_your_answers: "“Check your answers before submitting”."
     condition_goto_page_text: "%{goto_page_position}, “%{goto_page_text}”"
     condition_goto_page_text_with_errors: "[Question not selected]"
-    condition_name: Route %{route_index} of question %{page_index}’s routes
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -1021,7 +1021,7 @@ en:
     condition_goto_page_check_your_answers: "“Check your answers before submitting”."
     condition_goto_page_text: "%{goto_page_position}, “%{goto_page_text}”"
     condition_goto_page_text_with_errors: "[Question not selected]"
-    condition_name: Route %{route_index} of question %{page_index}’s routes
+    condition_name: Question %{page_index}’s route
```
thomasiles, line 1024: In the design it says `Question 2's routes` rather than `Question 2's route`. 

Should it be 'routes' instead of 'route'?

lfdebrux: Oops, yes it does, good spot, thanks!

thomasiles (approved):
nice!

PR #1762, opened 2025-02-07
Change name of default route/secondary skip route to "any other answer"

lfdebrux:

Trello card: https://trello.com/c/yW68wpBI/2121-change-route-2-name-to-route-for-any-other-answer-in-several-places 

We think it will make it clearer what the secondary skip route is to call it the "any other answer" route.

### Screenshots

Screenshot of the "Question x's routes" page with a summary card for a secondary skip route

Screenshot of the page to set questions to skip for secondary skip route

Screenshot of the page to confirm deletion of a secondary skip route


thomasiles (approved):
looks good

PR #1763, opened 2025-02-07
Add rake task to rename an organisation

chao-xian:

Trello card: https://trello.com/c/amETo48v/2120-live-forms-report-needs-amending


spec/lib/tasks/organisations.rake_spec.rb
```diff
@@ -67,4 +67,35 @@
...
+
+      expect(test_org.reload).to have_attributes(name: "Department for Testing and Validation")
+    end
+
+    it "does not rename an org from GOV.UK API" do
+      test_org = create :organisation, name: "Department for Testing", slug: "dft", govuk_content_id: Faker::Internet.uuid
+      test_org.clear_changes_information
+
+      expect { task.invoke("Department for Testing", "Department of Nope") }
+        .to output(/is from GOV.UK API/).to_stdout
```
lfdebrux, line 92: I think we also want to catch the SystemExit error from this test, to stop the tests from being aborted. Also, the test for the output needs tweaking slightly:

```suggestion
      expect { task.invoke("Department for Testing", "Department of Nope") }
        .to output(/is from the GOV.UK API/).to_stderr
        .and raise_error(SystemExit) { |e| expect(e).not_to be_success }
```

lfdebrux (approved):

jamie-o-wilkinson (approved):

PR #1767, opened 2025-02-11
Allow database seeds to be run in a review environment

AP-Hunt:

Trello card: https://trello.com/c/u0AOLJYF/579-terraform-for-deploying-forms-admin-to-integration-ecs-cluster

We need to be able to apply database seed data in review environments, but review environments use `RAILS_ENV=production` which prevented this from happening. Instead, we enable it for review environments also.


lfdebrux (approved):
Nice!

PR #1769, opened 2025-02-11
Rewrite database seed to not rely on FactoryBot

lfdebrux:


chao-xian (dismissed):
I dropped my database and recreated then ran db:seed - all seems fine

DavidBiddle (approved):
Ran `bin/rails db:reset` and things all look as expected 🎉

jamie-o-wilkinson:
I appreciate the new names. What a good place to be 😄

PR #1770, opened 2025-02-12
Stop showing question straight after page to skip from in list of pages to skip to

lfdebrux:

Trello card: https://trello.com/c/tYdE1AMd/2108-p2-stop-users-from-setting-up-route-to-the-next-page 

We've seen users in user research expect that trying to set up a route to the page straight after the page to route from would show an error. So we think it makes sense to not include the question straight after in the select options.

### Screen recording

Recording of adding a question route and editing a question route when the goto page has been moved to be straight after the routing page:

https://github.com/user-attachments/assets/abc73040-1099-4b4f-b8e5-c933d7299298

### DIFF
--- a/spec/requests/pages/conditions_controller_spec.rb
+++ b/spec/requests/pages/conditions_controller_spec.rb
@@ -134,18 +134,15 @@ RSpec.describe Pages::ConditionsController, type: :request do
   end
 
   describe "#create" do
+    let(:params) { { pages_conditions_input: { routing_page_id: 1, check_page_id: 1, goto_page_id: 3, answer_value: "Wales" } } }
+
     before do
       selected_page.id = 1
 
       allow(PageRepository).to receive(:find).and_return(selected_page)
+      allow(ConditionRepository).to receive(:create!).and_invoke(->(**attributes) { build :condition, **attributes })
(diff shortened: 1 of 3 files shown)


spec/requests/pages/conditions_controller_spec.rb
```diff
@@ -198,9 +197,12 @@
...
 
-      allow(Pages::ConditionsInput).to receive(:new).and_return(conditions_input)
-      allow(conditions_input).to receive(:check_errors_from_api)
-      allow(conditions_input).to receive(:assign_condition_values).and_return(conditions_input)
+      allow(Pages::ConditionsInput).to receive(:new).and_wrap_original do |original_method, *args, **kwargs|
+        @conditions_input = original_method.call(*args, **kwargs)
+        allow(conditions_input).to receive(:check_errors_from_api).and_call_original
+        allow(conditions_input).to receive(:assign_condition_values).and_call_original
+        conditions_input
+      end
```
thomasiles, line 205: nice! I'd never thought of this issue and fix for the tests before.

thomasiles (approved):
Nicely structured and works well locally 🏅

PR #1771, opened 2025-02-12
GitHub Actions workflow for review apps

AP-Hunt:

Trello card: https://trello.com/c/veseFayX/581-github-actions-workflows-for-managing-review-apps

Implements the GitHub Actions workflows needed for the full lifecycle of a `forms-admin` review app. This is a large pull request necessarily, because it combines multiple streams of work that came before it, and addresses any and all integration problems that arise.

@sarahseewhy and I have paired on this extensively.


sarahseewhy (approved):
Nothing immediately stands out as needing to change, although we did pair on this quite a bit so maybe a fresh pair of eyes would be useful.

🪄

PR #1774, opened 2025-02-13
Fix delete all routes journey

lfdebrux:

Trello card: https://trello.com/c/cFZnCzve/2127-bug-delete-all-routes-redirects-to-a-page-not-found-error 

Currently when a user tries to delete all routes for a question they are shown a page not found error rather being redirected to the "Add and edit questions" page as expected.

I think this is happening because of a change in forms-api that means deleting a condition can also delete secondary skip conditions in the same request (alphagov/forms-api#686).

This PR adds a test that simulates the behaviour of forms-api currently, and changes the `destroy` methods for each repository so that trying to delete a condition that has already been deleted doesn't result in an exception. This behaviour is closer to how ActiveRecord behaves, and fixes the problem so that users won't see the error page any more.

### DIFF
--- a/app/services/condition_repository.rb
+++ b/app/services/condition_repository.rb
@@ -30,6 +30,7 @@ class ConditionRepository
       condition = Api::V1::ConditionResource.new(record.attributes, true)
       condition.prefix_options = record.prefix_options
       condition.destroy # rubocop:disable Rails/SaveBang
+      record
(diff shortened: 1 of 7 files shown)


thomasiles (approved):
Fixing it across all the repositories is a really nice idea 👏

PR #1779, opened 2025-02-17
Add task to list groups which have a feature flag enabled

lfdebrux:

There isn't an easy way to see what feature flags are enabled for which groups currently; this PR adds a very simple Rake task to help with this.


stephencdaly (approved):
Nice! This will be useful. I tested it locally and it works as expected.

PR #1785, opened 2025-02-18
Update headings for page to add and edit a condition

lfdebrux:

Trello card: https://trello.com/c/faPNxjxY/2133-update-routing-pages-to-match-designs-snags-ticket 

We can hardcode as "Route 1" for now as questions can only have up to one route and one any other answer route.


hannahkc:
Love it!

thomasiles (approved):

PR #1789, opened 2025-02-19
Fix back link for edit route page

lfdebrux:

Trello card: https://trello.com/c/faPNxjxY/2133-update-routing-pages-to-match-designs-snags-ticket 

Now that the journey to get to the edit route page goes through the show routes page, the back link should take the user back to the show routes page.

### Screen recording

https://github.com/user-attachments/assets/f82a2e73-5b58-40d7-8eb9-ff2cfd233969


thomasiles (approved):
nice 👍

PR #1796, opened 2025-02-24
Refactor translation keys

lfdebrux:

I have been looking at the translations files and have noticed that we are inconsistent in how we name the variable for the number of a question… this PR makes the keys consistent for `question_number`, `question_text`, and `route_number`.


thomasiles (approved):
This is amazing, really appreciate the care you've taken over the names 🏅

PR #1803, opened 2025-02-25
.review_apps: upgrade to PostgreSQL 16.6

AP-Hunt:

Trello card: https://trello.com/c/E2d9wSBD/612-upgrade-postgresql-13-16

In a couple of weeks we're going to upgrade from PostgreSQL 13.3 to 16.6. We can upgrade the version in the review apps ahead of time, and this will serve as a useful test for anything that doesn't work under the new version.


sarahseewhy (approved):
Let's experiment 🔬 !

PR #1804, opened 2025-02-25
Change review apps PR workflow to leave only one comment

lfdebrux:

The way we work with GitHub in this repo means we need to rebase our PRs often; this on top of the usual back and forth when reviewing means that review apps might get deployed several times before a PR is merged. This is fine, except that every time the review apps are deployed there is a new comment left by the workflow, which is a bit noisy.

This PR changes the workflow to first delete any existing comments about review apps (similar to what our other sonarqube cloud bot does), so that there should only be one comment about review apps on a PR at a time, but it keeps its useful place in the timeline.

### DIFF
--- a/.github/workflows/review_apps_on_pr_change.yml
+++ b/.github/workflows/review_apps_on_pr_change.yml
@@ -76,9 +76,9 @@ jobs:
 
       - name: Comment on PR
         env:
+          COMMENT_MARKER: <!-- review apps on pr change -->
           GH_TOKEN: ${{ github.token }}
         run: |
-
           cat <<EOF > "${{runner.temp}}/pr-comment.md"
           :tada: A review copy of this PR has been deployed! You can reach it at ${{steps.deploy.outputs.REVIEW_APP_URL}}.
 
@@ -87,6 +87,13 @@ jobs:
           to debug, or otherwise ask an infrastructure person.
 
           For more details please see the [review app wiki page](https://github.com/alphagov/forms-team/wiki/Review-apps)
+
+          $COMMENT_MARKER
           EOF
+
+          old_comment_ids=$(gh api "repos/{owner}/{repo}/issues/${{github.event.pull_request.number}}/comments" --jq 'map(select((.user.login == "github-actions[bot]") and (.body | endswith($ENV.COMMENT_MARKER + "\n")))) | .[].id')
+          for comment_id in $old_comment_ids; do
+            gh api -X DELETE "repos/{owner}/{repo}/issues/comments/${comment_id}"
+          done
 
           gh pr comment "${{github.event.pull_request.html_url}}" --body-file "${{runner.temp}}/pr-comment.md"


AP-Hunt:
It's interesting that you thought of this too. Sarah and I thought about it was be built this and concluded "if it's needed it'll become evident".

The failing deployment is a permissions issue I need to fix from the `forms-deploy` side

AP-Hunt:
This script isn't quite working, evidenced by the (currently!) two comments in this thread. I ran the script through Shellcheck and it pointed out why

```
old_comment_ids=$(gh api "repos/{owner}/{repo}/issues/${{github.event.pull_request.number}}/comments" --jq 'map(select((.user.login == "github-actions[bot]") and (.body | endswith($ENV.COMMENT_MARKER)))) | .[].id')
                                                                                                           ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.

For more information:
  https://www.shellcheck.net/wiki/SC2016 -- Expressions don't expand in singl...
```

If you want an env var available in a JQ expression you should do. 

```
jq --arg name "${ENV_VAR}" 'expression using $name' input.json
```

Shellcheck says to use double quotes, but we generally avoid double quotes around JQ expressions because of the frequency with which we use double quotes in them.

lfdebrux:
> This script isn't quite working, evidenced by the (currently!) two comments in this thread. I ran the script through Shellcheck and it pointed out why
> 
> ```
> old_comment_ids=$(gh api "repos/{owner}/{repo}/issues/${{github.event.pull_request.number}}/comments" --jq 'map(select((.user.login == "github-actions[bot]") and (.body | endswith($ENV.COMMENT_MARKER)))) | .[].id')
>                                                                                                            ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.
> 
> For more information:
>   https://www.shellcheck.net/wiki/SC2016 -- Expressions don't expand in singl...
> ```
> 
> If you want an env var available in a JQ expression you should do.
> 
> ```
> jq --arg name "${ENV_VAR}" 'expression using $name' input.json
> ```
> 
> Shellcheck says to use double quotes, but we generally avoid double quotes around JQ expressions because of the frequency with which we use double quotes in them.

That's weird, it was working before 🤔

AP-Hunt:
Looking at `$ENV.COMMENT_MARKER` were you perhaps using GitHub Actions interpolation before? That would be `${{env.COMMENT_MARKER}}`. `$ENV.` isn't valid syntax for env vars in a POSIX shell afaik

lfdebrux:
> Looking at `$ENV.COMMENT_MARKER` were you perhaps using GitHub Actions interpolation before? That would be `${{env.COMMENT_MARKER}}`. `$ENV.` isn't valid syntax for env vars in a POSIX shell afaik


I was trying to use the special `$ENV` variable in jq (https://jqlang.org/manual/#$env-env), rather than do a parameter expansion. Is there any reason not to do that?

AP-Hunt:
None at all. I wasn't aware that existed!

lfdebrux:
@AP-Hunt fixed now (I was missing a newline)

AP-Hunt (approved):

PR #1806, opened 2025-02-25
.review_apps: use app autoscaling to turn review apps off at night, weekends

AP-Hunt:
> [!NOTE]
> This PR has been raised so that we can check the review app in the morning and see if it went down to zero.
>
> A matching PR has been raised to give the role permission to do this: https://github.com/alphagov/forms-deploy/pull/1391

Trello card: https://github.com/alphagov/forms-admin/pull/1806

AWS App Autoscaling allows us to scale AWS ECS services up and down on a schedule. To save money, we want to turn them off at 6pm each day, and then back on at 8am Monday to Friday.

The result should be them being turned off from 6pm to 8am Monday to Friday, and all day on weekends.

I let the scaling rules take effect overnight to demonstrate them working against the review app for this PR

image

image


sarahseewhy (approved):
Looks good 🪄 

I see no easter eggs 🥚

PR #1810, opened 2025-02-27
Fix question numbers in error messages for any other answer routes

lfdebrux:

Trello card: https://trello.com/c/xei3DTuv/2147-fix-question-numbers-in-error-messages-for-any-other-answer-routes 

For "normal" skip routes, the check page and the routing page for the condition are the same, but for the "any other answer" route in a branch the secondary skip condition has the check page as the page that starts the branch, which in the current model for branch routing is the page that is of interest when editing the routes for the branch.

This PR adds tests describing the behaviour we expect in terms of what the content should say, and changes the page list component to use the condition check page in all places to make those tests pass. It also does some small refactors.

It's probably a smell that we have to do this bookkeeping to model routes, and maybe we should have a record or class that models a route properly.

<details><summary><h3>Screenshots</h3></summary>

#### Before

Screenshot of the add and edit your questions page with errors in the any other answer route, before these changes

#### After

Screenshot of the add and edit your questions page with errors in the any other answer route, after these changes

</details>

### DIFF
--- a/app/components/page_list_component/error_summary/view.rb
+++ b/app/components/page_list_component/error_summary/view.rb
@@ -14,19 +14,23 @@ module PageListComponent
         OpenStruct.new(message: I18n.t("page_conditions.errors.#{error_name}", question_number: page.position), link: "##{self.class.error_id(condition.id)}")
       end
 
-      def conditions_with_routing_pages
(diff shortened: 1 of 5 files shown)


thomasiles (approved):
nice!

PR #1816, opened 2025-03-04
Change Mailchimp logging to avoid errors

thomasiles:
Mailchimp logging throws an error because the @response_body on the ApiError is not a public method.

This commit fixes the logging for Mailchimp errors.

When the mailchimp API returns an error, the exception contains information within the @response_body instance variable which isn't exposed.

The changes to log_mailchimp_error extract this information when it's valid and available and logs it as JSON.

It also takes care to filter out email addresses from the "detail" field in the response_body.

Trello card: 

### DIFF
--- a/lib/mailchimp_list_synchronizer.rb
+++ b/lib/mailchimp_list_synchronizer.rb
@@ -126,6 +126,34 @@ class MailchimpListSynchronizer
   end
 
   def log_mailchimp_error(action, subscriber_hash, error)
-    Rails.logger.warn "Could not #{action} user with subscriber hash #{subscriber_hash} from list #{list_id}. HTTP status: #{error.status}, response body: #{error.response_body}"
+    error_details = {}
+
+    response_body = error.instance_variable_get("@response_body")
+
+    begin
+      if response_body
+        parsed_body = JSON.parse(response_body)
+        error_details = parsed_body.is_a?(Hash) ? parsed_body : {}
+      end
+    rescue JSON::ParserError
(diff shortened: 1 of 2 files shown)


lib/mailchimp_list_synchronizer.rb
```diff
@@ -126,6 +126,34 @@ def remove_member(member)
   end
 
   def log_mailchimp_error(action, subscriber_hash, error)
-    Rails.logger.warn "Could not #{action} user with subscriber hash #{subscriber_hash} from list #{list_id}. HTTP status: #{error.status}, response body: #{error.response_body}"
+    error_details = {}
+
+    response_body = error.instance_variable_get("@response_body")
```
thomasiles, line 131: Getting the instance variable seems to be the only way to find out what went wrong.

The Mailchimp client only passes status and response body into the error, which get set as instance variables. Even though the error object has readers for other attributes, they are never set.

https://github.com/mailchimp/mailchimp-marketing-ruby/blob/dd4fd7adeb10cae5afbb0a4e8c332ae23345bb8b/lib/MailchimpMarketing/api_client.rb#L83

jamie-o-wilkinson: It's a shame we have to do this but you've clearly considered it and put tests around it, which mitigates the icky-ness!

jamie-o-wilkinson (approved):
Thanks for the explanation in your comment! With that in mind, this looks ok to me 👍

PR #1823, opened 2025-03-06
Add mailchimp role for MOU signers mailchimp list

thomasiles:
### Record role for the MOOU signers mailchimp list

Trello card: https://trello.com/c/jEXoX309/2125-change-mailchimp-audience-for-mou-updates-to-include-org-admins-as-well-as-mou-agreers

The updates in this commit will overwrite any existing merge_field values. We don't have any merge fields set but mailchimp adds some default values which are empty for almost all our users.

The role hasn't been added to the mailchimp list yet, I'll add it after this commit is approved.

### DIFF
--- a/app/services/mailchimp_list_sync_service.rb
+++ b/app/services/mailchimp_list_sync_service.rb
@@ -15,11 +15,23 @@ class MailchimpListSyncService
   end
 
   def mou_signers
-    MouSignature
-      .all
-      .map(&:user)
-      .filter { |user| user.has_access == true }
-      .pluck(:email)
-      .map { |email| MailchimpMember.new(email: email, status: "subscribed") }
+    User
+      .left_outer_joins(:mou_signatures)
+      .where(has_access: true)
+      .where("mou_signatures.id IS NOT NULL OR users.role = ?", "organisation_admin")
+      .distinct
+      .map { |user| MailchimpMember.new(email: user.email, status: "subscribed", role: mou_role(user)) }
+  end
+
+private
+
+  def mou_role(user)
+    if user.mou_signatures.present? && user.organisation_admin?
+      "Organisation admin agreed MOU"
+    elsif user.mou_signatures.present?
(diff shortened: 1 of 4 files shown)


lfdebrux:
So MailChimp members who have signed the MOU will have `ROLE` 'AGREED_MOU', while users who are organisation admins but haven't signed the MOU will have the role 'ORGANISATION_ADMIN'?

thomasiles:
> So MailChimp members who have signed the MOU will have `ROLE` 'Organisation admin agreed MOU', while users who are organisation admins but haven't signed the MOU will have the role 'Organisation admin'?

There should be three options based on two variables:
- have they signed the MOU
- are they an org admin

The three options:
- users who have signed the MOU but are not org admins should have "Agreed MOU"
- users who have signed the MOU and are org admins should have "Organisation admin agreed MOU"
- user who have not signed the MOU but are org_admins "Organisation admin"

The final case of not signed MOU and not org admin, shouldn't be on the list at all.

lfdebrux:
@thomasiles are we likely to want to use the role for other things in Mailchimp in the future?

thomasiles:
> @thomasiles are we likely to want to use the role for other things in Mailchimp in the future?

I'm not sure tbh. Adding role to the MailchimpMember class implies that all mailchimp lists might want to use it. I think that's a bit heavy handed but I wanted to keep things fairly simple as we only have two mailchimp lists.

What do you think?

lfdebrux:
> > @thomasiles are we likely to want to use the role for other things in Mailchimp in the future?
> 
> I'm not sure tbh. Adding role to the MailchimpMember class implies that all mailchimp lists might want to use it. I think that's a bit heavy handed but I wanted to keep things fairly simple as we only have two mailchimp lists.
> 
> What do you think?

I was mainly wondering if people would want to segment audiences in Mailchimp using the role for purposes other than the MOU signature emails. But it's probably fine, I'm just overthinking it!

PR #1826, opened 2025-03-07
2152 add feature flag for exit pages

SamJamCul:

https://trello.com/c/cp1oXNvH/2152-add-feature-flag-for-exit-pages

Adds a feature flag that will allow us to control the exit pages feature group-by-group. I've followed the instructions in this commit 16fb756a05fcfc7aebdeb4b38ad4c1139394d6d4

Also added a rake task which is run using `rake "groups:enable_exit_pages[<group_external_id>]"` or `disable_exit_pages` as needed. 

I've currently set it to be enabled by default for all groups in the dev environment, the same as our `branch_routing` feature flag

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -3,6 +3,8 @@ features:
   groups: true # Do not switch off!
   branch_routing:
     enabled_by_group: true
+  exit_pages:
+    enabled_by_group: true
 
 forms_api:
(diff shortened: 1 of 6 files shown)


thomasiles (approved):
looks good and works nicely for me locally 👍

PR #1853, opened 2025-03-24
Use chromedriver v128 in github actions

thomasiles:
We are seeing intermittent failures when feature tests run in github actions. These failures are related to selenium and chromedriver. They are false positives and do not mean the code is broken.

This commit copies the solution gov.uk applied to their codebase.

https://github.com/alphagov/publisher/pull/2596/commits/68adc8858782949a1a38f4eb582a17462eb0e51c

This change should only be needed until the chromedriver/chrome version issue is fixed.

Trello card: 

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,17 @@ jobs:
       RAILS_ENV: "test"
       DATABASE_URL: "postgres://postgres:postgres@localhost:5432/forms_admin_test"
     steps:
+      # TODO: remove these steps once we can use latest Chrome again (see https://github.com/teamcapybara/capybara/issues/2800)
+      - uses: nanasess/setup-chromedriver@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -32,6 +32,16 @@ jobs:
       RAILS_ENV: "test"
       DATABASE_URL: "postgres://postgres:postgres@localhost:5432/forms_admin_test"
     steps:
+      - uses: nanasess/setup-chromedriver@v2
```
lfdebrux, line 35: Let's add a comment explaining why we're doing this

```suggestion
      # TODO: remove these steps once we can use latest Chrome again (see https://github.com/teamcapybara/capybara/issues/2800)
      - uses: nanasess/setup-chromedriver@v2
```

lfdebrux (dismissed):

lfdebrux (approved):

PR #1858, opened 2025-03-25
Add task for moving users and groups from one organisation to another

lfdebrux:

Trello card: https://trello.com/c/Fq6YpoIA/2124-merging-cddo-users-forms-and-groups-into-gds 

It often happens that organisations are merged into each other. In the GOV.UK organisation API, this is modelled as an organisation being closed and superseded by another. When this happens however this doesn't affect existing users and groups in our system.

This commit adds a task to merge a closed `source_organisation` into a `target_organisation`, so that all users and groups end up in the target organisation.

We add the restriction that the source organisation should be closed before the task is run so that no users can be added to that organisation while the task is running.


thomasiles (dismissed):
Nice!

It's fiddly but is it worth adding tests for the MOU signature and overlapping name cases?

Happy for it to be merged so that it can be used though, looks really good 🏅

lfdebrux:
> It's fiddly but is it worth adding tests for the MOU signature and overlapping name cases?

Fair point, I've added those tests now

lib/tasks/organisations.rake
```diff
@@ -51,10 +51,62 @@ namespace :organisations do
     org.name = new_name
     puts "Renamed org #{old_name} to #{new_name}" if org.save!
   end
+
+  desc "Merge an organisation into another"
+  task :merge, %i[source_organisation_slug target_organisation_slug] => :environment do |_, args|
```
DavidBiddle, line 56: When running `organisations:merge`, the usage message doesn't include the task name - I think because the first argument should be `task`:

```suggestion
  task :merge, %i[source_organisation_slug target_organisation_slug] => :environment do |task, args|
```

lfdebrux: Ah, good spot, thanks! That should be fixed now.

DavidBiddle (approved):

PR #1864, opened 2025-03-28
Remove test forms from reports

DavidBiddle:

Trello card: https://trello.com/c/amETo48v/2120-live-forms-report-needs-amending

We want to exclude our test forms from our reports. This PR adds:
- a new `internal` column in the organisations table
- rake tasks for setting and unsetting the new column
- logic to exclude this org's forms from the FormDocumentsService response. That's used in the CSV and feature reports as of #1872 and @stephencdaly is in the process of updating the other reports to use it.

### DIFF
--- a/app/services/reports/form_documents_service.rb
+++ b/app/services/reports/form_documents_service.rb
@@ -13,7 +13,7 @@ class Reports::FormDocumentsService
         page = 1
         loop do
           form_documents_response = live_form_documents_page(page)
-          form_documents_response.forms.each { |f| yielder << f }
+          form_documents_response.forms.each { |f| yielder << f unless is_in_internal_organisation?(f) }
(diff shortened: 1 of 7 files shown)


config/settings.yml
```diff
@@ -96,3 +96,4 @@ act_as_user_enabled: false
 
 reports:
   forms_api_forms_per_request_page: 100
+  organisation_to_exclude_from_live_report: null
```
stephencdaly, line 99: I think we'll need to make this setting accept a list of IDs so that we can exclude both the "GOV.UK Forms team" and "GOV.UK Forms End to end tests" organisations

stephencdaly (approved):
Tested locally and everything works as expected. Nice work 🏅

PR #1865, opened 2025-03-28
Add server side double click prevention for creating a form

lfdebrux:

Trello card: https://trello.com/c/VUQH4gJS/2153-bug-sometimes-creating-a-form-creates-two-or-more-new-forms-with-the-same-name 

In production we've been seeing 1% to 2% of create a form journeys result in multiple forms with the same name being created. After spending time looking at logs and experimenting I've concluded that is likely because the browser is sometimes sending multiple requests if the button to submit the form is repeatedly activated.

We want to make sure that on the create a form page clicking the submit button more than once (or using the enter key more than once) doesn't result in multiple forms being created. Normally we have the `prevent-double-click` JavaScript from GOV.UK Frontend to do this. I haven't been able to find any bug in this, or recreate the issue of multiple requests being sent except by having JavaScript disabled.

So this PR I've adds a test which has JavaScript disabled and clicks the button to create a new form twice. This test fails when it is initially committed, we then commit a fix to turn it green.

The typical way to make a create request idempotent is to have a uniqueness constraint in the database. I experimented with locks of various kinds, but in the end I couldn't find something as reliable.

Because we don't keep the form records in the forms-admin database yet, this PR adds a new `create_form_events` table that we can enforce our constraint on. This can be thought of as a sort of event coalescing.

However, we don't actually want to restrict what forms can be named, so instead we add `dedup_version` column, which we can increment if there is a name collision that doesn't look to be due to a double click. This is partly inspired by the optimistic locking in ActiveRecord.

Note that to be able to have a browser test which does not use JavaScript, this PR adds Cuprite and Ferrum, drop-in replacements for Selenium and ChromeDriver that are pure Ruby.

### DIFF
--- a/app/services/create_form_service.rb
+++ b/app/services/create_form_service.rb
@@ -13,10 +13,16 @@ class CreateFormService
       # and return the form ID of the previously created form
       previous_event = CreateFormEvent.order(created_at: :desc).find_by(group:, form_name: name)
 
+      # We might need to wait for the form to be created by the previous request
       if previous_event.user == creator && previous_event.created_at > 1.second.ago
-        while previous_event.form_id.blank?
-          sleep(0.01)
-          previous_event.reload
+        timeout_message = "CreateFormService#create! timed out waiting for form to be created by previous invocation: " \
+          "did something go wrong with creating form '#{name}' in group #{group.id} by user #{creator.id}?"
+
+        Timeout.timeout(1, nil, timeout_message) do
+          while previous_event.form_id.blank?
+            sleep(0.01)
+            previous_event.reload
+          end
         end
 
         previous_event
--- a/app/services/create_form_service.rb
+++ b/app/services/create_form_service.rb
@@ -1,7 +1,37 @@
 class CreateFormService
+  class CreateFormEvent < ApplicationRecord
+    belongs_to :user
+    belongs_to :group
+  end
+
   def create!(creator:, group:, name:)
-    form = FormRepository.create!(creator_id: creator.id, name:)
-    GroupForm.create!(group:, form_id: form.id)
+    event = begin
+      CreateFormEvent.create!(group:, form_name: name, user: creator, dedup_version: 1)
+    rescue ActiveRecord::RecordNotUnique
+      # If a form with the same name in the same group was created by the same
+      # user less than a second ago, assume that this was a duplicate request
+      # and return the form ID of the previously created form
+      previous_event = CreateFormEvent.order(created_at: :desc).find_by(group:, form_name: name)
+
+      if previous_event.user == creator && previous_event.created_at > 1.second.ago
+        while previous_event.form_id.blank?
+          sleep(0.01)
+          previous_event.reload
+        end
+
+        previous_event
+      else
+        CreateFormEvent.create!(group:, form_name: name, user: creator, dedup_version: previous_event.dedup_version + 1)
+      end
+    end
+
+    if event.form_id.present?
+      form = FormRepository.find(form_id: event.form_id)
+    else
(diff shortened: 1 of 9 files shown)


lfdebrux:
Tests are failing, I wondering if Ferrum will work with our workaround for flaky Selenium tests in #1853 😢

lfdebrux:
> Tests are failing, I wondering if Ferrum will work with our workaround for flaky Selenium tests in #1853 😢

Ah, it turned out I just needed to read the https://github.com/rubycdp/ferrum/?tab=readme-ov-file#docker part of the Ferrum readme properly.

spec/services/create_form_service_spec.rb
```diff
@@ -25,5 +25,63 @@
 
       expect(GroupForm.last).to have_attributes(form_id: 1, group_id: 1000)
     end
+
+    # rubocop: disable RSpec/ContextWording
```
stephencdaly, line 29: Is there the need to disable this rule? I think "when" makes more sense than "if" for the outer context, and the other contexts could be worded to include a "when" and would probably be easier to read in the context of reading this file than without.

app/services/create_form_service.rb
```diff
@@ -1,7 +1,32 @@
...
+  end
+
   def create!(creator:, group:, name:)
-    form = FormRepository.create!(creator_id: creator.id, name:)
-    GroupForm.create!(group:, form_id: form.id)
+    event = begin
+      CreateFormEvent.create!(group:, form_name: name, user: creator, dedup_version: 1)
+    rescue ActiveRecord::RecordNotUnique
+      previous_event = CreateFormEvent.order(created_at: :desc).find_by(group:, form_name: name)
+      if previous_event.user == creator && previous_event.created_at > 1.second.ago
```
stephencdaly, line 12: Maybe it's because it's late in the day but my brain took slightly too long to figure out how this meant that the previous event was created more recently than 1 second ago, rather then more than one second ago. Reckon it's worth adding a tiny comment to help other people with slow brains?

app/services/create_form_service.rb
```diff
@@ -1,7 +1,32 @@
...
+        while previous_event.form_id.blank?
+          sleep(0.01)
```
stephencdaly, line 14: What if there is an error creating the form in forms-admin? I guess the request will just eventually time out? Will Rails kill the thread successfully?

stephencdaly (dismissed):
Nice solution. I've left some comments but none of them need to be actioned if you don't agree with them.

app/services/create_form_service.rb
lfdebrux: Yeah it's not just you, I also find the ordering in time comparisons annoying, I wish there was a more readable way of writing it. I'll add a comment, good suggestion.

spec/services/create_form_service_spec.rb
lfdebrux: Hmm, I struggled with this... an example would help... what would you suggest rewriting this to?

> if a form with that name was already created in that group at the same time

stephencdaly: "when a form with the same name was already created in the same group"
> "when both forms are created at the same time"
> "when the previous form was created less then one second ago"
> "when the previous form was created more than one second ago"
> "when the previous form was created by a different user"

It's a bit more verbose. But, personally I'd prefer if we either disabled this rule globally if we don't think it's helpful or followed it so we have consistency.

lfdebrux: Makes sense to me. I like the rule in general, I've followed your suggestion to make it more consistent in these specs.

stephencdaly (approved):

PR #1869, opened 2025-04-01
Add breadcrumbs for super admins on show group page

lfdebrux:

> As a super admin editing a group from another organisation
> I need to see which organisation the group is in,
> So that I can edit the group confidently.

This PR changes the back link on the page showing a group to be a breadcrumb to the page showing groups for the organisation the current group is in.

If the group is in the same organisation as the user the breadcrumb will take the user to the same page as the back link would; if the group is in a different organisation to the super admin user, the breadcrumb will take the user to the groups for that organisation as if they had searched for that organisation using the autocomplete.

This change should only affect super admins.

Screenshot 2025-04-01 at 11 07 33

### DIFF
--- a/app/views/groups/show.html.erb
+++ b/app/views/groups/show.html.erb
@@ -1,5 +1,13 @@
 <% set_page_title(@group.name) %>
-<% content_for :back_link, govuk_back_link_to(groups_path, t("back_link.groups")) %>
+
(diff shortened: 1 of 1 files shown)


thomasiles (approved):
I can't comment on the design or content but it works for me 👍

PR #1871, opened 2025-04-02
Remove groups feature flag

lfdebrux:

We are now fully on groups for user management, so we don't need the feature flag any more.

This PR removes the groups feature flag from the app and any code that depends on the flag being off.


DavidBiddle (approved):
Nice! Deletions all look good, commit messages make sense and I don't see any issues when clicking around groups on my local machine 🎉

PR #1872, opened 2025-04-02
Use live form data for feature report

stephencdaly:

Trello card: https://trello.com/c/SgDCKXcY/145-improve-our-feature-reports

Use the FormDocuments for live forms retrieved from forms-api to construct the features usage report. The FormDocument is the snapshot of the form made when a form is made live. Previously, we were using the data for the draft version of the form for live forms which can be different if there have been edits to the draft since the form was last made live.

This only updates the [Feature and answer type usage in live forms](https://admin.forms.service.gov.uk/reports/features) report for now. Other reports will be updated subsequently.


DavidBiddle (approved):
This approach makes sense, the new services keep it tidy and reusable. All looks good on my machine.

PR #1873, opened 2025-04-02
Use new add another answer report endpoint

stephencdaly:

Trello card: https://trello.com/c/SgDCKXcY/145-improve-our-feature-reports

Use the separate `/api/v1/reports/add-another-answer-forms` to render the forms with add another answer report. The old
`/api/v1/reports/features` endpoint that was returning this data is being removed.


DavidBiddle (approved):
Code looks good and the report still works locally 🎉

PR #1878, opened 2025-04-04
Refactor delete confirmation pages

lfdebrux:

While trying to add a new delete confirmation page (see https://github.com/alphagov/forms-admin/pull/1879), I found a few papercuts with the existing examples. This PR cleans them up to try and make it easier to add new delete confirmation pages in future.


jamie-o-wilkinson (approved):
Some nice refactoring. Form deletion still works as intended so LGTM 👏

PR #1883, opened 2025-04-07
Firebreak: branching form visualisation

AP-Hunt:

Allows form creators to visualise how their branching configuration will take form fillers through the questions

I think this is mostly complete from a feature perspective (it needs tests), but I could use some guidance on what other things I should have thought about.


AP-Hunt:
It was just a firebreak prototype

PR #1888, opened 2025-04-08
Remove warning from delete route confirmation page

hannahkc:

This removes the inset warning from the confirmation page for deleting a route 1.

We don't need this warning and it may not be factually correct (depending on the form and its questions) so we have agreed to remove it.

Trello card: https://trello.com/c/BEDjOcLE/2201-cheeky-ticket-remove-warning-from-the-delete-route-1-confirmation-page

- I  have edited a ruby/html file! please check I did this right! 

### DIFF
--- a/app/views/pages/conditions/delete.html.erb
+++ b/app/views/pages/conditions/delete.html.erb
@@ -33,10 +33,6 @@
 
       end %>
 
-      <%= govuk_inset_text do %>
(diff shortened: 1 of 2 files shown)


hannahkc:
oooof that review apps thing is well good. :)

lfdebrux (approved):
🙌

PR #1894, opened 2025-04-14
Change live view for branches

thomasiles:

This PR covers two trello cards because they are closely related and the issue with exit pages stops either being reviewed in the review apps.

They could be split into two very small PR's if preferred.

Fix the question view for live forms with exit pages and update the content to correctly show secondary skips.

Trello cards:
https://trello.com/c/Do0lsSF2/2789-change-the-page-that-shows-questions-in-a-live-form-to-work-with-exit-pages
https://trello.com/c/9Ff0wFPH/2204-change-the-page-that-shows-questions-in-a-live-form-to-work-with-branching-routes

image

## The design

[The design is in mural](https://app.mural.co/t/gaap0347/m/gaap0347/1728478914347/6a7e0b709f50f00f81cab37414fdf44ec8601a70) and linked from the trello cards.

image

### DIFF
--- a/app/resources/api/v1/condition_resource.rb
+++ b/app/resources/api/v1/condition_resource.rb
@@ -22,4 +22,8 @@ class Api::V1::ConditionResource < ActiveResource::Base
   def secondary_skip?
     answer_value.blank? && check_page_id != routing_page_id
   end
+
+  def exit_page?
+    attributes.include?("exit_page_markdown") && !attributes["exit_page_markdown"].nil?
+  end
 end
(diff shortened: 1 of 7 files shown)


app/resources/api/v1/condition_resource.rb
```diff
@@ -22,4 +22,8 @@ def errors_with_fields
   def secondary_skip?
     answer_value.blank? && check_page_id != routing_page_id
   end
+
+  def exit_page?
```
SamJamCul, line 26: I think I'm going to refactor a bit of the bigger #1902 pr using this

SamJamCul (approved):
Looks good!

PR #1913, opened 2025-04-22
Log answer type

stephencdaly:

Trello card: https://trello.com/c/R267eUxr/2178-log-answer-type-in-splunk-in-admin-and-runner

Add an answer_type attribute to log lines for pages where we are adding or editing questions. Get the answer_type from the DraftQuestion we are editing if there is one present, else from the page. Add this as an `after_action` in the controller so that we make sure we log the answer_type on requests that set it as part of the controller action.


lfdebrux (approved):
LGTM!

PR #1920, opened 2025-04-25
Fix server error when question with route is at the end of a form

lfdebrux:

Trello card: https://trello.com/c/T8ww3HhL/2208-bug-if-you-move-a-question-at-the-start-of-a-route-1-to 

We’re seeing errors in production because on the routes page we want to tell the user about the any other answer route, but in the case where the question with the route has been moved to the end of the form we can't find the question object for the next question number.

This commit fixes the error by not trying to find the next question if the route question is at the end of the form, and instead showing content telling the user the form filler will see the check your answers page next.

Curiously it looks like something like this was originally the plan, because there was already a spec for the show routes view for a question at the end of the form; I just fluffed the test when writing it 😅

### Screenshots

Screenshot of question's routes page when question is at the end of a form

### DIFF
--- a/app/views/pages/routes/show.html.erb
+++ b/app/views/pages/routes/show.html.erb
@@ -24,7 +24,11 @@
       <% if route_summary_card_data_presenter.routes.map(&:secondary_skip?).none? %>
         <h2 class="govuk-heading-m"><%= t(".any_other_answer.heading") %></h2>
 
-        <p class="govuk-body"><%= t(".any_other_answer.will_continue_to", next_question_number: route_summary_card_data_presenter.next_page&.position) %></p>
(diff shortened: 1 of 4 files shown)


hannahkc:
oh. We need to hide that second paragraph and the button to set questions to skip - because they can't skip any. Think we may have overlooked that this needs to be removed when the question is the last one in the form. Does this need to be a separate ticket?

lfdebrux:
@hannahkc good point! I've made that change now, and updated the screenshot.

hannahkc (approved):
Yay. Thanks so much for removing that line and button. I've also checked everything else in the review app and the journey and content all looks perfect. thank you!

PR #1922, opened 2025-04-28
Add service navigation

DavidBiddle:

Trello card: https://trello.com/c/oMuEl17Z/2205-implement-service-navigation-in-forms-admin

Adds service navigation to forms-admin.

Includes:
- copying across the component from forms-product-page
- making some tweaks for compatibility with `NavigationItemsService`
- moving the navigation links from the header to the new nav
- adding the new 'Your groups' item to the navigation


app/components/service_navigation_component/view.rb
```diff
@@ -0,0 +1,29 @@
+module ServiceNavigationComponent
+  class View < ViewComponent::Base
+    attr_accessor :navigation_items
+
+    def initialize(navigation_items: [], featured_link: nil)
```
stephencdaly, line 5: It doesn't look like we use the `featured_link` currently? It looks like we add the class to the Profile navigation item before we pass it in - and does the Sign in link just get pushed right because it comes after the Profile link?

DavidBiddle: Good point - copied it across from product page and I don't think we need it here. I'll remove the logic

Correct, the profile link pushes any subsequent items to the right

DavidBiddle: Done!

stephencdaly: Would there be any effect of adding `app-service-navigation__item--featured` to the Sign in item too? I'm thinking that if there isn't, maybe it's worthwhile adding to make it clear that we expect Sign in to be on the right. I'm thinking what would happen if we decided to move the profile link but forgot to update the Sign in link. If you don't think that's a worry though then fine as it is.

DavidBiddle: At the moment the way it's set up is that the featured item will stretch to fill as much unused space as possible and then the link/text inside it is right-aligned: 
image

If we added the class to both they'd share the unused space as it's currently set up:
image

I could revisit the styles but we're a bit constrained by the markup we get from the component, and I'm wary of customising it too much in case future bumps to govuk-components or frontend break things. So I'm inclined to leave it for now but might revisit if we make any more changes to the nav

stephencdaly: Cool, thanks for the explanation. It seems fine to me, I think we'd notice pretty quickly if we messed up the styling so shouldn't be a concern.

stephencdaly (approved):
Behaviour is as expected and the code all looks good

deberny:
Looks good to me 🎉  I wasn't able to review as superadmin in the review app, but I think it should be fine 😄

PR #1942, opened 2025-05-08
Add feature test for adding branching

SamJamCul:

Trello card: https://trello.com/c/HMEGyzxK/2211-add-feature-tests-for-branching-to-forms-admin

This adds a feature test which takes a form that's ready for routing, and adds branching routes to it. This involves first adding a route that goes from the first question to the last question - and then adding a secondary skip which goes from the second last question to the check your answers page. 

### DIFF
--- a/spec/factories/models/pages.rb
+++ b/spec/factories/models/pages.rb
@@ -48,6 +48,17 @@ FactoryBot.define do
       answer_settings { DataStruct.new(only_one_option:, selection_options:) }
     end
 
+    trait :with_api_selection_settings do
+      transient do
+        only_one_option { "true" }
+        selection_options { [DataStruct.new(attributes: { name: "Option 1" }), DataStruct.new(attributes: { name: "Option 2" })] }
+      end
+
+      question_text { Faker::Lorem.question }
+      answer_type { "selection" }
+      answer_settings { DataStruct.new(only_one_option:, selection_options:) }
+    end
+
     trait :with_text_settings do
       transient do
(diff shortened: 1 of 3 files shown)


spec/factories/models/pages.rb
```diff
@@ -48,6 +48,17 @@ def as_json(*args)
       answer_settings { DataStruct.new(only_one_option:, selection_options:) }
     end
 
+    trait :with_api_selection_settings do
```
lfdebrux, line 51: Rather than adding a new trait, can we instead change the existing `:with_selection_settings` trait to use a `DataStruct`?

SamJamCul: Annoyingly, after a lot of digging, it looks like this is a bit tricky to fix. We currently use the selection settings primarily in two places - when we're creating and editing selection type questions, and when we're doing routing stuff. 

For the create/edit selection type questions, we're actually doing this as a `DraftQuestion`, which means that we're converting the stuff we get back from the API [in this part](https://github.com/alphagov/forms-admin/blob/85db457d18ae2a6c060f66908b639654bdb6e0ba/app/models/draft_question.rb#L10-L15) of the `DraftQuestion` model. This in turn means the way we access information about the `answer_settings` is different depending on which part of the app we're in. 

I'm not completely sure how to resolve this without reworking the way we read answer settings for selection questions from the API.

lfdebrux: Arghh, that is annoying. Thanks for looking into it and explaining it so thoroughly. Let's go with what we've got here then.

PR #1949, opened 2025-05-12
Fix deleting group when forms have been created in it

lfdebrux:

We were getting foreign key constraint errors (see https://govuk-forms.sentry.io/issues/6562756176/) when trying to delete groups because the create_form_event table has associations with groups.

We currently only use the events for checking whether a form has already been created with the same name (to catch double clicks), so we can just delete the events.

This commit fixes the errors by adding a before_destroy callback to remove create form events before deleting the groups.

### DIFF
--- a/db/migrate/20250513062011_change_foreign_keys_create_form_events.rb
+++ b/db/migrate/20250513062011_change_foreign_keys_create_form_events.rb
@@ -1,0 +1,9 @@
+class ChangeForeignKeysCreateFormEvents < ActiveRecord::Migration[8.0]
+  def change
+    remove_foreign_key :create_form_events, :groups
+    add_foreign_key :create_form_events, :groups, on_delete: :cascade, validate: false
+
+    remove_foreign_key :create_form_events, :users
+    add_foreign_key :create_form_events, :users, on_delete: :cascade, validate: false
+  end
+end
--- a/db/migrate/20250513065144_validate_foreign_keys_create_form_events.rb
+++ b/db/migrate/20250513065144_validate_foreign_keys_create_form_events.rb
@@ -1,0 +1,6 @@
+class ValidateForeignKeysCreateFormEvents < ActiveRecord::Migration[8.0]
+  def change
+    validate_foreign_key :create_form_events, :groups
+    validate_foreign_key :create_form_events, :users
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_04_10_160700) do
+ActiveRecord::Schema[8.0].define(version: 2025_05_13_065144) do
   # These are extensions that must be enabled in order to support this database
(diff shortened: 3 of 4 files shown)


thomasiles:
I'm a bit out of my depth here so point me in the right direction if this is way off.

Could this be done with a change in the DB schema for the create_form_events table?

If the groups reference in the `create_form_event` table had `on_delete: :cascade`, whenever a group is deleted, the `create_form_event` is also deleted (I think?).

I think this might be better if it never makes sense to have a `create_form_event` without a group as it's enforced within the DB and cuts out the need for a lifecycle action?

It could be changed with a migration like this, I think

```
def up
    remove_foreign_key :create_form_events, :groups
    add_foreign_key :create_form_events, :groups, on_delete: :cascade
end
```

lfdebrux:
> I'm a bit out of my depth here so point me in the right direction if this is way off.
> 

I don't think you're out of your depth at all!

> Could this be done with a change in the DB schema for the create_form_events table?
> 
> If the groups reference in the `create_form_event` table had `on_delete: :cascade`, whenever a group is deleted, the `create_form_event` is also deleted (I think?).
> 
> I think this might be better if it never makes sense to have a `create_form_event` without a group as it's enforced within the DB and cuts out the need for a lifecycle action?
>

This is a great idea, much nicer than what I did. I was looking at the `dependant` options for `belongs_to` associations but that didn't work the way round I wanted it to, I didn't think to look at the database level. 

> 
> It could be changed with a migration like this, I think
> 
> ```
> def up
>     remove_foreign_key :create_form_events, :groups
>     add_foreign_key :create_form_events, :groups, on_delete: :cascade
> end
> ```

I think this migration would work, but I think it would also lock out the `create_form_events` table while it ran... at least I think that's what https://github.com/ankane/strong_migrations?tab=readme-ov-file#adding-a-foreign-key is telling me.

thomasiles:
> > I'm a bit out of my depth here so point me in the right direction if this is way off.
> 
> I don't think you're out of your depth at all!
> 
> > Could this be done with a change in the DB schema for the create_form_events table?
> > If the groups reference in the `create_form_event` table had `on_delete: :cascade`, whenever a group is deleted, the `create_form_event` is also deleted (I think?).
> > I think this might be better if it never makes sense to have a `create_form_event` without a group as it's enforced within the DB and cuts out the need for a lifecycle action?
> 
> This is a great idea, much nicer than what I did. I was looking at the `dependant` options for `belongs_to` associations but that didn't work the way round I wanted it to, I didn't think to look at the database level.
> 
> > It could be changed with a migration like this, I think
> > ```
> > def up
> >     remove_foreign_key :create_form_events, :groups
> >     add_foreign_key :create_form_events, :groups, on_delete: :cascade
> > end
> > ```
> 
> I think this migration would work, but I think it would also lock out the `create_form_events` table while it ran... at least I think that's what https://github.com/ankane/strong_migrations?tab=readme-ov-file#adding-a-foreign-key is telling me.

ah, thats good thinking. That page is really useful!

thomasiles (approved):
looks good and works locally

PR #1969, opened 2025-05-21
Remove file upload feature flags

jamie-o-wilkinson:

This PR removes the global and per group file upload feature flags.

Trello card: https://trello.com/c/CLmKzjAX

### DIFF
--- a/app/controllers/pages/type_of_answer_controller.rb
+++ b/app/controllers/pages/type_of_answer_controller.rb
@@ -93,6 +93,6 @@ private
   end
 
   def answer_types
-    Page::ANSWER_TYPES_INCLUDING_FILE
+    Page::ANSWER_TYPES
   end
 end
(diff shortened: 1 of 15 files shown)


app/controllers/pages/type_of_answer_controller.rb
```diff
@@ -93,6 +93,6 @@ def set_answer_types
   end
 
   def answer_types
-    Page::ANSWER_TYPES_INCLUDING_FILE
+    Page::ANSWER_TYPES
```
stephencdaly, line 96: Do you reckon it's worth undoing commit [19b5a21faf350ab3208683aaee52bfcd4b81d499](https://github.com/alphagov/forms-admin/commit/19b5a21faf350ab3208683aaee52bfcd4b81d499) and reference the constant from the various places rather than having this method?

I guess the advantage of keeping this method is its easier if we need to have new question types behind a feature flag?

stephencdaly (approved):
Looks good to me. Just a small question

PR #1970, opened 2025-05-21
Refactor feature report to make it easier to add new feature reports (part 2)

lfdebrux:

I wanted to add a new feature report, but I found the feature report code was very repetitive and I had to add a lot of files to do that, and I felt like this situation could be improved. So in this PR, I try to make the feature report code more DRY.

The aim was to get it to the point where (in the simplest case) adding a new feature report required only adding a new method to `Reports::FeatureReportService`, and all the related routes, controller actions, views should automatically be generated, including for the CSV downloads.

A big part of this is making the responsibilities for each class involved in making a feature report more clearly deliniated and less overlapping, with the controller creating the pipeline to get the final result. It is now expected that FormDocumentsService will be called first, and that passed to FeatureReportService with is responsible just for filtering form documents and optionally converting it into a list of questions. There are helpers that turn this array of form documents or question page documents into a GOV.UK styles table; if a CSV is wanted for the end result, CsvReportService generates the CSV for either an array of form documents or an array of question page documents.

There is a little bit of metaprogramming magic required for this, to make the FeatureReportService aware of what methods the other things should be generated for there is a new class method `.define_report`:

```
class Reports::FeatureReportService
  define_report forms_for_juggling_licences
    form_documents
      .select { |form| form["content"]["name"] == "Apply for a juggling licence }
  end
end
```

For a more realistic example, see #1971.

The changes in this PR also increase the reusability to the point where it's feasible for us to add feature reports for draft forms, see #1972 for this in action.

### DIFF
--- a/app/services/reports/csv_report_service.rb
+++ b/app/services/reports/csv_report_service.rb
@@ -1,0 +1,38 @@
+class Reports::CsvReportService
+  delegate :csv, to: :@csv_service
+
+  def initialize(records)
+    record_type = detect_record_type(records)
+    @csv_service = csv_service_class(record_type).new(records)
+  end
+
+private
+
+  class NilCsvService
+    def initialize(_records); end
+
+    def csv
+      ""
+    end
+  end
+
+  def csv_service_class(record_type)
+    return NilCsvService if record_type.nil?
+
+    "Reports::#{record_type.capitalize}CsvReportService".constantize
+  end
+
+  def detect_record_type(records)
+    return if records.blank?
+
+    type = records.first.fetch("type", "form")
+
+    if type == "form"
+      :forms
+    elsif type == "question_page"
+      :questions
+    else
+      raise "type of records '#{type}' is not one of 'forms', 'question_page'" unless type
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


app/services/reports/csv_report_service.rb
```diff
@@ -0,0 +1,38 @@
...
+    def csv
+      ""
+    end
+  end
+
+  def csv_service_class(record_type)
+    return NilCsvService if record_type.nil?
+
+    "Reports::#{record_type.capitalize}CsvReportService".constantize
+  end
```
stephencdaly, line 23: I fear this is might make it more difficult to understand than it makes it easier to maintain.

stephencdaly (commented):
I was following this quite well until the point CsvReportService was added, and then it started getting a bit confusing.

Up to this point, I think all the changes made sense and there were some really nice improvements in here.

Could you potentially make the PR stop at that point and then raise another PR that combines all the routes into one so we can look at that in isolation? I'd fear that although you've made it potentially simpler to add a report, it might be a bit confusing to understand the magic behind it.

PR #1973, opened 2025-05-21
2245 add edit an exit page page

SamJamCul:

Trello card: https://trello.com/c/TpkOCn5o/2245-add-edit-an-exit-page-page

Adds the ability to add and edit an exit page when modifying an existing route. 

To test, create a route-qualifying form, and add a route. You can opt to also add an exit page as part of this. Once you've done this, edit the route. You'll see the exit page options available in the goto dropdown. 

image

When adding an exit page here, you'll be taken to the `Edit exit page` page. 

image

Once you save your heading and content, you'll go back to the edit route page, with the new exit page as the goto option - and a link to edit the exit page. 

image

If your page has a secondary skip route already, then you won't see the exit page option. 

I've added a very rudimentary deletion bit, so that if you choose to route to a question instead of an existing exit page, it'll delete the exit page. This is to prevent it breaking and makes it easier to test with. There's a followup card to properly implement deleting. 

### DIFF
--- a/app/controllers/pages/conditions_controller.rb
+++ b/app/controllers/pages/conditions_controller.rb
@@ -56,7 +56,11 @@ class Pages::ConditionsController < PagesController
     condition_input = Pages::ConditionsInput.new(form_params)
 
     if condition_input.update_condition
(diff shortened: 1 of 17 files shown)


thomasiles (approved):
This looks good, is nicely structured and works for me locally 🏅

thomasiles (approved):
looks good 👍

PR #1976, opened 2025-05-22
Refactor feature reports  to make it easier to add new feature reports

lfdebrux:

I wanted to add a new feature report, but I found the feature report code was very repetitive and I had to add a lot of files to do that, and I felt like this situation could be improved. So in this PR, I try to make the feature report code more DRY.

A big part of this is making the responsibilities for each class involved in making a feature report more clearly deliniated and less overlapping, with the controller creating the pipeline to get the final result. It is now expected that FormDocumentsService will be called first, and that passed to FeatureReportService with is responsible just for filtering form documents and optionally converting it into a list of questions. There are helpers that turn this array of form documents or question page documents into a GOV.UK styles table; if a CSV is wanted for the end result, `FormsCsvReportService` generates the CSV for  an array of form documents and `QuestionsCsvReportService` generates the CSV for an array of question page documents.

This could be taken further still, to the point where adding a new report requires only adding a new method to `FeatureReportService`; this PR was split out from #1970.


stephencdaly (commented):
Really nice improvements that are logically broken down. 

I've compared the output of the reports and the CSV downloads with some local data and they look the same after these changes from what I can see.

Just the minor issue around the table formatting when an organisation isn't set for a form that we talked about on slack, but otherwise it's good to go.

lfdebrux:
> Just the minor issue around the table formatting when an organisation isn't set for a form that we talked about on slack, but otherwise it's good to go.

Fixed now in [Fix organisation name when form is not in group](https://github.com/alphagov/forms-admin/pull/1976/commits/d9c02619c06e5ecca09dee2de5377e1596851a50).

stephencdaly (approved):

PR #1978, opened 2025-05-23
Add volume hints to the Dockerfile

AP-Hunt:

Trello card: https://trello.com/c/KoHvaEUA/681-aws-m112-ecs-read-only-root-filesystem-configuration

We can use VOLUME instructions in our Dockerfile to instruct the runtime to create a mount point for an external volume.

When AWS ECS sees the instruction (from the built container), it automatically attaches ephemeral storage at that location [1]. This is useful for us because we want to enable read only root filesystems, but the applications still need to be able to write to a small portion of it. We can achieve that by mounting ephemeral storage atop the paths it needs to write to.

[1] https://docs.aws.amazon.com/AmazonECS/latest/developerguide/bind-mounts.html#bind-mount-considerations

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,11 @@ RUN npm ci --ignore-scripts --only=production
 
 FROM base AS app
 
+# Each directory that Rails or our application needs to
+# write to under /app/tmp/ must be added individually
+VOLUME "/tmp/"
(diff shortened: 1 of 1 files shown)


sarahseewhy (approved):
Thank you 🚀

PR #1994, opened 2025-05-30
Add markdown editor to edit an exit page

SamJamCul:

Trello card: https://trello.com/c/nImD1lje/2246-add-markdown-editor-to-edit-an-exit-page

Introduces the markdown editor component to the exit page add and edit bits. Exit page content will be input using the markdown editor. This involved adding a preview endpoint for exit page stuff.

This also adds a character length limit validation to exit page heading and content fields of 5000 chars, as this was missed in the initial input object's creation. 

Try adding a new exit page, or editing an existing one. The markdown editor should be present, and will validate your content accordingly. 

image

image

image

### DIFF
--- a/app/input_objects/pages/update_exit_page_input.rb
+++ b/app/input_objects/pages/update_exit_page_input.rb
@@ -2,6 +2,7 @@ class Pages::UpdateExitPageInput < BaseInput
   attr_accessor :form, :page, :record, :exit_page_markdown, :exit_page_heading
 
   validates :exit_page_heading, :exit_page_markdown, presence: true
+  validates :exit_page_heading, :exit_page_markdown, length: { maximum: 4999 }
 
   def submit
     return false if invalid?
--- a/config/locales/pages/conditions.yml
+++ b/config/locales/pages/conditions.yml
@@ -23,8 +23,10 @@ en:
               blank: Select the answer to base this route on
             exit_page_heading:
               blank: Enter a page heading
+              too_long: The page heading must be shorter than 5,000 characters
             exit_page_markdown:
               blank: Enter page content
+              too_long: The page content must be shorter than 5,000 characters
         pages/routing_page_input:
           attributes:
             routing_page_id:
@@ -34,8 +36,10 @@ en:
           attributes:
             exit_page_heading:
               blank: Enter a page heading
+              too_long: The page heading must be shorter than 5,000 characters
             exit_page_markdown:
               blank: Enter page content
+              too_long: The page content must be shorter than 5,000 characters
   helpers:
     hint:
       pages_exit_page_input:
--- a/app/controllers/pages/exit_page_controller.rb
+++ b/app/controllers/pages/exit_page_controller.rb
@@ -70,6 +70,14 @@ class Pages::ExitPageController < PagesController
     redirect_to new_condition_path(@current_form.id, @page.id), success: t("banner.success.exit_page_deleted")
   end
 
+  def render_preview
+    authorize current_form, :can_view_form?
+    exit_page_input = Pages::ExitPageInput.new(exit_page_markdown: params[:markdown])
+    exit_page_input.validate if params[:check_preview_validation] == "true"
+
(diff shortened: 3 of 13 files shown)


config/locales/pages/conditions.yml
```diff
@@ -23,8 +23,10 @@ en:
               blank: Select the answer to base this route on
             exit_page_heading:
               blank: Enter a page heading
+              too_long: Heading text must be 5,000 characters or less
```
hannahkc, line 26:
```suggestion
              too_long: Page heading must be 5,000 characters or less
```

config/locales/pages/conditions.yml
```diff
@@ -23,8 +23,10 @@ en:
...
             exit_page_markdown:
               blank: Enter page content
+              too_long: Content text must be 5,000 characters or less
```
hannahkc, line 29:
```suggestion
              too_long: Page content must be 5,000 characters or less
```

config/locales/pages/conditions.yml
```diff
@@ -34,8 +36,10 @@ en:
           attributes:
             exit_page_heading:
               blank: Enter a page heading
+              too_long: Heading text must be 5,000 characters or less
```
hannahkc, line 39:
```suggestion
              too_long: Page heading must be 5,000 characters or less
```

config/locales/pages/conditions.yml
```diff
@@ -34,8 +36,10 @@ en:
...
             exit_page_markdown:
               blank: Enter page content
+              too_long: Content text must be 5,000 characters or less
```
hannahkc, line 42:
```suggestion
              too_long: Page content must be 5,000 characters or less
```

hannahkc (commented):
4 very minor content tweaks - to reflect the specific name of the relevant field in the error message.

config/locales/pages/conditions.yml
hannahkc: That's the opposite of ignorant because, strictly speaking, you are correct. However it's a style thing to use 'or less' in this case. Even though people might argue it's not grammatically correct, it IS how most people would say it. This is based on the error message examples in the Design System: https://design-system.service.gov.uk/components/text-input/#if-the-input-is-too-long 

That said, my preference - and what we use in most other places - would be to say "The page heading must be shorter than 5,000 characters" because then that doesn't look weird to fans of grammar rules. But "shorter than 5,001" is a bit odd so I went with this... I guess we could change the maximums to 4,999? That's the pattern we tend to follow elsewhere I think - though we're not always consistent!

SamJamCul: Ah we can definitely do that here, change the limit to 4999 and the message to the `shorter than 5000` version. Would that be preferable?

hannahkc:
Yes that would be best I think. Hope that's ok! And thank you!

app/input_objects/pages/update_exit_page_input.rb
```diff
@@ -2,6 +2,7 @@ class Pages::UpdateExitPageInput < BaseInput
   attr_accessor :form, :page, :record, :exit_page_markdown, :exit_page_heading
 
   validates :exit_page_heading, :exit_page_markdown, presence: true
+  validates :exit_page_heading, :exit_page_markdown, length: { maximum: 5000 }
```
thomasiles, line 5: We are adding a new validation to an existing field, do we need to think about the forms people have already created?

Exit pages is behind a flag so I don't think it will matter? The worst thing which will happen is that someone won't be able to edit their exit page if they have already made a really long one.

I guess we also don't need to worry that someone has put "unsupported" markdown features in the current text box.

thomasiles (approved):
Looks good and works locally for me from a technical point of view.

config/locales/pages/conditions.yml
hannahkc: Yes that would be best I think. Hope that's ok! And thank you!

thomasiles (approved):

PR #1995, opened 2025-05-30
2260 changing a route that goes to an exit page to go to a question should delete the exit page

thomasiles:
###  Changing a route that goes to an exit page to go to a question deletes the exit page

Trello card: https://trello.com/c/PioTk5zP/2259-add-page-for-deleting-exit-pages

This PR adds an extra screen when a route is changed from being an exit page to being a normal route.

When this happens, the users exit page content will be deleted. The new screen asks the user to confirm the change before their data is deleted.

image


SamJamCul (dismissed):
Works for me!

SamJamCul (approved):

PR #2004, opened 2025-06-04
Disable validation for markdown preview when editing a new exit page

SamJamCul:
This change means that the markdown preview component will not do live validation when you first go to edit an exit page. This is to ensure that validation does not trigger prematurely when there is no exit page content, and instead will only start triggering once the exit page data is submitted. From that point on, there will be live preview validation.

https://trello.com/c/bg9aHVVv/2284-launch-exit-pages

Fixes an issue where validation would trigger before any content had been submitted for an exit page. This only mattered when creating a new exit page from an existing route - as there wouldn't be any page content yet, but the markdown preview validates dynamically. 

To test, try changing an existing route to an exit page route - you shouldn't get prematurely validated on your markdown, and will only get validation errors after you try to submit. The two validations are for presence and length being < 5000

### DIFF
--- a/app/controllers/pages/exit_page_controller.rb
+++ b/app/controllers/pages/exit_page_controller.rb
@@ -25,7 +25,7 @@ class Pages::ExitPageController < PagesController
 
     update_exit_page_input = Pages::UpdateExitPageInput.new(form: current_form, page:, record: condition).assign_condition_values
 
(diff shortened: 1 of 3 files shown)


thomasiles (approved):
looks good to me 👍

PR #2028, opened 2025-06-19
Add validation errors to Google analytics

thomasiles:

Trello card: https://trello.com/c/b6jmnphx/2288-track-branching-validation-errors

This PR adds a way to pass analytics events from backend code to the frontend. It then uses this to add validation error tracking to all `input objects` and the routing and condition errors found on the pagelist.

This is marked as WIP because I think we need Anne to review the event names and properties before merging.

I think we can still review the technical approach though.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -7,6 +7,7 @@ class PagesController < ApplicationController
   def index
     @pages = FormRepository.pages(current_form)
     @mark_complete_input = Forms::MarkPagesSectionCompleteInput.new(form: current_form).assign_form_values
+    log_validation_errors(@pages)
     render :index, locals: { current_form: }
   end
 
@@ -137,6 +138,21 @@ private
       CurrentLoggingAttributes.answer_type = @draft_question.answer_type
     elsif @page.present?
       CurrentLoggingAttributes.answer_type = @page.answer_type
+    end
+  end
+
+  def log_validation_errors(pages)
+    # these validation errors don't come from an input object, so we log them ourselves
+    errors = pages.flat_map(&:routing_conditions).flat_map(&:validation_errors)
+    CurrentLoggingAttributes.validation_errors = errors.map { |error| "PageList: #{error.name}" } if errors.any?
+
+    pages.each do |page|
+      page.routing_conditions.each do |condition|
+        condition.validation_errors.each do |error|
+          error_type = condition.secondary_skip? ? "any_other_answer_route.#{error.name}" : error.name
+          AnalyticsService.track_validation_errors(input_object_name: "PageList", field: :condition, error_type:)
+        end
+      end
     end
   end
 end
--- a/app/services/analytics_service.rb
+++ b/app/services/analytics_service.rb
@@ -1,0 +1,29 @@
+class AnalyticsService
+  class << self
+    def track_validation_errors(input_object_name:, field:, error_type:)
+      add_event("validation_error", {
+        event_name: "validation_error",
+        form_name: input_object_name,
+        error_field: field,
+        error_type: error_type,
+      })
+    end
+
+    def add_events_from_flash(flash)
+      flash.each do |event|
+        add_event(
+          event["name"] || event[:name],
+          event["properties"] || event[:properties],
+        )
+      end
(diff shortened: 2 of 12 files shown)


spec/models/current_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  describe "ActiveSupport::CurrentAttributes behavior" do
+    it "isolates attributes between threads" do
+      # This test demonstrates the core behavior of CurrentAttributes
+      described_class.add_analytics_event("main_thread")
+
+      thread_events = nil
+      Thread.new {
+        described_class.add_analytics_event("background_thread")
+        thread_events = described_class.analytics_events
+      }.join
```
lfdebrux, line 84: An alternative way to write those would be to use [Thread#value](https://ruby-doc.org/3.4.1/Thread.html#method-i-value):

```suggestion
      thread = Thread.new {
        described_class.add_analytics_event("background_thread")
        described_class.analytics_events
      }
      thread_events = thread.value
      
```

app/controllers/pages_controller.rb
```diff
@@ -139,4 +140,19 @@ def set_answer_type_logging_attribute
       CurrentLoggingAttributes.answer_type = @page.answer_type
     end
   end
+
+  def log_validation_errors(pages)
```
lfdebrux, line 144: 🙌

spec/models/current_spec.rb
thomasiles: good idea, I didn't know that, I've made the change.

app/services/analytics_service.rb
```diff
@@ -0,0 +1,29 @@
+class AnalyticsService
+  class << self
+    def track_validation_errors(input_object_name:, field:, error_type:)
+      add_event("validation_error", {
+        event_name: "validation_error",
+        form_name: input_object_name,
+        error_field: field,
+        error_type: error_type,
```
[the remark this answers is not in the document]
lfdebrux: Ooh maybe the request/trace ID would be helpful?

SamJamCul (approved):

app/views/layouts/base.html.erb
```diff
@@ -60,6 +60,19 @@
...
+    <% if AnalyticsService.analytics_events.present? && @current_user&.collect_analytics? %>
+      <script data-analytics-events>
+        window.dataLayer = window.dataLayer || [];
+
+        <% AnalyticsService.analytics_events.each do |event| %>
+            window.dataLayer.push({
+              event: "event_data",
+              event_data: <%= raw event[:properties].to_json %>,
+            });
+      <% end %>
```
lfdebrux, line 72: I think the indentation on this line is slightly wrong?

```suggestion
        <% end %>
```

app/views/layouts/base.html.erb
```diff
@@ -60,6 +60,19 @@
...
+      </script>
```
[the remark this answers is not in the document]
lfdebrux: I think what you've got works for now, but leaves it open to do what you suggest in future, so let's leave as is?

app/views/layouts/base.html.erb
```diff
@@ -60,6 +60,19 @@
     <%= vite_client_tag %>
     <%= vite_javascript_tag 'application' %>
 
+    <% if AnalyticsService.analytics_events.present? && @current_user&.collect_analytics? %>
+      <script data-analytics-events>
+        window.dataLayer = window.dataLayer || [];
+
+        <% AnalyticsService.analytics_events.each do |event| %>
+            window.dataLayer.push({
+              event: "event_data",
```
lfdebrux, line 69: Why is the event called `event_data`?

lfdebrux: And we use `event_data` for all our custom events? If we have multiple events called `event_data` does the event data get coalesced by GA into one event?

thomasiles: I don't think the events get coalesced at any point and I think they can always be distinguished by name.

I'm not that up on how these are documented or at where @AnneCarr might be able to help. Originally there was talk of a single data environment shared across *.gov.uk properties. I don't know any specifics though.

PR #2030, opened 2025-06-20
2336 remove feature flag for exit pages

thomasiles:

Trello card: https://trello.com/c/y7g5N9Yn/2336-remove-feature-flag-for-exit-pages

Remove the exit_pages feature flag now the feature has been released.

This PR ignores the column on groups for the feature flag.

A later PR will actually drop the column.

When reviewing, we should double check this works the way we expect it to. The PR to remove the branch_routing feature flag caused an incident when deployed.


lfdebrux (approved):
Looks good to me!

Thanks for the nice, clearly explained individual commits, they made this really easy to review 😁

PR #2042, opened 2025-06-27
Add notification banner when deleting route with exit page

SamJamCul:
Adds a notification banner to the delete-route page, when there's an exit page for that route.

This commit also uses the exit page heading in place of the goto page question text in the route summary, which would otherwise be blank (as exit page routes have no goto page)

Trello card: https://trello.com/c/9M6hc0mI/2348-add-a-banner-notification-warning-to-the-confirmation-page-when-deleting-a-route-that-goes-to-an-exit-page

**Before**
___
image
___

**After**
___
image
___


thomasiles (approved):
Looks good 👍

PR #2048, opened 2025-07-02
Add ActiveRecord models for forms, pages, and conditions

lfdebrux:

Trello card: https://trello.com/c/CMLGnkYl/2333-change-repositories-in-forms-admin-to-replicate-to-database 

We want to start saving models related to forms to the forms-admin database, using ActiveRecord. To do this we need ActiveRecord classes for each model; this PR adds the bare minimum to allow the replication to work.

See commits for details. See https://github.com/alphagov/forms-admin/pull/2049 for an example of this code being used.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,6 +81,9 @@ gem "MailchimpMarketing", "~> 3.0"
 # For generating CSV reports
 gem "csv"
 
+# Used for sorting/ordering of pages object
+gem "acts_as_list"
+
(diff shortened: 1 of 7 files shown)


thomasiles (approved):
This looks good to me and it's really nicely structured and easy to read 🏅

DavidBiddle (approved):
Nice work 🙌 all looks sensible, well documented, and matches what's in API

PR #2049, opened 2025-07-02
Add replication to database for forms records

lfdebrux:

Trello card: https://trello.com/c/CMLGnkYl/2333-change-repositories-in-forms-admin-to-replicate-to-database 

We want to start migrating records related to editing forms from the forms-api database to the forms-admin database.

This PR adds an extra layer to each of the repository clasess for accessing forms; they will now persist the data to the database before passing it along to the caller.

Specifically, this PR makes sure that:
- CRUD operations on forms, pages and conditions save the data for that object to the databse
- CRUD operations for a page save the data for the routing conditions for that page to the database
- CRUD operations for a page delete any routing conditions that are associated with the page in the database but not in the API response
- Reading the pages for a form saves the data for those pages to the database
- Reading the pages for a form saves the data for those pages' routing conditions to the database
- Reading the pages for a form deletes any pages that are associated with the form in the database but not in the API response
- Reading the pages for a form deletes any conditions that are (transitively) associated with the form in the database but not in the API response

I _think_ this is everything needed to keep the databases in sync, but I want to do some experiments where I run the end to end tests and compare the database state before and after before I mark this PR as ready to review.

### DIFF
--- a/app/resources/api/v1/form_resource.rb
+++ b/app/resources/api/v1/form_resource.rb
@@ -17,6 +17,7 @@ class Api::V1::FormResource < ActiveResource::Base
   def database_attributes
     attributes
       .slice(*Form.attribute_names)
+      .with_defaults(external_id: id.to_s)
   end
 
   def group
--- a/app/services/form_repository.rb
+++ b/app/services/form_repository.rb
@@ -2,7 +2,7 @@ class FormRepository
   class << self
     def create!(creator_id:, name:)
       form = Api::V1::FormResource.create!(creator_id:, name:)
-      Form.create!(form.database_attributes)
+      save_to_database!(form)
       form
     end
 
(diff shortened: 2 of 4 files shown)


SamJamCul (dismissed):
Left a bigger comment above, but in summary I think we're good to go!

lfdebrux:
Thanks for the review, I want to figure out why the created at/updated at isn't being copied across exactly, I hadn't spotted that before. I'll have a little more playing around.

lfdebrux:
Ah, I've spotted another issue, a form that is read or updated without being created in the forms-admin database first won't have the external ID set.

lfdebrux:
Turns out the timestamps were like that because that's all the digits forms-api was sending; there's a fix for that in https://github.com/alphagov/forms-api/pull/790

lfdebrux:
I've pushed commit https://github.com/alphagov/forms-admin/pull/2049/commits/45176d458d204b738c32cafd399a1823236783fa to fix the form external ID

SamJamCul (approved):

PR #2050, opened 2025-07-02
Order users list in the database rather than in memory

DavidBiddle:

Trello card: https://trello.com/c/UXjZG0ZG/2946-improve-load-speed-on-users-page

This is an attempt to fix the performance issue we see on the Users page doing the ordering in the database query, rather than ordering the items in memory in the app after they've been fetched from the database.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -9,15 +9,7 @@ class UsersController < WebController
   def index
     authorize current_user, :can_manage_user?
 
-    roles = User.roles.keys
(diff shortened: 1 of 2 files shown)


DavidBiddle:
I haven't figured out how to test this yet, since the issue is only noticeable in prod

theseanything:
Think this is a good idea regardless, our current implementation is causes a database query per user (i.e. 2000+ query's to load the users page) due to the `user.organisation&.name`, which means it needs to fetch each organisation individually.

stephencdaly:
This was included in https://github.com/alphagov/forms-admin/pull/2229

PR #2054, opened 2025-07-04
Omit secondary skip questions from route options

stephencdaly:

Trello card: https://trello.com/c/rWs8nitM

On the page to add a new route for a form, do not include questions which already have a secondary skip condition associated with them in the options for adding a new route.

Previously, it was possible to select questions with a secondary skip, and then the user was shown a page to add a new secondary skip. When the "Set questions to skip" button on this page was clicked, they were taken back to the questions list page with no error message.


lfdebrux (approved):
Looks good to me, thanks for fixing the broken test

PR #2062, opened 2025-07-07
Add validation to privacy URL to prevent example

thomasiles:
# Prevent form fillers using the example privacy policy URL 

Trello card: https://trello.com/c/ho1fH3dm/2356-add-validation-and-error-when-people-use-the-example-privacy-notice-link

When form creators are asked to enter a privacy policy URL, they are shown an example of a privacy policy URL. This URL is not suitable for forms to use, so we should prevent them from entering it.

This commit adds a validation to to prevent the example URL from being used.

It only flags the example given. Determined form creators could change the URL to http, add `/` to the end or make other changes if they really want to use this URL.

We only prevent the example case because I think copying and pasting is the issue, but I'm happy to look at URL normalisation or more elaborate ways to check the URL.

image

## Existing invalid values

This PR only change the validation for new values. If a form already has an invalid value, it won't change. A form filler who has saved this privacy policy will be able to view it from the link in task list. It will show without an error unless they attempt to save it again, when the validation message will be shown.

 


config/locales/input_objects/privacy_policy.yml
```diff
@@ -7,6 +7,7 @@ en:
           attributes:
             privacy_policy_url:
               blank: Enter a link to privacy information
+              exclusion: https://www.gov.uk/help/privacy-notice is the privacy notice for the GOV.UK website and cannot be used for your form.
```
hannahkc, line 10:
```suggestion
              exclusion: https://www.gov.uk/help/privacy-notice is the privacy notice for the GOV.UK website and cannot be used for your form
```

DavidBiddle (approved):
Code looks good and works in the review app 🎉

PR #2063, opened 2025-07-08
Show correct message for groups with no org admin

DavidBiddle:

Trello card: https://trello.com/c/w5aByeFc/2358-bug-wrong-banner-being-shown-to-group-admins-of-a-trial-group

- Adds a missing logic branch so that group admins are shown the correct banner if their organisation has no org admins
- Updates the message in the banner to make the requirements clearer


app/views/groups/show.html.erb
DavidBiddle, line 1: This view has a lot of logic and I think would benefit from a refactor. I tried this but it ended up being bigger than expected so I want to tackle it separately from this PR

thomasiles (approved):
Looks good, nicely documented and works locally 👍

PR #2070, opened 2025-07-10
Enable new typography scale

DavidBiddle:

Trello card: https://trello.com/c/O48UNuIf/2364-switch-to-new-type-scale-for-admin

Switches on the [new govuk-frontend type scale](https://design-system.service.gov.uk/get-started/new-type-scale/).


thomasiles (approved):
Looks good on the review app ⚖️

PR #2072, opened 2025-07-11
2373 move feature report code from forms api to forms admin

SamJamCul:

Trello card: https://trello.com/c/5WfLpFRG/2373-move-feature-report-code-from-forms-api-to-forms-admin

Refactor all reports that are derived from the Reports model have been refactored into using a bespoke service - which has been copied from Form API. The main two are the Add Another Answer report and the Selection Questions report. 


lfdebrux (approved):
🚢

PR #2091, opened 2025-07-23
Add model method for moving forms between groups

DavidBiddle:

Trello card: https://trello.com/c/OOpJ005p/2394-add-model-method-and-unit-tests-for-moving-a-form-to-another-group

Adds a method to the form model which allows the form to be moved to a new group.

We'll use this later when we build the UI for moving forms between groups.

### DIFF
--- a/spec/models/form_spec.rb
+++ b/spec/models/form_spec.rb
@@ -296,4 +296,36 @@ RSpec.describe Form, type: :model do
       end
     end
   end
+
+  describe "#move_to_group" do
+    context "when the form is in an existing group" do
+      let(:form) { create(:form_record) }
+      let(:old_group) { create :group, name: "Group 1" }
+      let(:group_form) { GroupForm.find_by!(form_id: form.id) }
+
+      before do
+        GroupForm.create!(form_id: form.id, group: old_group)
+      end
+
+      context "when the user tries to move the form to a different group" do
+        let(:new_group) { create :group, name: "Group 2" }
+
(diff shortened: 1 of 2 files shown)


spec/models/form_spec.rb
```diff
@@ -296,4 +296,44 @@
...
+      context "when the new group does not exist" do
+        it "throws an ActiveRecord::RecordNotFound error" do
+          expect { form.move_to_group("some_nonexistent_external_id") }.to raise_error(ActiveRecord::RecordNotFound)
+        end
+      end
+    end
+
+    context "when the form is not already in a group" do
+      it "throws an ActiveRecord::RecordNotFound error" do
+        expect { form.move_to_group("some_nonexistent_external_id") }.to raise_error(ActiveRecord::RecordNotFound)
```
chao-xian, line 335: Is it possible for a form to not be in a group?

chao-xian: I think the before block needs to be inside the other context for this to not have a group? Or at least not have a GroupForm...

DavidBiddle: Good spot, have moved the before block!

It might be overly defensive - there _shouldn't_ be any forms in that state since we moved them all in bulk when we added groups. I figured it might still be theoretically possible, so if it happens an error would be useful so we know about it and can figure out why it happened. But I'm happy to remove it if you reckon it's unnecessary!

thomasiles (approved):

PR #2092, opened 2025-07-24
Allow org admins to delete empty groups

DavidBiddle:

Trello card: https://trello.com/c/Xfye5rPT/2398-enable-deleting-a-group-for-org-admins

We previously implemented the ability for super admins to delete empty groups. This PR updates the permissions to allow org admins to do the same thing.


thomasiles (approved):
works for me locally 👍

PR #2105, opened 2025-07-29
Add payment links to live view of form

thomasiles:

Trello card: https://trello.com/c/hbtuJ3ty/2379-show-if-a-payment-link-has-been-added-or-not-to-the-live-form-view-page

When displaying a live form, we don't show any information about payment
links.

This commit adds a payment link to the live view of a form if one is
present.

We have to be slightly careful to not rely on payment_url? as it may not
exist for forms published before we added the payment_url field.

image


deberny:
LGTM 🚀

DavidBiddle (approved):
Code change looks good and works for me in Review 🎉

PR #2109, opened 2025-07-30
Use Spring and Bootsnap to speed up RSpec

lfdebrux:

Recently RSpec load times have been in the region of 10 seconds on my machine, even when running tests for just one spec file. This makes the unit testing loop quite painful.

On my machine using Bootsnap and Spring with this PR brings the time to load RSpec down to less than a second.

It appears that most of the time is spent in loading Rails, but loading modules also has a contribution of a few seconds.

This PR aims to improve the situation by using some of the tools common to Rails for preloading applications, namely [Bootsnap] and [Spring]. It's a fairly heavy handed approach, but I haven't yet figured out what is taking so long. I have observed that on some machines RSpec is much quicker, so this approach might not be necessary, but as a quick and dirty fix it'll do for now.

Bootsnap is enabled by default, it can be disabled if needed with `DISABLE_BOOTSNAP=true`. Spring is optional, to use it run the tests as `bundle exec spring rspec`.

[Bootsnap]: https://github.com/Shopify/bootsnap
[Spring]: https://github.com/rails/spring

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -123,4 +123,6 @@ group :test do
 
   # axe-core for running automated accessibility checks
   gem "axe-core-rspec"
+
+  gem "spring-commands-rspec"
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -511,6 +511,9 @@ GEM
     snaky_hash (2.0.1)
       hashie
       version_gem (~> 1.1, >= 1.1.1)
+    spring (4.3.0)
+    spring-commands-rspec (1.0.4)
+      spring (>= 0.9.1)
     stringio (3.1.7)
     terminal-table (4.0.0)
       unicode-display_width (>= 1.1.1, < 4)
@@ -617,6 +620,7 @@ DEPENDENCIES
   sentry-rails
   sentry-ruby
   simplecov (~> 0.22.0)
+  spring-commands-rspec
   tzinfo
   tzinfo-data
   validate_url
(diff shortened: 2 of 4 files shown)


theseanything:
Does spring or bootsnap alone improve things?

lfdebrux:
> Does spring or bootsnap alone improve things?

Spring on it's own brings it down to around 3 seconds. Bootsnap on it's own brings it down to about 6 seconds. Both those load times are bit painful still.

It's worth noting that using the two together should be fine, it's explicitly called out [in the Bootsnap readme](
https://github.com/Shopify/bootsnap#:~:text=Note%3A%20Bootsnap%20and%20Spring%20are%20orthogonal%20tools.%20While%20Bootsnap%20speeds%20up%20the%20loading%20of%20individual%20source%20files%2C%20Spring%20keeps%20a%20copy%20of%20a%20pre%2Dbooted%20Rails%20process%20on%20hand%20to%20completely%20skip%20parts%20of%20the%20boot%20process%20the%20next%20time%20it%27s%20needed.%20The%20two%20tools%20work%20well%20together.):

> Note: Bootsnap and [Spring](https://github.com/rails/spring) are orthogonal tools. While Bootsnap speeds up the loading of individual source files, Spring keeps a copy of a pre-booted Rails process on hand to completely skip parts of the boot process the next time it's needed. The two tools work well together.

PR #2121, opened 2025-08-05
Add move forms

chao-xian:

Add ability for Org Admins to move forms into different groups they belong to.

To test: go to a group, then append `/forms/<form-id>/edit` for a form that's already in that group.

Trello card: https://trello.com/c/hetFDETo/2434-create-a-skeleton-page-for-moving-a-form


app/controllers/group_forms_controller.rb
```diff
@@ -41,18 +41,16 @@ def update
...
-    if @group.external_id == receiving_group.external_id
-      flash[:message] = "Form is already in this group."
-      render :edit
-    else
-      success_message = "'#{form.name}' has been moved to '#{receiving_group.name}'"
-      form.move_to_group(receiving_group.external_id)
+    if @group_select.valid?
+      receiving_group = Group.find(@group_select.group)
 
+      @group_form.update!(group: receiving_group)
```
thomasiles, line 49: I changed the from the `form.move_to_group` because it doesn't seem worth having a whole method to do it.

Feel free to change back though.

app/controllers/group_forms_controller.rb
```diff
@@ -41,18 +41,16 @@ def update
 
     form = Form.find(params[:id])
 
-    receiving_group = Group.find(group_select_params[:group])
-    @group_select = Forms::GroupSelect.new(group: receiving_group, form: form)
+    @group_select = Forms::GroupSelect.new(group_select_params.merge(form: form))
 
-    # TODO add spec for checking: compare @group with @group_select.group to check if it's changed
```
thomasiles: I took this check out because I'm not sure it's still a case we need to cover?

spec/support/selenium_error_patch.rb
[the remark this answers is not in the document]
chao-xian: I forgot to drop this - it's in a draft PR from Tom

config/locales/en.yml
```diff
@@ -434,6 +434,12 @@ en:
       receive_csv_subsection:
         receive_csv: Get completed forms as CSV files
         title: Optional task
+  group_forms:
+    edit:
+      body_html: "<p> We’ll send an email to members of the current group to let them know the form has moved and they may no longer have access to it. </p>\n"
+      form_name_label: Form name
+      group_name_label: Group name
```
hannahkc, line 441:
```suggestion
      group_name_label: Current group
```

spec/views/group_forms/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    assign(:form, form)
+    assign(:group, group)
+    assign(:group_form, group.group_forms.first)
+    assign(:group_select, group_select)
+    assign(:group_select_presenter, group_select_presenter)
+  end
+
+  it "renders the page title" do
+    render
+    expect(rendered).to have_css("h1", text: /Move form/)
```
DavidBiddle, line 23: This should fix the other test failures:

```suggestion
    expect(rendered).to have_css("h1", text: /Move this form/)
```

DavidBiddle (commented):
A few small comments from me but the overall approach looks good and it works for me in review

spec/features/group_form/move_form_spec.rb
```diff
@@ -0,0 +1,116 @@
+require "rails_helper"
+
+feature "Move a form", type: :feature do
+  describe "moving a form to another group" do
+    let(:group) { create(:group, organisation: organisation_admin_user.organisation) }
+    # rubocop:disable RSpec/LetSetup
+    let!(:another_group) { create(:group, organisation: organisation_admin_user.organisation) }
+    # rubocop:enable RSpec/LetSetup
```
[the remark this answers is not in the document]
chao-xian: I'm happier with explicitly showing that we are doing the rubocop bypass because the comment in the before block doesn't convey enough to say why we're doing that in the part of the code rather than as a let... my instinct if I were to look at the before change would be "why isn't this is a let?" and then I'd change it and rubocop would then complain...

spec/support/selenium_error_patch.rb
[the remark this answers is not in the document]
chao-xian: Have removed this commit now

PR #2122, opened 2025-08-07
WIP: Add auto complete for moving forms between groups

thomasiles:

Trello card: https://trello.com/c/ctgaaiHF/2435-frontend-work-to-select-a-group-to-send-forms-to

    This is a work in progress to flesh out the view for moving form between
    groups.
    
    - Add content from deigns to the view
    - Add form components based on number of groups
    - Add simple unit tests
    
    Still to do:
    
    - Clarify if select and autocomplete should start off blank, like radio buttons
    - Move content into locale files
    - Check against designs
    - Write commit messages

image


thomasiles:
This has been replaced by another PR which was merged.

PR #2131, opened 2025-08-12
Add total number of users to platform reports

alice-carr:

Previously people were manually adding up the number of users from the report to get a figure for total number of users. This commit instead adds the total number of users to the report itself

Trello card: https://trello.com/c/6JrjiXTN/2440-add-total-number-of-users-to-platform-reports


app/services/reports/users_report_service.rb
```diff
@@ -32,4 +35,9 @@ def org_name_user_count
       .order("organisations.name")
       .pluck("organisations.name", "COUNT(users.id)")
   end
+
+  def total_user_count
+    User.left_joins(:organisation)
+      .pick("COUNT(users.id)")
```
DavidBiddle, line 41: Since we're not using the organisation in this query, I think this can be simplified by using [ActiveRecord's count method](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-count):

```suggestion
    User.count
```

PR #2142, opened 2025-08-15
Rename unprocessable_entity to unprocessable_content

DavidBiddle:
When the 422 status was initially proposed in [RFC4918](https://datatracker.ietf.org/doc/html/rfc4918) (2007), it was called 'Unprocessable entity'.

When it became a standard in [RFC9110](https://datatracker.ietf.org/doc/html/rfc9110) (2022), it was renamed 'Unprocessable content'.

Rack now supports it as `:unprocessable_content` and [deprecated `:unprocessable_entity` in v3.1.0](https://github.com/rack/rack/blob/main/CHANGELOG.md#deprecated-1). This commit updates it to use the new name.

### DIFF
--- a/app/controllers/account/names_controller.rb
+++ b/app/controllers/account/names_controller.rb
@@ -16,7 +16,7 @@ module Account
         DefaultGroupService.new.create_user_default_trial_group!(current_user)
         redirect_to next_path
       else
-        render :edit, status: :unprocessable_entity
+        render :edit, status: :unprocessable_content
       end
     end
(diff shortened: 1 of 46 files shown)


DavidBiddle:
@lfdebrux Looks like it was deprecated in 3.1.0 which was a while ago. I'm not sure why we didn't see the deprecation message locally - I only spotted it in the CI output this morning

lfdebrux (approved):

PR #2155, opened 2025-08-19
Prepend empty line to markdown block elements

DavidBiddle:

Trello card: https://trello.com/b/9cKfKKdR/govuk-forms-sprint-board

Markdown block elements (in our case, headings and list items) should have an empty line between them and any previous content.

If this empty line isn't present, it's possible the markdown will not be rendered as the user intended.

This PR adds logic so that if the user adds a block element using the markdown toolbar, they'll automatically have an empty line added if there isn't already one present.


stephencdaly (approved):
Changes look good and are well tested. I've checked locally and it works as expected.

PR #2160, opened 2025-08-20
Fix parallel rspec link and gem environment

SamJamCul:

We have to do this so that the gem's rake task can be run using the command line without needing to specify the environment
   


stephencdaly (dismissed):

lfdebrux:
This is weird, it works fine on my machine...

PR #2173, opened 2025-08-28
Add form_id attribute to form_document serialization

lfdebrux:

Trello card: https://trello.com/c/sSOKmA23/2463-add-form-documents-table-to-forms-admin 

Form documents should have a form_id that has the ID of the form that created the form document (rather than its own ID). This commit makes the `Form#as_form_document` method match this behaviour, same as what is in forms-api.

We can fix any form documents that were created in the forms-admin database without the form_id attribute by running the `form_documents:sync` task.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -141,6 +141,7 @@ class Form < ApplicationRecord
       except: %i[state external_id pages question_section_completed declaration_section_completed share_preview_completed],
       methods: %i[start_page steps],
     )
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):
This matches what is returned by the forms-api API endpoint, although confusingly not what we store in the FormDocuments table in forms-api, where we store both, I think.

It makes sense to only include the form's ID once so this is the best way to keep it consistent.

PR #2187, opened 2025-09-01
Use localhost for test database connections

chao-xian:
This fixes an issue if using psql container in docker and running the tests on the host machine. By specifying "localhost" this ensures the connection is done via TCP rather than sockets, which is needed to connect to the container.

Follows https://github.com/alphagov/forms-runner/pull/1588

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -58,6 +58,9 @@ development:
 test:
   <<: *default
   database: forms_admin_test
+  username: postgres
+  password: postgres
+  host: localhost
   gssencmode: disable # needed to stop spring from segfaulting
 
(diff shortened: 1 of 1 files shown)


DavidBiddle (approved):

DavidBiddle:
I assumed this would work because it was copied across from forms-runner, but when I run `bin/setup` I get an error during the migration step:

> bin/rails aborted!
   ActiveRecord::StatementInvalid: PG::InsufficientPrivilege: ERROR:  permission denied for table schema_migrations (ActiveRecord::StatementInvalid)

DavidBiddle (approved):

PR #2191, opened 2025-09-03
Build FormReports using factories in report specs

stephencdaly:

Trello card: https://trello.com/c/KIdXtAro/2470-change-feature-reports-to-use-form-documents-in-forms-admin-database

Previously, we were loading a pre-made API response body in the format that forms-api would return from a JSON document.

To prepare for using FormDocuments retrieved from the forms-admin database rather than forms-api, replace use of the JSON document with creating FormDocuments from Forms created using the factories.


SamJamCul (approved):

lfdebrux (approved):
Looks good to me, thanks for splitting this out into a separate PR

PR #2193, opened 2025-09-04
Add custom error for missing draft question

DavidBiddle:

Trello card: https://trello.com/c/Cv7Df8ho/2452-navigating-to-the-new-page-view-causes-an-error-if-the-answer-type-and-answer-settings-arent-valid

Adds a new custom error page for when a user attempts to use the new or create question actions without having a draft question. Currently we just display a generic 500 error in this case.

### Testing instructions

The easiest way to see the new page is to go straight to the 'new question' page without setting up a draft question first: https://pr-2193.admin.review.forms.service.gov.uk/forms/1/pages/new/question

Screenshots:
Before:
A GOV.UK Forms error page with heading 'Sorry, there is a problem with the service' and a single line 'Please try again later' underneath

After:
A GOV.UK Forms error page with heading 'There’s a problem with this page' and three lines reading 'This page cannot be loaded because it’s trying to use expired data. This can happen if you use the back button or multiple tabs on your browser. Add a question or go to your list of questions to continue building your form.' underneath

### DIFF
--- a/app/controllers/pages/questions_controller.rb
+++ b/app/controllers/pages/questions_controller.rb
@@ -1,7 +1,5 @@
 class Pages::QuestionsController < PagesController
-  before_action do
-    raise "No answer type set for draft question" if draft_question.answer_type.blank?
-  end
+  before_action :check_draft_question, only: %i[new create]
 
   def new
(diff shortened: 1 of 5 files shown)


chao-xian (approved):

hannahkc:
Content looks good! thanks!

PR #2205, opened 2025-09-09
Add logging of request params

lfdebrux:

We don't systematically log user input to forms-admin, which makes it hard to reproduce and debug issues that occur in production. This problem will only grow once we have removed forms-api, because then we won't be able to look at the params in the forms-api request to work out what changes were made to a form.

This commit adds request parameters to our current logging attributes for all web controller actions. This includes path parameters such as a form ID and page ID, and also the parameters for inputs submitted with a POST/PATCH/PUT request.

For the most part we don't need to worry about logging PII here, because a) forms-admin deals with creating and editing forms rather than data collected by a form, and b) we already have a legal basis for logging user data to Splunk, and we do already log it.

### DIFF
--- a/app/controllers/web_controller.rb
+++ b/app/controllers/web_controller.rb
@@ -153,7 +153,7 @@ private
   end
 
   def params_to_log
-    @params_to_log ||= params.except(:controller, :action, :format, :id).to_unsafe_h
+    @params_to_log ||= params.except(:controller, :action, :format, :id, :authenticity_token).to_unsafe_h
   end
 
   def set_analytics_events
--- a/app/controllers/authentication_controller.rb
+++ b/app/controllers/authentication_controller.rb
@@ -83,4 +83,8 @@ private
     # the username/password flow
     params.permit(:auth)[:auth] == "e2e"
   end
+
+  def params_to_log
+    nil
+  end
 end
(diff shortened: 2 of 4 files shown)


app/controllers/web_controller.rb
```diff
@@ -147,6 +147,8 @@ def set_logging_attributes
         CurrentLoggingAttributes.user_organisation_slug = current_user.organisation&.slug
       end
     end
+
+    CurrentLoggingAttributes.params = params.except(:controller, :action, :format, :id).to_unsafe_h
```
stephencdaly, line 151: Should we also be excluding `authenticity_token`?

stephencdaly (approved):
Works as expected. Nice refactor for the logging tests!

Small questions about whether we should also exclude `authenticity_token`

app/controllers/web_controller.rb
lfdebrux: Huh, I hadn't seen that in `params` when running on my machine, where did it come up?

lfdebrux: Oh I see it now, I had only been paying attention to what was logged in the tests, but that doesn't include the CSRF token. Good catch, I'll fix this.

stephencdaly (approved):

PR #2221, opened 2025-09-15
Update to allow archived forms to move to any group

chao-xian:
This was in the design but the logic was wrong.


DavidBiddle (changes requested):
I think you also need to update the GroupSelectPresenter - I've archived a form in review and I'm still seeing 'live' content: https://pr-2221.admin.review.forms.service.gov.uk/groups/rNc9Xtbd/forms/3/edit

image

PR #2222, opened 2025-09-15
Handle a form in the wrong state when visiting /live or /archived

stephencdaly:

When visiting /live or /live/pages routes, check that the form is in a live state. If it has been archived, redirect to the equivalent page for an archived form. If the form has not been made live, render the 404 page.

Do the equivalent for /archived routes, redirecting to /live if the form is live, and rendering the 404 page if the form has never been made live.

Instead of rendering the not found page directly, raise a new NotFoundError which is handled by the error handling middleware to render the not found page. Remove unnecessary rescue_from calls from controllers and rely on the error handling middleware instead.

### DIFF
--- a/app/controllers/group_forms_controller.rb
+++ b/app/controllers/group_forms_controller.rb
@@ -27,7 +27,7 @@ class GroupFormsController < WebController
   end
 
   def edit
-    render template: "errors/not_found", status: :not_found unless set_group_form
+    raise NotFoundError unless set_group_form
 
     @form = FormRepository.find(form_id: params[:id])
     @group_select = Forms::GroupSelect.new(group: @group, form: @form)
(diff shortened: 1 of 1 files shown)


lfdebrux:
Is there a way we can get `current_form.archived_form_document` to raise `ActiveRecord::RecordNotFound`? Then errors would be raised through the normal route.

stephencdaly:
> Is there a way we can get `current_form.archived_form_document` to raise `ActiveRecord::RecordNotFound`? Then errors would be raised through the normal route.

I couldn't spot a way to do that. Do you know of one?

I kind of like checking the state of the form rather than whether there's a form document as it will raise an error to Sentry if somehow the form is in the correct state but the relevant FormDocument doesn't exist.

lfdebrux (dismissed):

PR #2229, opened 2025-09-17
Add filters and pagination to users page

stephencdaly:

Trello card: https://trello.com/c/IlcvX9A9/186-add-filters-and-pagination-for-users-page

Add a filter component an pagination to the users page so it loads quicker and it's easier to find users.

50 users are displayed per page - this could be changed if desired.

Add indexes on the users table to make the search using the filters as efficient as possible.

Screenshot 2025-09-18 at 16 46 11

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -3,6 +3,29 @@
     {
       "warning_type": "Mass Assignment",
       "warning_code": 105,
+      "fingerprint": "4aef20c77272b01619b5f67292f088db289359805fe0e21ef7bc05b6191d87ce",
+      "check_name": "PermitAttributes",
+      "message": "Potentially dangerous key allowed for mass assignment",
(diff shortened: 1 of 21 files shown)


config/brakeman.ignore
```diff
@@ -3,50 +3,49 @@
...
+        "class": "UsersController",
+        "method": "filter_params"
       },
       "user_input": ":role",
       "confidence": "Medium",
       "cwe_id": [
         915
       ],
-      "note": "Changing role for Membership is allowed."
+      "note": ""
```
DavidBiddle, line 24: Even though it's reasonably clear from the context, it feels like it would be useful to add a note to make it clear why we need this exception, and it probably good practice for us to always add a note when adding stuff to this file anyway.

DavidBiddle (approved):

PR #2232, opened 2025-09-19
Don't raise an error if form creator tries to make live form live

lfdebrux:

Although we don't have a link to the page to make a form live if the form is live without draft changes, it has occasionally happened that users have managed to double submit the request to make a form live. When this happens they will see a 500 error page, because our state machine raises an exception if it receives the `make_live` event when the form is already live (and doesn't have any draft changes). This isn't helpful for anyone, so instead if the form is in the live state we just show the confirmation page without doing anything else.


theseanything (approved):
Thanks this is a sensible iteration, rather than raising an error. In future, it might be worth letting the user know that the form was already live.

PR #2234, opened 2025-09-19
Add Users CSV download

stephencdaly:

Trello card: https://trello.com/c/IlcvX9A9

Display a CSV download link on the users index page. This link will download a CSV containing data about all users if no filters are applied, or the filtered users if filters are applied.

Screenshot 2025-09-19 at 16 07 32

Screenshot 2025-09-19 at 16 07 56

### DIFF
--- a/app/views/users/index.html.erb
+++ b/app/views/users/index.html.erb
@@ -65,6 +65,13 @@
 </div>
 
 <% if @users.any? %>
+
+  <% if @filter_input.has_filters? %>
+    <p><%= govuk_link_to t('users.index.download_filtered'), @filtered_download_path, no_visited_state: true %></p>
+  <% else %>
+    <p><%= govuk_link_to t('users.index.download_all'), download_users_path, no_visited_state: true %></p>
+  <% end %>
+
   <%= govuk_pagination(pagy: @pagy) %>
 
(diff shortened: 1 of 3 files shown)


app/views/users/index.html.erb
```diff
@@ -65,6 +65,13 @@
 </div>
 
 <% if @users.any? %>
+
+  <% if @filter_input.has_filters? %>
+    <p><%= govuk_link_to t('users.index.download_filtered'), @filtered_download_path, no_visited_state: true %></p>
+  <% else %>
+    <p><%= govuk_link_to t('users.index.download_all'), download_users_path, no_visited_state: true %></p>
```
SamJamCul, line 72: Just wanted to find out why the `no_visited_state` is set here

stephencdaly: > Use the govuk-link--no-visited-state modifier class where it is not helpful to distinguish between visited and unvisited states, for example when linking to pages with frequently-changing content such as the dashboard for an admin interface.

So because each time you download the CSV its potentially different, it doesn't help to have the visited state. As opposed to a link that takes you to a page with some content, where you'd want the appearance to change to the visited state after you've clicked it.

PR #2245, opened 2025-09-25
Add available languages column to Form model

chao-xian:

Trello card: https://trello.com/c/wLycLCbm/2480-add-availablelanguages-field-to-forms-in-forms-admin

### DIFF
--- a/app/models/form_document/content.rb
+++ b/app/models/form_document/content.rb
@@ -7,6 +7,7 @@ class FormDocument::Content
   attribute :form_id, :integer
   attribute :live_at, :datetime
   attribute :name, :string
+  attribute :available_languages, array: true
   attribute :language, :string
   attribute :form_slug, :string
   attribute :created_at, :datetime


db/migrate/20250925162215_set_en_to_existing_forms.rb
```diff
@@ -0,0 +1,8 @@
+class SetEnToExistingForms < ActiveRecord::Migration[8.0]
+  class Form < ApplicationRecord
+  end
+
+  def change
+    Form.where(language: nil).update_all(language: %w[en])
+  end
+end
```
thomasiles, line 8: I'm not sure we need this migration because the language column has always been not null with a default value of "en" so this migration has probably never been run.

The migration is also using the wrong value`%w[en]` is an array with a single "en" value. The language field expects text so it should be "en". Trying to set %[en] on language throws an error.

spec/models/form_spec.rb
```diff
@@ -67,6 +67,18 @@
...
+      expect(form).to be_invalid
+      expect(form.errors[:available_languages]).to include("can't be blank")
+    end
+
+    it "disallows invalid available languages" do
+      form.available_languages = %w[cn jp]
+      expect(form).to be_invalid
+      expect(form.errors[:available_languages]).to include("is not included in the list")
+    end
+
```
thomasiles, line 81: I think we need a test case for the valid case where we set avalable_languages to ["en", "cy"]

thomasiles (commented):
I've left a couple of comments, which maybe more related to my understanding of how language and available_languages work.

app/models/form.rb
```diff
@@ -15,10 +15,16 @@ class Form < ApplicationRecord
...
+  enum :language, {
+    en: "en",
+    cy: "cy",
+  }
+
   validates :name, presence: true
   validates :payment_url, url: true, allow_blank: true
   validate :marking_complete_with_errors
   validates :submission_type, presence: true
+  validates :available_languages, presence: true, inclusion: { in: ->(form) { form.language } }
```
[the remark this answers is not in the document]
chao-xian: Gah need more tests - thanks for kicking the tyres

db/migrate/20250925162215_set_en_to_existing_forms.rb
chao-xian: I double checked and you're right the default value gets set anyway (for the right column which is `available_languages` and that is supposed to be an array) so I'm going to drop the commit that adds this migration

db/migrate/20250924180913_add_available_languages_to_forms.rb
```diff
@@ -0,0 +1,5 @@
+class AddAvailableLanguagesToForms < ActiveRecord::Migration[8.0]
+  def change
+    add_column :forms, :available_languages, :text, array: true, default: %w[en]
+  end
```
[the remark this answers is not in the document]
chao-xian: I added it anyway

DavidBiddle (approved):

PR #2246, opened 2025-09-26
Add lockfile checksums

lfdebrux:

Bundler 2.6 added support for checksum verification as an opt-in beta feature [[1]]. We're on Bundler 2.7 now, and there don't appear to be any major issues with this feature, and the next major version of Bundler will include checksums in lockfiles by default.

So, we should enable it for our apps, as it protects against (some kinds of) supply chain attacks.

This commit includes the changes after running

    bundle lock --add-checksums

[1]: https://bundler.io/blog/2024/12/19/bundler-v2-6.html
[2]: https://github.com/rubygems/rubygems/blob/master/doc/bundler/UPGRADING.md


thomasiles (approved):
nice!

PR #2257, opened 2025-10-01
Add language as param to runner API

thomasiles:
### Add language query param when fetching form documents.
Trello card: https://trello.com/c/ZzFlTpQv/2481-support-retrieving-multilingual-forms-in-forms-admin

Add a language column to FormDocuments and an optional parameter called language to the API used by the runner.

These default to "en" but allow forms to have multiple FormDocuments in different languages.


lfdebrux (approved):
Looks good to me, tested locally

PR #2261, opened 2025-10-02
Remove FormRepository

stephencdaly:

The FormRepository was added for the purpose of migrating from making requests to forms-api using ActiveResource to updating forms in the forms-admin database using ActiveRecord.

Now that the migration is completed, we can remove this service and call methods on the `Form` ActiveRecord model directly.


spec/lib/tasks/forms.rake_spec.rb
```diff
@@ -147,19 +140,15 @@
...
-          task.invoke(*valid_args)
+          expect {
+            task.invoke(*valid_args)
+          }.to change { form.reload.submission_email }.to(submission_email)
         end
 
         it "updates the email confirmation status" do
-          expect(FormRepository).to receive(:save!).with(an_object_having_attributes(email_confirmation_status: :email_set_without_confirmation))
-          task.invoke(*valid_args)
+          expect(form.reload.email_confirmation_status).to eq(:email_set_without_confirmation)
```
lfdebrux, line 151: This shared example no longer calls `task.invoke` anywhere, is that right?

spec/services/create_form_service_spec.rb
```diff
@@ -10,28 +10,23 @@
...
 
-      expect(GroupForm.last).to have_attributes(form_id: created_form_id, group_id: 1000)
+      expect(GroupForm.last).to have_attributes(form_id: Form.last.id, group_id: 1000)
     end
 
     context "when a form with that name was already created in that group" do
       before do
-        allow(FormRepository).to receive(:create!).and_invoke(->(**attributes) { create(:form, **attributes) })
-        allow(FormRepository).to receive(:find).and_invoke(->(form_id:) { Form.find(form_id) })
+        allow(Form).to receive(:create!).and_invoke(->(**attributes) { create(:form, **attributes) })
```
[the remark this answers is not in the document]
stephencdaly: Thanks, have updated

spec/lib/tasks/forms.rake_spec.rb
stephencdaly: Good spot, have updated

PR #2262, opened 2025-10-03
Remove PageRepository

stephencdaly:

The PageRepostory was added for the purpose of migrating from making requests to forms-api using ActiveResource to updating forms in the forms-admin database using ActiveRecord.

Now that the migration is completed, we can remove this service and call methods on the Page ActiveRecord model directly


app/models/page.rb
```diff
@@ -25,6 +25,30 @@ class Page < ApplicationRecord
...
+      form_id:,
+      question_text:,
+      hint_text:,
+      is_optional:,
+      is_repeatable:,
+      answer_settings:,
+      page_heading:,
+      guidance_markdown:,
+      answer_type:,
+    )
```
lfdebrux, line 47: Rather than listing all the attributes and creating extra work for ourself, we can use the Ruby argument forwarding syntax [[1]]

```suggestion
  def self.create_and_update_form!(...)
    page = Page.new(...)
```

[1]: https://docs.ruby-lang.org/en/3.4/syntax/methods_rdoc.html#label-Argument+Forwarding

app/controllers/pages_controller.rb
```diff
@@ -63,7 +63,7 @@ def destroy
       return redirect_to @back_url
     end
 
-    unless PageRepository.destroy(page)
+    unless page.destroy
```
lfdebrux, line 66: I think we need to call `destroy_and_update_form` here?

stephencdaly: Ah good spot, missed that we had that method 🤦

stephencdaly: Thanks, I've updated and made sure we handle the `ActiveRecord::RecordNotFound` error in the controller and added a test for this.

PR #2268, opened 2025-10-06
Remove condition repository

stephencdaly:

The ConditionRepository was added for the purpose of migrating from making requests to forms-api using ActiveResource to updating forms in the forms-admin database using ActiveRecord.

Now that the migration is completed, we can remove this service and call methods on the Condition ActiveRecord model directly.


app/models/condition.rb
```diff
@@ -10,10 +10,8 @@ class Condition < ApplicationRecord
 
   def save_and_update_form
     save!
-    # TODO: https://trello.com/c/dg9CFPgp/1503-user-triggers-state-change-from-live-to-livewithdraft
-    # Will not be needed when users can trigger this event themselves through the UI
-    form.create_draft_from_live_form if form.live?
```
lfdebrux, line 15: Huh, I didn't even notice this bug before, saving a condition didn't switch `archived` to `archived_with_draft`. That's a plus for this commit reducing repetition!

spec/models/condition_spec.rb
```diff
@@ -64,6 +64,78 @@
...
+      }.to change(described_class, :count).by(1)
+    end
+
+    it "creates the condition with the given attributes" do
+      created_condition = described_class.create_and_update_form!(**condition_params)
+      expect(created_condition.check_page_id).to eq(routing_page.id)
+      expect(created_condition.routing_page_id).to eq(routing_page.id)
+      expect(created_condition.goto_page_id).to eq(goto_page.id)
+      expect(created_condition.answer_value).to eq("Yes")
+      expect(created_condition.skip_to_end).to be(false)
```
lfdebrux, line 93: It might be simpler to use the `have_attributes` [[1](https://rubydoc.info/gems/rspec-expectations/3.13.5/RSpec/Matchers#have_attributes-instance_method)] matcher?

```suggestion
      expect(created_condition).to have_attributes(
        check_page_id: routing_page.id,
        routing_page_id: routing_page.id,
        goto_page_id: goto_page.id,
        answer_value: "Yes",
        skip_to_end: false,
      )
```

spec/requests/pages/conditions_controller_spec.rb
```diff
@@ -314,10 +310,6 @@
         it "redirects to the confirm exit page deletion page" do
           expect(response).to redirect_to confirm_change_exit_page_path(form_id: form.id, page_id: page.id, condition_id: condition.id, params: { answer_value: "Wales", goto_page_id: 3 })
         end
-
-        it "does not call save! for the condition" do
-          expect(ConditionRepository).not_to have_received(:save!)
-        end
```
lfdebrux, line 323: Can we have a test that checks that the condition is not updated to replace this one?

lfdebrux (approved):
Looks good to me! A couple of comments, but none important

spec/requests/pages/conditions_controller_spec.rb
stephencdaly: Added this, also added a missing check that the condition was updated for the happy path

PR #2444, opened 2025-12-17
Fix viewcomponent preview links

DavidBiddle:

Trello card: 


lfdebrux:
I think this can be closed as we made the same change in https://github.com/govuk-forms/forms-admin/pull/2636 (I somehow missed this draft PR at the time 😅)

PR #2671, opened 2026-03-30
Add optional track

thomasiles:
### Track the CSV download on the Welsh translation page

Trello card: https://trello.com/c/oIf48Bv9/2862-add-google-analytics-event-for-downloading-the-translation-spreadsheet

This PR adds some new javascript to track links which have the `data-track-link` attribute set.

For example:
```HTML
<a href="/path" data-track-link>Link text</a>
```

It then adds this attribute to the Welsh CSV download link so that it should be tracked, in the same way we track external links.

It also adds the CSV format to the Welsh download link, to make it clearer that it's a CSV file.

### DIFF
--- a/app/frontend/entrypoints/application.js
+++ b/app/frontend/entrypoints/application.js
@@ -7,7 +7,8 @@ import {
   installAnalyticsScript,
   sendPageViewEvent,
   attachExternalLinkTracker,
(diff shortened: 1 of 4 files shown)


stephencdaly (approved):
Code makes sense to me and it's well tested

PR #2691, opened 2026-04-09
Log start and end for rake tasks

SamJamCul:

https://trello.com/c/eBgfVxRZ/227-get-our-rake-tasks-to-show-up-in-splunk-more-clearly

There are a couple of things happening here, but the jist of it is that every task will be wrapped with a starting log, setting the current logging attribute for the task name, and then depending on how the task resolves (normally or with an error etc.) it’ll do a finish log too.

### DIFF
--- a/lib/tasks/logging.rake
+++ b/lib/tasks/logging.rake
@@ -1,0 +1,43 @@
+require "English"
+
+Rake::Task.define_task(:environment).enhance do
+  task_finished_normally = false
+
+  at_exit do
+    unless task_finished_normally
+      exit_cause = $ERROR_INFO ? $ERROR_INFO.class.name : "Signal or exit"
+
+      Rails.logger.error "Task terminated early", {
+        task: Rake.application.top_level_tasks.first,
+        exit_cause:,
+      }
+    end
+  end
+
+  Rake.application.top_level_tasks.each do |task_name|
+    task_name_clean = task_name.gsub(/\[.*\]$/, "")
+    next unless Rake::Task.task_defined?(task_name_clean)
+
+    task = Rake::Task[task_name_clean]
+    original_actions = task.actions.dup
+    task.actions.clear
+
+    task.enhance do |t, args|
+      CurrentTaskLoggingAttributes.task_name = task_name_clean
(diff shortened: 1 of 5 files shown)


app/models/current_logging_attributes.rb
```diff
@@ -15,5 +15,6 @@ class CurrentLoggingAttributes < ActiveSupport::CurrentAttributes
             :answer_type,
             :auth0_session_id,
             :validation_errors,
-            :params
+            :params,
+            :task_name
```
stephencdaly: Do you reckon we should put this in a separate CurrentAttributes class to keep it clear what we should expect to log? In forms-runner we have CurrentJobLoggingAttributes for logging for jobs. Should we add a CurrentTaskLoggingAttributes class for this?

SamJamCul: good shout, I'll do another class, maybe the args can go in there too.

lib/tasks/logging.rake
```diff
@@ -0,0 +1,43 @@
...
+      Rails.logger.info "Task started", { args: args.to_a }
+
+      begin
+        original_actions.each { |action| action.call(t, args) }
+        Rails.logger.info "Task finished"
+      rescue SystemExit => e
+        Rails.logger.error "Task aborted", { exit_message: e.message }
+        raise e
+      rescue StandardError => e
+        Rails.logger.error "Task failed", { class: e.class.name, error_message: e.message }
```
lfdebrux: Might be worth calling this `error_class`?


```suggestion
        Rails.logger.error "Task failed", { error_class: e.class.name, error_message: e.message }
```

lfdebrux: Alternatively, we could log `exception: [class name, error message]`, to match what we do with lograge elsewhere (search Splunk for `index="gds_dsp_production_forms" "exception{}"="*"` for examples):

```suggestion
        Rails.logger.error "Task failed", { exception: [e.class.name, e.message] }
```

SamJamCul: good shout!

lfdebrux (approved):

PR #2694, opened 2026-04-13
Remove DefaultGroupService

stephencdaly:
This service existed because Forms orginially were not created in groups. When groups were added, we moved all live forms for users who were a member of an organisation into a group for their organisation, and draft forms into a trial group for the user.

There were some users that had not selected an organisation, so their forms were not moved into groups. The DefaultGroupService was used to move those forms into a default trial group for the user after they selected their organisation.

There are no longer any forms that don't have a group, so this code can be removed.


thomasiles (approved):
LGTM

PR #2717, opened 2026-04-22
Add separate make live confirmations for each language

DavidBiddle:

Trello card: https://trello.com/c/VUw58sEk/2891-add-confirmation-pages-for-english-welsh-when-a-form-version-has-been-made-live

Adds new pages for making English and Welsh live for the first time separately. 

These pages will be made visible in the tasklist in a future ticket.

#### View for making English live:
image

#### English confirmation (when form's Welsh translation is not ready to be made live)
image

#### English confirmation (when form's Welsh translation is ready to be made live)
image

#### View for making English live:
image

#### Welsh confirmation
image

### Testing instructions
You won't be able to access the new views through the UI, but you can get there by going directly to the routes:
- `/forms/{form_id}/make-live/en`
  - your form will need to:
    - be ready to be made live
    - not have a live or archived English version
    - not have a live welsh version
- `/forms/{form_id}/make-live/cy`
  - your form will need to:
    - be ready to be made live
    - have a live English version
    - not have a live welsh version

### DIFF
--- a/app/controllers/forms/make_language_live_controller.rb
+++ b/app/controllers/forms/make_language_live_controller.rb
@@ -1,0 +1,37 @@
+module Forms
+  class MakeLanguageLiveController < FormsController
+    def new
+      authorize current_form, :can_make_language_live?
+      @make_language_live_input = MakeLiveInput.new(form: current_form)
+
+      render "new", status:, locals: { current_form: }
+    end
+
+    def create
+      authorize current_form, :can_make_language_live?
+
+      @make_language_live_input = MakeLiveInput.new(**make_language_live_input_params)
+
+      return render_new(status: :unprocessable_content) unless @make_language_live_input.valid?
+      return redirect_to form_path(@make_language_live_input.form.id) unless @make_language_live_input.confirmed?
+
+      @make_form_live_service = MakeFormLiveService.call(current_form:, current_user:, language: params[:language])
+      # TODO: actually make the language live
+
+      @go_to_make_welsh_live_input = GoToMakeWelshLiveInput.new
+
+      render "confirmation", locals: {
+        current_form:,
+        confirmation_page_title: @make_form_live_service.page_title,
+        confirmation_page_body: @make_form_live_service.confirmation_page_body,
+        language: params[:language],
+      }
+    end
+
+  private
+
+    def make_language_live_input_params
+      params.require(:forms_make_live_input).permit(:confirm).merge(form: current_form)
+    end
+  end
+end
--- a/app/controllers/forms/make_language_live_controller.rb
+++ b/app/controllers/forms/make_language_live_controller.rb
@@ -3,8 +3,7 @@ module Forms
     def new
       authorize current_form, :can_make_language_live?
       @make_language_live_input = MakeLiveInput.new(form: current_form)
-
-      render "new", status:, locals: { current_form: }
+      render_new
     end
 
     def create
@@ -15,12 +14,10 @@ module Forms
       return render_new(status: :unprocessable_content) unless @make_language_live_input.valid?
       return redirect_to form_path(@make_language_live_input.form.id) unless @make_language_live_input.confirmed?
 
-      @make_form_live_service = MakeFormLiveService.call(current_form:, current_user:, language: params[:language])
+      @make_form_live_service = MakeFormLiveService.call(current_form:, current_user:)
       # TODO: actually make the language live
 
-      @go_to_make_welsh_live_input = GoToMakeWelshLiveInput.new
-
-      render "confirmation", locals: {
+      render "forms/make_live/confirmation", locals: {
         current_form:,
         confirmation_page_title: @make_form_live_service.page_title,
         confirmation_page_body: @make_form_live_service.confirmation_page_body,
@@ -32,6 +29,10 @@ module Forms
 
     def make_language_live_input_params
       params.require(:forms_make_live_input).permit(:confirm).merge(form: current_form)
+    end
+
+    def render_new(status: :ok)
+      render "new", status:, locals: { current_form:, language: params[:language] }
     end
   end
 end
--- a/app/controllers/forms/make_language_live_controller.rb
+++ b/app/controllers/forms/make_language_live_controller.rb
@@ -2,7 +2,10 @@ module Forms
   class MakeLanguageLiveController < FormsController
     def new
       authorize current_form, :can_make_language_live?
+      return redirect_to form_path(form_id: current_form.id) unless current_form.can_make_language_live?(language: params[:language])
+
       @make_language_live_input = MakeLiveInput.new(form: current_form)
+
       render_new
     end
 
@@ -13,9 +16,10 @@ module Forms
 
       return render_new(status: :unprocessable_content) unless @make_language_live_input.valid?
       return redirect_to form_path(@make_language_live_input.form.id) unless @make_language_live_input.confirmed?
+      return redirect_to form_path(form_id: current_form.id) unless current_form.can_make_language_live?(language: params[:language])
 
       @make_form_live_service = MakeFormLiveService.call(current_form:, current_user:, language: params[:language])
-      # TODO: actually make the language live
+      @make_form_live_service.make_language_live
 
       redirect_to make_language_live_show_confirmation_path
     end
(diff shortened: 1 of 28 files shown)


app/services/form_document_sync_service.rb
```diff
@@ -52,6 +52,44 @@ def synchronize_archived_welsh_form
     end
   end
 
+  def synchronize_live_english_form
+    FormDocument.transaction do
+      # Ensure we only make English version live if there is no existing live Welsh version
+      raise ActiveRecord::RecordNotFound, "Cannot make English version live if there is already a live Welsh version." if FormDocument.where(form:, tag: LIVE_TAG, language: "cy").exists?
```
stephencdaly: The comment and error message here confused me a little as we can make changes to the English version live if there is already a live Welsh version, we just don't use this method.

```suggestion
      # If we've already made the Welsh version live, changes to the Welsh version must be made live at the same time by calling a different method
      raise ActiveRecord::RecordNotFound, "Cannot make changes to only the live English form if there is already a live Welsh version." if FormDocument.where(form:, tag: LIVE_TAG, language: "cy").exists?
```

app/services/form_document_sync_service.rb
```diff
@@ -52,6 +52,44 @@ def synchronize_archived_welsh_form
     end
   end
 
+  def synchronize_live_english_form
```
stephencdaly: If we added the word "only", this might make it clearer that this method isn't called when both languages are updated at the same time. I was getting a bit confused about what method would be called when, and if this is always called when we make changes to multilingual forms.

```suggestion
  def synchronize_only_live_english_form
```

app/models/form.rb
```diff
@@ -268,6 +274,30 @@ def after_make_live
...
+  def before_make_welsh_live
+    before_make_live
+  end
+
+  def after_make_welsh_live
+    FormDocumentSyncService.new(self).synchronize_live_welsh_form
+  end
+
+  def can_make_english_version_live?
+    draft? && ready_for_live && live_form_document.blank? && live_welsh_form_document.blank?
```
stephencdaly: Is it correct that we can't make the English version live if the task to add the Welsh translation is in progress? I had to remove the existing Welsh translations for `ready_for_live` to return true.

app/controllers/forms/make_language_live_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      render_new
+    end
+
+    def create
+      authorize current_form, :can_make_language_live?
+
+      @make_language_live_input = MakeLiveInput.new(**make_language_live_input_params)
+
+      return redirect_to form_path(@make_language_live_input.form.id) unless @make_language_live_input.confirmed?
+      return render_new(status: :unprocessable_content) unless @make_language_live_input.valid?
```
stephencdaly, line 17: I think we need to check the input is valid before we check whether it's confirmed. If I don't select an option, I don't see an error.

app/models/form.rb
stephencdaly: Also, can we not make changes to the English version live after it's already live but the Welsh version hasn't been made live yet?

app/services/form_document_sync_service.rb
```diff
@@ -52,6 +52,44 @@ def synchronize_archived_welsh_form
...
+      update_or_create_form_document(LIVE_TAG, content, "en")
+
+      # Update the content of the live English version to to not include Welsh in available_languages
+      FormDocument.where(form:, tag: [LIVE_TAG], language: "en").find_each do |live_document|
+        live_document.content["available_languages"] = %w[en]
+        live_document.save!
+      end
+
+      # A new live version replaces any previous archived version
+      delete_form_documents_by_tag(ARCHIVED_TAG)
```
stephencdaly, line 70: If there was previously a Welsh version that was archived, this will delete the Welsh archived form document. Do we want to do that?

app/controllers/forms/make_language_live_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+        language: params[:language],
+      }
+    end
+
+    def submit_confirmation
+      authorize current_form, :can_make_language_live?
+
+      @go_to_make_welsh_live_input = GoToMakeWelshLiveInput.new(**go_to_make_welsh_live_input_params)
+
+      if @go_to_make_welsh_live_input.confirmed?
```
stephencdaly, line 45: Should we validate the input for this and show an error message if a radio button isn't selected?

app/controllers/forms/make_language_live_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      return redirect_to form_path(form_id: current_form.id) unless current_form.can_make_language_live?(language: params[:language])
+
+      @make_form_live_service = MakeFormLiveService.call(current_form:, current_user:, language: params[:language])
+      @make_form_live_service.make_language_live
+
+      if current_form.state_previously_changed?
+        OrgAdminAlertsService.new(form: current_form, current_user:).form_made_live
+      end
+
+      @go_to_make_welsh_live_input = GoToMakeWelshLiveInput.new
```
stephencdaly: I don't think we need this as we're redirecting?

stephencdaly (commented):
Some very complex logic going on here, but you've made it pretty clear and the tests are pretty comprehensive.

There's just a couple of places where we're not validating the radio button input successfully.

Also left a few questions about some of the logic and a couple of style suggestions that you don't need to take if you don't think they're improvements.

app/services/form_document_sync_service.rb
DavidBiddle: I wasn't 100% sure on this, but I went with this because I think it matches the existing behaviour if you:
- archive the Welsh version
- make a new English version live

stephencdaly: Ok makes sense! If it doesn't work how people expect we can always change it later

app/models/form.rb
DavidBiddle: Thanks, good spot - I've updated this behaviour a bit to make it possible to:
- make English live when the Welsh is still in progress
- make English live if there's already an English version and the Welsh hasn't been made live

DavidBiddle:
Thanks for the feedback @stephencdaly, I think I've now updated everything

app/models/form.rb
```diff
@@ -267,6 +273,30 @@ def after_make_live
...
+  def before_make_welsh_live
+    before_make_live
+  end
+
+  def after_make_welsh_live
+    FormDocumentSyncService.new(self).synchronize_only_live_welsh_form
+  end
+
+  def can_make_english_version_live?
+    draft? && ready_for_live(ignore_missing_welsh: true) && live_welsh_form_document.blank?
```
stephencdaly: If we want to be able to make changes to the English version live again before the Welsh is completed, this needs to return true if the state is live_with_draft

```suggestion
    has_draft_version && ready_for_live(ignore_missing_welsh: true) && live_welsh_form_document.blank?
```

app/controllers/forms/make_language_live_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      render_new
+    end
+
+    def create
+      authorize current_form, :can_make_language_live?
+
+      @make_language_live_input = MakeLiveInput.new(**make_language_live_input_params)
+
+      return render_new(status: :unprocessable_content) unless @make_language_live_input.valid?
```
stephencdaly, line 17: I think we need to make a change to the `MakeLiveInput` to not validate the Welsh task is complete if we're just making the English version live. If the Welsh task is in progress and I try to make the English form live, I see an error on the confirmation page 

Image

stephencdaly (commented):
Most of it is now working as expected, I think there's 2 more small changes we need to make.

app/input_objects/forms/make_live_input.rb
```diff
@@ -8,12 +8,16 @@ class Forms::MakeLiveInput < ConfirmActionInput
...
-    form.all_incomplete_tasks.each do |section|
+    form.all_incomplete_tasks(ignore_missing_welsh:).each do |section|
       errors.add(:confirm, section.to_sym)
     end
 
     errors.empty?
   end
+
+  def ignore_missing_welsh
+    language == "cy"
```
stephencdaly: Should this be:

```suggestion
    language == "en"
```

DavidBiddle: Good catch - yes, it should

app/controllers/forms/make_language_live_controller.rb
stephencdaly: This still isn't working, I think because the controller is trying to get the `language` param from the `forms_make_live_input` whereas it's a path param so is at the root level of the `params`

DavidBiddle: I think this should be resolved now, thanks for flagging it!

stephencdaly (approved):

PR #2721, opened 2026-04-23
Update routing content from "check your answers page" to "end of form"

stephencdaly:

Trello card: https://trello.com/c/LLGjGbXI

In the new journey flow, the ‘end’ of a form will now be the question about whether the user wants a copy of their answers (if this setting is turned on).

Update the wording for routing from "check your answers page" to "end of form" to reflect this.

Also update the internal values used in the code for adding routing to refer to "end_of_form" rather than "check_your_answers". Have added some input object validation to make sure we don't break routing for users if anyone is adding/editing a route when this change is deployed.

### DIFF
--- a/app/input_objects/pages/conditions_input.rb
+++ b/app/input_objects/pages/conditions_input.rb
@@ -48,7 +48,7 @@ class Pages::ConditionsInput < BaseInput
 
   def goto_page_options
     page_options = pages_after_position(page.position + 1).map { |p| OpenStruct.new(id: p.id, question_text: p.question_with_number) }
-    page_options << OpenStruct.new(id: "check_your_answers", question_text: I18n.t("page_conditions.check_your_answers"))
+    page_options << OpenStruct.new(id: "end_of_form", question_text: I18n.t("page_conditions.end_of_form"))
 
     page_options
   end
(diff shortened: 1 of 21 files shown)


StephenGill:
@stephencdaly copy changes all look good to me 👍

app/input_objects/pages/conditions_input.rb
```diff
@@ -94,6 +95,15 @@ def secondary_skip?
 
 private
 
+  def goto_page_id_valid
+    return if goto_page_id.nil?
+    return if %w[end_of_form create_exit_page exit_page].include? goto_page_id
+
+    unless form.pages.find { |page| page.id.to_s == goto_page_id.to_s }
+      errors.add(:goto_page_id, :invalid, message: I18n.t("activemodel.errors.models.pages/secondary_skip_input.attributes.goto_page_id.invalid"))
```
DavidBiddle: This is very minor but I think you could make this a `return if` as well for consistency with the other valid cases:

```suggestion
    return if %w[end_of_form create_exit_page exit_page].include? goto_page_id
    return if form.pages.find { |page| page.id.to_s == goto_page_id.to_s }
    
    errors.add(:goto_page_id, :invalid, message: I18n.t("activemodel.errors.models.pages/secondary_skip_input.attributes.goto_page_id.invalid"))
```

DavidBiddle (dismissed):
Looks good to me - have raised one minor suggestion but feel free to merge if you don't think it's an improvement 🎉

app/input_objects/pages/conditions_input.rb
stephencdaly: Yeh that's clearer, thanks. I've updated.

DavidBiddle (approved):

PR #2722, opened 2026-04-24
Fix tests being skipped by Rake tasks aborting (and hopefully prevent same in future)

lfdebrux:

We had an issue recently where tests were being silently skipped in CI because a test added in PR #2691 was raising SystemExit and not catching it, and we had to revert that PR.

This PR restores the changes in that PR, adds a guard to try and catch similar issues in future (where Rake tasks call `abort` in specs and it's not `expect`ed), and fixes the (now) failing tests that caused the problems before.

We also refactor the way we write Rake task specs a bit to reduce boilerplate, and hopefully make it easier to write them (as well as safer).

### DIFF
--- a/app/lib/application_logger.rb
+++ b/app/lib/application_logger.rb
@@ -32,7 +32,11 @@ private
     msg = yield if block_given?
 
     begin
-      message_to_hash(msg).merge(attribs || {}).merge(CurrentLoggingAttributes.attributes).compact
+      message_to_hash(msg)
+        .merge(attribs || {})
+        .merge(CurrentLoggingAttributes.attributes)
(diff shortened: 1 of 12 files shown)


stephencdaly:
Thanks for looking at this, the changes look sensible and I can see that all the tests are now completing - but the rake task tests are currently failing

lfdebrux:
> Thanks for looking at this, the changes look sensible and I can see that all the tests are now completing - but the rake task tests are currently failing

Yep, I just noticed that also 🤦 I'll put this in draft and mark it ready to review hopefully soon.

lfdebrux:
@stephencdaly Okay, tests are fixed now!

stephencdaly (approved):

PR #2723, opened 2026-04-24
Bring existing org admin alerts into new structure

SamJamCul:

https://trello.com/c/QpJ0wA3H/2919-bring-existing-org-admin-alerts-into-new-structure

 - `org_admin_alerts` has been renamed to `admin_alerts`, as there are alerts that are sent to group_admins as well as org_admins that will come under this heading. 
 - Existing alerts that are sent to org and group admins have been relocated under `admin_alerts`. 
 - An update to the Notify personalisation for the alerts when a form moves group - this will need to go in before the Notify template gets updated. 


thomasiles (approved):
LGTM

PR #2726, opened 2026-04-27
Update Notify personalisation for move group alert

SamJamCul:
I missed the alternate content that shows when the group is in a trial state, and we'll need to toggle it on in those cases. 

https://trello.com/c/QpJ0wA3H/2919-bring-existing-org-admin-alerts-into-new-structure

### DIFF
--- a/app/mailers/admin_alerts/group_forms_move_mailer.rb
+++ b/app/mailers/admin_alerts/group_forms_move_mailer.rb
@@ -12,7 +12,13 @@ class AdminAlerts::GroupFormsMoveMailer < GovukNotifyRails::Mailer
   end
 
   def form_moved_email(to_email:, form_name:, old_group_name:, new_group_name:, org_admin_email:, org_admin_name:, group_active:)
-    group_active_text = group_active ? "yes" : "no"
+    if group_active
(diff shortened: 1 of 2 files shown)


alice-carr (dismissed):

app/mailers/admin_alerts/group_forms_move_mailer.rb
```diff
@@ -13,6 +13,7 @@ def form_moved_email_group_admin(...)
 
   def form_moved_email(to_email:, form_name:, old_group_name:, new_group_name:, org_admin_email:, org_admin_name:, group_active:)
     group_active_text = group_active ? "yes" : "no"
+    group_trial_text = group_active ? "no" : "yes"
```
lfdebrux: It might be clearer to have this as an "if else" block?

```suggestion
    if group_active
      group_active_text = "yes"
      group_trial_text = "no"
    else
      group_active_text = "no"
      group_trial_text = "yes"
    end
```

SamJamCul: Sure thing!

lfdebrux (approved):

PR #2728, opened 2026-04-27
add send_filler_answers feature flag

alice-carr:

added the feature flag to enable or disable sending a filler's answers to them. This is a feature flag for the whole environment rather than a group level flag.

This feature flag isn't being used to toggle functionality yet, this is in advance of the implementation of sending the filler's answers.

Trello card: https://trello.com/c/lr2stOI0/2905-create-new-feature-flag-in-admin-for-send-fillers-answers-feature

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -4,6 +4,7 @@ features:
     enabled_by_group: true
   multiple_branches:
     enabled_by_group: true
+  send_filler_answers: true
 
 forms_api:
   # Authentication key to authenticate with forms-api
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -4,6 +4,7 @@ features:
     enabled_by_group: true
   multiple_branches:
     enabled_by_group: true
+  send_filler_answers: true
```
stephencdaly, line 7: do we usually set this to false so it's disabled by default as we develop the feature?

alice-carr: I set this to true because the ticket said to enable it in local testing, but maybe I should set it to false and whenever someone is working on the feature they can manually change it?

stephencdaly: Ah, as long as the forms-deploy PR is merged first it probably doesn't matter much either way for this feature. Maybe we leave it off by default usually if it will interrupt other development if its switched on, but for this feature it won't.

stephencdaly (approved):

PR #2744, opened 2026-04-30
Display multiple branches on add/edit page

SamJamCul:
This is based on the Group's multiple_branches_enabled feature flag - so if this is toggled off, then it will use the original add and edit page view.

When the flag is turned on for the group, the display logic assumes that the form has been created using the new Routes process, and will not be particularly compatible with forms that have been created using the old routing pages (although it might work if the routing is relatively simple).

There is no validation displayed for these changes, as we're not ready to start deciding how to validate the new multiple-branch routes.

I've removed the edit links for now, we'll need to add them in once we've got the questions listed out on the routes page, as they should be anchored links. 

https://trello.com/c/IgUz33Ml/3070-make-edit-questions-list-display-with-new-routes

### DIFF
--- a/app/components/page_list_component/view.html.erb
+++ b/app/components/page_list_component/view.html.erb
@@ -32,33 +32,57 @@
             </dd>
           </div>
 
-          <%  page.routing_conditions.each do |condition| %>
-            <% condition_page = self.condition_page(condition) %>
-            <% condition_page_position = self.condition_page_position(condition) %>
-            <div id="<%= PageListComponent::ErrorSummary::View.error_id(condition.id) %>" class="govuk-summary-list__row app-page-list__row <%= class_names( "govuk-form-group--error": condition.has_routing_errors?) %>">
-              <dt class="govuk-summary-list__key govuk-summary-list__key app-page-list__key">
-                <%= t("page_conditions.condition_name", question_number: condition_page_position) %>
+          <% if FeatureService.new(group: @form.group).enabled?(:multiple_branches) && page.routing_conditions.present? %>
+            <div class="govuk-summary-list__row">
+              <dt class="govuk-summary-list__key app-page-list__key">
+                <%= t("page_conditions.condition_name", question_number: page.position) %>
               </dt>
-
               <dd class="govuk-summary-list__value">
-                <ul class="govuk-list govuk-!-margin-0">
-                  <% condition.validation_errors.each do |error| %>
-                    <li class="app-page_list__route-text--error">
-                      <%= PageListComponent::ErrorSummary::View.generate_error_message(error.name, condition:, page: condition_page) %>
-                    </li>
-                  <% end %>
-                </ul>
-
-                <p>
-                  <%= condition_description(condition) %>
-                </p>
-
-              <dd class="govuk-summary-list__actions govuk-!-padding-bottom-6">
-                <%= govuk_link_to show_routes_path(form_id: @form.id, page_id: condition.check_page_id) do %>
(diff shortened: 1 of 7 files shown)


app/components/page_list_component/view.html.erb
```diff
@@ -32,33 +32,57 @@
             </dd>
           </div>
 
-          <%  page.routing_conditions.each do |condition| %>
-            <% condition_page = self.condition_page(condition) %>
-            <% condition_page_position = self.condition_page_position(condition) %>
-            <div id="<%= PageListComponent::ErrorSummary::View.error_id(condition.id) %>" class="govuk-summary-list__row app-page-list__row <%= class_names( "govuk-form-group--error": condition.has_routing_errors?) %>">
+          <% if @form.group.multiple_branches_enabled? && page.routing_conditions.present? %>
```
lfdebrux: I'd suggest using the feature service to check whether the feature is enabled, rather than looking at the form record directly.

```suggestion
          <% if FeatureService.new(@form.group).enabled?(:multiple_branches) && page.routing_conditions.present? %>
```

Then in the tests whether the flag is enabled or not can be changed by mocking the feature service with the helpers in spec/support/features.rb, removing the need to add a group to the form in each test.

app/components/page_list_component/view.html.erb
```diff
@@ -32,33 +32,57 @@
...
+            <div class="govuk-summary-list__row">
               <dt class="govuk-summary-list__key govuk-summary-list__key app-page-list__key">
```
lfdebrux:
```suggestion
              <dt class="govuk-summary-list__key app-page-list__key">
```

app/components/page_list_component/view.rb
```diff
@@ -28,6 +28,14 @@ def condition_description(condition)
       end
     end
 
+    def branch_condition_description(condition)
```
lfdebrux: What about calling this `condition_description2` or something like that? `branch_condition_description` sounds to me like it's only for branch conditions.

app/components/page_list_component/view.html.erb
SamJamCul: Sounds good, thanks for the suggestion

app/components/page_list_component/view.html.erb
SamJamCul: Good spot!

app/components/page_list_component/view.rb
SamJamCul: Done

spec/components/page_list_component/page_list_component_preview.rb
```diff
@@ -3,15 +3,15 @@ class PageListComponent::PageListComponentPreview < ViewComponent::Preview
 
   def default
     pages = []
-    form = build(:form, id: 0, pages:)
+    form = build(:form, :with_group, id: 0, pages:)
```
lfdebrux, line 8: I can't get the component preview working locally on my machine after this change, I get an error

```
PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "organisations_pkey"
DETAIL: Key (id)=(1) already exists.
```

I think because the `:group` factory tries to create an organisation with ID 1, to make the tests more predictable.

I'm not sure what the fix is here, the issue with hardcoded organisation IDs in the tests has been around for a while and might be a pain to fix, and I'm not sure the previews not working is a big enough issue to block merging this PR... Maybe the easiest thing to do is to put a `TODO` comment reminding us to remove the `:with_group` trait when we no longer need the multiple branches feature flag.

lfdebrux (dismissed):

spec/components/page_list_component/page_list_component_preview.rb
SamJamCul: Sounds like a plan

lfdebrux (approved):

PR #2748, opened 2026-05-05
Add API endpoint to get the group for a form

stephencdaly:

Trello card: https://trello.com/c/CsEqMtDL

Add an API endpoint that returns the group details for the form with the given ID. This includes the group admin users, and the organisation details including the organisation admin users as embedded resources.

This endpoint will be used by forms-runner to get the group and organisation details for the form and the contact details for the group and organisation admin users in order to send email notifications when submissions bounce.

### DIFF
--- a/spec/requests/api/form_documents_controller_spec.rb
+++ b/spec/requests/api/form_documents_controller_spec.rb
@@ -3,7 +3,7 @@ require "rails_helper"
 RSpec.describe Api::FormDocumentsController, type: :request do
   let(:headers) { { "ACCEPT": "application/json" } }
 
-  describe "GET /show" do
+  describe "#show" do
     context "when the form exists" do
       context "when the tag is draft" do
         let(:draft_form_name) { "Draft form" }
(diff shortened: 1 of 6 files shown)


spec/requests/api/form_documents_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe Api::FormDocumentsController, type: :request do
   let(:headers) { { "ACCEPT": "application/json" } }
 
-  describe "GET /show" do
+  describe "GET #show" do
```
lfdebrux: I think what was there before was correct, based on the other request specs in this repo, and the [template in rspec-rails](https://github.com/rspec/rspec-rails/blob/main/lib/generators/rspec/request/templates/request_spec.rb)

```suggestion
  describe "GET /show" do
```

lfdebrux (dismissed):

spec/requests/api/form_documents_controller_spec.rb
stephencdaly: Hmm actually I think neither are consistent with how we usually do it elsewhere.
Looks like we do `describe "GET /path"` or `describe "#action"`

lfdebrux: Oh yeah you're right `/show` is an action not a path

stephencdaly: Updated to be `describe "#show"` in line with other tests

lfdebrux (approved):

PR #2756, opened 2026-05-06
Make task status service aware of current user

DavidBiddle:

Trello card: No card for main change, but also includes https://trello.com/c/TVN9BNxQ/3448-refactor-formtaskstatusservice-to-use-dependency-injection

Passes the current user to the task status service. This follows the work in https://github.com/govuk-forms/forms-admin/pull/2745 so that the task statuses are consistent everywhere.

This resolves an issue we were seeing where:
- the Welsh translation page had a non-gov.uk email address which matched the user's domain
- the page as a whole was valid and marked complete
- but the task in the task list was still showing as 'In progress' 
- the form could not be made live.

Also includes a factory change to make testing Welsh forms easier (cherry-picked from https://github.com/govuk-forms/forms-admin/pull/2717)

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -25,7 +25,7 @@ private
 
   def load_form
     @current_form = Form.find(params[:form_id])
-    @current_form.set_task_status_service(TaskStatusService.new(form: @current_form))
+    @current_form.set_task_status_service(TaskStatusService.new(form: @current_form, current_user:))
     @initial_form_state = @current_form.state
   end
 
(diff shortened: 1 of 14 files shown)


stephencdaly (dismissed):
Changes look good and the bug is fixed. 

Do you reckon it's worth adding a test for the scenario where the user enters a non-government email in the Welsh translations to check that the validation passes to prevent a regression?

DavidBiddle:
@stephencdaly good shout, I've added a test for that now

stephencdaly (approved):

PR #2757, opened 2026-05-07
Add multiple branch update

thomasiles:
### Allow form creators to add multiple routes to forms

Trello card: https://trello.com/c/HB5UeyuU/3069-make-routing-screen-update-routes

There is a lot in this PR. It also skips a lot to make it easier to review. As multiple branches are still behind the feature flag I think it's better we refine it as we go.

This PR doesn't attempt to cover:
- exit pages
- any validations
- removing earlier values from the selection boxes
- forms with either 0 or 1 question

We are adding a new form to allow form creators to view and change the routes for each page in their form.

Selection questions which only take one answer can have multiple routes; one for each selection option. Every other question type has one route. If the form creator hasn't specified a route for a page or option, it will route to the next page by page position. The last page in a form cannot have a route from it, it always goes to the next page.

We are not changing the data model for our forms, we are still modelling routes with `Condition`s. To reduce the complexity of displaying, creating and validating routes, we add two new input objects, `RouteInput` and `RoutesInput`.

`RoutesInput` represents all the routes for a given form. It's main job is to be a container for `RouteInputs`, which represent the routes for all the pages within a form. We can add validations which take involve multiple routes or pages here. 

To stop `RoutesInput` becoming too complex (which has happened with other large input objects), I've added two supporting services, `Routes::BuildService` and `Routes::SyncService`. These are responsible for converting a form with pages and conditions into an array of `RouteInputs` and then from an array back into `Condition`s to be saved.

An example form:
image

Success banner:
image

### Content questions:
- How should we display "None of the above" options. For now they are just shown as an extra option.
- What should the success banner content be?
- Normal pages have a hidden label above the selection question for screen readers which says "Go to". We don't have to have it but I think we should consider labels, otherwise it might not be clear what we are asking form creators to do. We might consider it in other places, like the question numbers too?
- What should we display on the page when no questions have been added to the form yet? What about when there is only one question, which wouldn't have any options for routing?

### Technical things which can be improved in separate commits:
- The view takes some shortcuts. The labels are defined in the input objects, which mixes presentation a bit. It also puts all routes in a list, even when there is only one. I think in general it could be refactored and there might be better HTML element choices we could make.
- I haven't tested performance or checked that there are no N+1 queries
- Making sure we aren't relying too much on user input. Syncing relies on the generated RouteInput's being valid. I think we could do a bit more checking to make sure the pages are valid within the form and that only selection questions can accept RouteInputs with answer_values
- Saving the routes doesn't completely clear all current routes and build only the ones needed. I'm not sure what behaviour we need here.
- There is no special handling of the last page in a form. I think that should always be given the default route so that any existing routes are cleared.

I also think we should consider taking this pattern further and make the RoutesInput a plain ActiveModel and using it everywhere we need to display routes or validation errors. It might make development easier having the logic in a single place.

### DIFF
--- a/app/input_objects/forms/route_input.rb
+++ b/app/input_objects/forms/route_input.rb
@@ -1,0 +1,31 @@
+class Forms::RouteInput < BaseInput
+  include ActiveModel::Attributes
+
+  END_OF_FORM_VALUE = "end_of_form".freeze
+  DEFAULT_VALUE = "default".freeze
+
+  attribute :id # id of the Condition
+  attribute :page_id, :integer
+  attribute :goto
+  attribute :answer_value
+
+  attr_accessor :page, :goto_options, :label
+
+  def goes_to_default_next_page?
+    goto == DEFAULT_VALUE
+  end
+
+  def goes_to_end_of_form?
+    goto == END_OF_FORM_VALUE
+  end
+
+  def condition_attributes
+    if goes_to_end_of_form?
+      { goto_page_id: nil, skip_to_end: true, check_page_id: page.id }
+    elsif goes_to_default_next_page?
+      nil
+    else
+      { goto_page_id: goto, skip_to_end: false, check_page_id: page.id }
+    end
+  end
+end
--- a/app/services/routes/build_service.rb
+++ b/app/services/routes/build_service.rb
@@ -1,0 +1,101 @@
+##
+# Takes a form (with pages and conditions) and builds an array of RouteInput objects
+# based on the pages and conditions.
+class Routes::BuildService
+  END_OF_FORM_OPTION = [I18n.t("page_conditions.end_of_form"), Forms::RouteInput::END_OF_FORM_VALUE].freeze
+  NONE_OF_THE_ABOVE_OPTION = DataStruct.new(value: :none_of_the_above.to_s, name: I18n.t("page_conditions.none_of_the_above"))
+
+  attr_reader :form
+
+  def initialize(form:)
+    @form = form
+  end
+
+  def build_routes
+    conditions_by_key = form.conditions.index_by do |c|
+      [c.routing_page_id, c.answer_value.presence]
+    end
+
+    form.pages.flat_map do |page|
+      if page.answer_type == "selection" && page.answer_settings.only_one_option == "true"
+        build_routes_for_selection_page(page, conditions_by_key)
+      else
+        build_route_for_generic_page(page, conditions_by_key)
+      end
+    end
+  end
+
+  def options_for_page(page)
+    return [] unless page.has_next_page?
+
+    # Don't include the current page in the options
+    options_without_page = all_goto_options.reject { |_, value| value == page.id }
+
+    # Replace the next page with the default option
+    options_without_page.map do |name, value|
+      if value == page.next_page
+        ["Go to question #{page.position.next}", Forms::RouteInput::DEFAULT_VALUE]
+      else
+        [name, value]
+      end
+    end
+  end
+
+private
+
+  def build_routes_for_selection_page(page, conditions_by_key)
+    options = page.answer_settings&.selection_options || []
+
+    options << NONE_OF_THE_ABOVE_OPTION if page.is_optional
+
+    options.map.with_index(1) do |option, index|
+      answer_value = option["value"]
+      answer_value_label = answer_value == NONE_OF_THE_ABOVE_OPTION[:value] ? I18n.t("page_conditions.none_of_the_above") : option["name"]
+      key = [page.id, answer_value]
+      condition = conditions_by_key[key]
+
+      Forms::RouteInput.new(
+        id: condition&.id,
+        page_id: page.id,
+        page:,
+        answer_value:,
+        goto: goto_value_for(condition),
+        goto_options: options_for_page(page),
+        label: { text: "Option #{index}: #{answer_value_label}" },
+      )
+    end
+  end
+
+  def build_route_for_generic_page(page, conditions_by_key)
+    key = [page.id, nil]
+    condition = conditions_by_key[key]
+
+    [
+      Forms::RouteInput.new(
+        id: condition&.id,
+        page_id: page.id,
+        page:,
+        goto: goto_value_for(condition),
+        goto_options: options_for_page(page),
+        label: { text: "Go to", hidden: true },
+      ),
+    ]
+  end
+
+  def all_goto_options
+    @all_goto_options ||= begin
+      page_opts = form.pages.map { |p| ["#{p.position}. #{p.question_text}", p.id] }
+      page_opts + [END_OF_FORM_OPTION]
+    end
+  end
+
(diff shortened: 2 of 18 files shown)


app/services/routes/build_service.rb
```diff
@@ -0,0 +1,101 @@
...
+      condition = conditions_by_key[key]
+
+      Forms::RouteInput.new(
+        id: condition&.id,
+        page_id: page.id,
+        page:,
+        answer_value:,
+        goto: goto_value_for(condition),
+        goto_options: options_for_page(page),
+        label: { text: "Option #{index}: #{answer_value_label}" },
```
thomasiles, line 64: It feels a bit awkward to include labels in the model, but it makes things easier. I think we could consider somewhere else for this logic.

app/views/routes/show.html.erb
```diff
@@ -8,5 +8,42 @@
...
+      <% end %>
+
+      <%= govuk_summary_list(actions: false) do |summary_list|
+
+        @routes_input.routes.group_by(&:page).each do |page, routes|
+          summary_list.with_row do |row|
+            row.with_key(classes: "app-routes-list__key", html_attributes: { id: "page-#{page.position}"}) { page.position.to_s }
+            row.with_value do
+              capture do %>
+              <p> <%= page.question_text %> </p>
```
thomasiles, line 24: I'm not sure which element is best for this, so I've stuck with paragraphs.

app/input_objects/forms/route_input.rb
```diff
@@ -0,0 +1,29 @@
+class Forms::RouteInput < BaseInput
+  include ActiveModel::Attributes
+
+  END_OF_FORM_VALUE = "end_of_form".freeze
+  DEFAULT_VALUE = "default".freeze
```
lfdebrux, line 5: The commit message says these are symbols, is there a reason to use frozen strings instead?

app/input_objects/forms/route_input.rb
```diff
@@ -0,0 +1,29 @@
...
+  attribute :goto
+  attribute :answer_value
+
+  attr_accessor :page, :goto_options, :label
+
+  def goes_to_default_next_page?
+    goto == DEFAULT_VALUE
+  end
+
+  def goes_to_check_your_answers?
```
lfdebrux: With the One Login work, the end of the form might not be the check your answers page any more (see https://github.com/govuk-forms/forms-admin/pull/2721), maybe we should call this `goes_to_end_of_form?` instead?

```suggestion
  def goes_to_end_of_form?
```

app/input_objects/forms/route_input.rb
```diff
@@ -0,0 +1,29 @@
...
+    goto == END_OF_FORM_VALUE
+  end
+
+  def condition_attributes
```
lfdebrux, line 22: Should this return `nil` if the route goes to the next page?

app/services/routes/build_service.rb
```diff
@@ -0,0 +1,101 @@
+##
+# Takes a form (with pages and conditions) and builds an array of RouteInput objects
+# based on the pages and conditions.
+class Routes::BuildService
+  END_OF_FORM_OPTION = ["Check your answers before submitting", Forms::RouteInput::END_OF_FORM_VALUE].freeze
```
lfdebrux: I think with the changes in https://github.com/govuk-forms/forms-admin/pull/2721 this should be "End of form" (or better, `I18n.t("page_conditions.end_of_form")`.

```suggestion
  END_OF_FORM_OPTION = [I18n.t("page_conditions.end_of_form"), Forms::RouteInput::END_OF_FORM_VALUE].freeze
```

app/services/routes/build_service.rb
```diff
@@ -0,0 +1,101 @@
...
+  def options_for_page(page)
+    return [] unless page.has_next_page?
+
+    # Don't include the current page in the options
+    options_without_page = all_goto_options.reject { |_, value| value == page.id }
+
+    # Replace the next page with the default option
+    options_without_page.map do |name, value|
+      if value == page.next_page
+        ["Go to question #{page.position.next}", :default]
```
lfdebrux: Should this be `Forms::RouteInput::DEFAULT_VALUE`?

```suggestion
        ["Go to question #{page.position.next}", Forms::RouteInput::DEFAULT_VALUE]
```

christophercameron-ixd:
**How should we display "None of the above" options. For now they are just shown as an extra option.**

I _think_ we would not include these as something they can interact with, for now atleast, just so we don’t confuse it with the existing feature where a ‘none of the above’ answer can lead to an extra text input or screen. BUT I am happy to have my mind changed on this if others have strong feelings. 

**What should the success banner content be?**

“Your routes have been saved” feels fine to me for adding or removing routes, as long as something has changed ideally. 

**Normal pages have a hidden label above the selection question for screen readers which says "Go to". We don't have to have it but I think we should consider labels, otherwise it might not be clear what we are asking form creators to do. We might consider it in other places, like the question numbers too?**

Good point! We will need to make each label unique to each question, i.e. not just “Go to”, but I’m not sure what makes the most sense. Would it makes sense to use the question text itself as the label, or is that too complicated/wouldn’t make sense? Otherwise, we could just go with “After question X go to” as a hidden label until Will fixes it 😁  

**What should we display on the page when no questions have been added to the form yet? What about when there is only one question, which wouldn't have any options for routing?**

Either we could hide the “Edit routes” button so they don’t see the screen OR we could just have some plain text under the H1 that says something like “You have no questions in your form.” 🤷    
Where it has only 1 question I think it’s fine to show it without the drop-down (as it is technically the last question in the form so should act that way?)

app/input_objects/forms/route_input.rb
thomasiles: They started off as symbols because I think it makes more sense for constants like these.

They get compared to the goto value, which is either an integer or a string. Because goto can be either type, and integers don't have `.to_sym`, I converted the constants to strings to do the comparison.

I decided to just make the constants strings because it seemed more direct. I'll happily make them symbols again if it's clearer or there is something I'm missing. Symbols feel nicer.

app/input_objects/forms/route_input.rb
thomasiles: yes, good point!

I realised halfway through check your answers isn't the right thing anymore but didn't catch them all.

app/input_objects/forms/route_input.rb
thomasiles: Good idea. I hadn't done it because we skip them when saving them. Without adding this case, it's returning nonsense. I've fixed it and added some tests, which were missing.

app/services/routes/build_service.rb
thomasiles: good point!

lfdebrux (approved):

app/services/routes/build_service.rb
lfdebrux: I think this is probably fine for now

PR #2763, opened 2026-05-11
Add column to store enabling/disabling Copy of answers on forms

chao-xian:

This is a small PR to add the database column that stores whether a Form can be set to ask fillers if they want a Copy Of Their Answers.

Trello card: https://trello.com/c/KZuqH0bA 

### DIFF
--- a/app/models/form_document/content.rb
+++ b/app/models/form_document/content.rb
@@ -32,6 +32,7 @@ class FormDocument::Content
   attribute :what_happens_next_markdown, :string
   attribute :send_daily_submission_batch, :boolean
   attribute :send_weekly_submission_batch, :boolean
+  attribute :send_copy_of_answers, :string
 
   alias_attribute :id, :form_id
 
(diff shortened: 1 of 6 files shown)


alice-carr (dismissed):

app/models/form_document/content.rb
```diff
@@ -32,6 +32,7 @@ class FormDocument::Content
   attribute :what_happens_next_markdown, :string
   attribute :send_daily_submission_batch, :boolean
   attribute :send_weekly_submission_batch, :boolean
+  attribute :email_copy_answers, :string
```
stephencdaly: Very much an opinion but could this column name be made more readable? Maybe just by adding "of" - `email_copy_of_answers`, or `get_copy_of_answers` or just `copy_of_answers`

stephencdaly (dismissed):

stephencdaly (dismissed):

stephencdaly (approved):

PR #2764, opened 2026-05-12
Mark slow specs with `:slow` tag

lfdebrux:

On my machine (MacBook Pro, Apple M1 Pro, 16 GB RAM, macOS Tahoe 26.4.1), running the specs in this repo takes around 17 minutes.

This PR adds a `:slow` tag to examples that consistently take over 1 second to run on my machine.

This is partly for documentation purposes, but also allows developers to skip slow specs when running tests locally with

```
rspec --tag ~slow
```

See the [RSpec docs on the `--tag` option](https://rspec.info/features/3-12/rspec-core/command-line/tag/) for details on how this works.

The changes in this PR should not affect how the tests/PR checks run in CI.

On my machine, with the changes from this PR, running `rspec --tag ~slow` takes around 7 minutes.

#### Notes on slow tests

While trying to figure what to tag as slow, I found three different kinds of slow tests:

1. Tests that rely on browser automation

  Anything that uses headless Chrome for testing is automatically a resource hog; I didn't look too closely at these. Instead I just added a bit of configuration to make all feature specs be marked as slow.

  In theory we could try and speed some feature specs up by seeing which ones require a real browser and which could use the RackTest simulator that comes with Capybara [[1](https://github.com/teamcapybara/capybara/#racktest)]. However, in practice, most of our feature specs require a real browser because they a) rely on JavaScript for the sign in journey, and b) test for WCAG failures with aXe tools, which requires Chrome. So it would be quite a lot of work to try and speed up our feature specs this way (although it might become worth it at some point).

2. Tests that are consistently slow

  There were a few specs that were consistently significantly slower than a second when run on their own. These mostly appear to be weird outliers, and there also aren't that many of them. I've fixed two (see commit messages for details), and left the rest marked with the `:slow` tag.

3. Tests that are intermittently slow

  This was the biggest category of issue I found. When repeatedly running tests with `rspec --profile --tag ~slow`, after having marked tests in the above two categories, I found there was little consistency between runs in which tests were in the top ten slowest list.

  After doing some profiling with [ruby-prof](https://ruby-prof.github.io/) using a modified version of https://gist.github.com/palkan/73395cc201a565ecd3ff61aac44ad5ae, I found one issue related to ViteRuby and added a workaround for that (see commit messages for details), but adding that workaround did not fully resolve the intermittent slowness. I also found that some specs can be slow when run in isolation, but quick when run as part of a group, likely because there is some shared caching happening.

  Basically, some of the slowness is likely caused by a combination of Factory Bot (a known issue generally), bootsnap caching, thread contention, and general operating system constraints, and these are all difficult to pin down and fix.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,7 +56,7 @@ jobs:
         run: bin/rails db:schema:load
       # Add or replace test runners here
       - name: Run Ruby tests
-        run: bundle exec rspec
+        run: bundle exec rspec --profile
       - name: Run JS tests
         run: npm test
       # Add or replace any other lints here
--- a/config/vite.json
+++ b/config/vite.json
@@ -19,6 +19,7 @@
     "autoBuild": true
   },
   "test": {
+    "host": "::1",
     "port": 3037,
     "autoBuild": true
   }
--- a/spec/i18n_spec.rb
+++ b/spec/i18n_spec.rb
@@ -2,7 +2,7 @@
 
 require "i18n/tasks"
 
-RSpec.describe I18n do
+RSpec.describe I18n, :slow do
   let(:i18n) { I18n::Tasks::BaseTask.new }
   let(:missing_keys) { i18n.missing_keys }
   let(:unused_keys) { i18n.unused_keys }
--- a/spec/requests/authentication_controller_spec.rb
+++ b/spec/requests/authentication_controller_spec.rb
@@ -83,7 +83,7 @@ RSpec.describe AuthenticationController, type: :request do
         GDS::SSO::Config.auth_valid_for = Settings.auth_valid_for
       end
 
-      it "re-authenticates after the configured time" do
+      it "re-authenticates after the configured time", :slow do
         login_as_standard_user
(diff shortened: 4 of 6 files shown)


lfdebrux:
One thing I should have done sooner before starting to look at this is compare how slow the tests are on a local dev machine to how long they take in GitHub Actions. They are so much faster in GitHub Actions it's ridiculous, and the fact they do it with less resources makes me wonder if it's a macOS thing, or something related to endpoint protection or other systems on our developer laptops. That would also align with what I saw when profiling, that there were no consistent hotspots, which might suggest influences not related to the code under test.

At some point I will try running the tests on a personal laptop just to see how fast they are there; however I don't have a personal macOS laptop currently, just Linux machines, so that might not fully resolve matters.

Also, in that comment I didn’t spend too much time looking at browser tests; in hindsight that may have been a mistake. There’s definitely some speed-up possible, because we’re seeing tests be so much faster in GitHub Actions. On the other hand, looking at the browser tests might not have revealed much more than I found anyway.

PR #2774, opened 2026-05-13
Update add edit routes content

SamJamCul:

https://trello.com/c/IgUz33Ml/3070-make-edit-questions-list-display-with-new-routes-and-content-update

Updates to the content for the Add and edit questions page for multiple branches. This does _not_ handle exit pages at all. 

Additionally, the list of answer values, when there are multiple that lead to the same page,  is not ordered, and will shuffle around if routes are added and deleted. I found it was a bit fiddly to try and fix that, I'll take another crack at it in a bit.

___
### When a page has a single condition based on an answer value

Note also that the key shows it's a singular `route`

image

___

### When a page has multiple conditions
Conditions are grouped by their destination page's position, with skip to end conditions at the end. The key also shows that there are multiple routes, so uses the plural

image

___
### Unconditional route to a question
image

___
### Unconditional skip to the end

image

### DIFF
--- a/app/components/page_list_component/view.html.erb
+++ b/app/components/page_list_component/view.html.erb
@@ -35,7 +35,7 @@
           <% if FeatureService.new(group: @form.group).enabled?(:multiple_branches) && page.routing_conditions.present? %>
             <div class="govuk-summary-list__row">
               <dt class="govuk-summary-list__key app-page-list__key">
-                <%= t("page_conditions.condition_name", question_number: page.position) %>
+                <%= t("page_conditions.condition_list_key", count: page.routing_conditions.count, question_number: page.position) %>
               </dt>
               <dd class="govuk-summary-list__value">
                 <% if page.routing_conditions.first.answer_value.present? %>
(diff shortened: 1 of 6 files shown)


app/components/page_list_component/view.html.erb
```diff
@@ -39,17 +39,19 @@
...
                 <% if page.routing_conditions.first.answer_value.present? %>
-                  <p><%= I18n.t("page_conditions.if_answer_is") %></p>
-                  <ul class="govuk-list govuk-list--bullet">
-                    <% page.routing_conditions.each do |condition| %>
-                      <li>
-                        <%= condition_description2(condition) %>
-                      </li>
-                    <% end %>
-                  </ul>
+                  <% answer_value_groups(page).each do |group| %>
```
lfdebrux: In the code within this block we use `group[1]`, I think it would be more idiomatic to split each item from the enumerator with `|_, group|`, and then use `group`

lfdebrux (dismissed):

app/components/page_list_component/view.html.erb
SamJamCul: Done!

lfdebrux (dismissed):

hannahkc:
content in the screenshots all looks good to me!

lfdebrux (approved):

PR #2775, opened 2026-05-14
Add new groups confirmation page

thomasiles:
### Add a new page with content to slow when creating a new group

Trello card: https://trello.com/c/KETW4If8/3099-add-a-new-page-after-a-user-clicks-create-a-group

This PR adds a new page between clicking "Create group" and giving the group a name. It also updates the content on the new groups page.

The new page is to confirm the user wants to create a group and give them more information about what a group is. The content is different for organisations which have an admin.

The existing new group page can still be accessed directly through URL. The create group button link has been changed to this new page.

> [!NOTE]
> I made up some content for the validation message the user sees when they don't check yes or no and submit the page. 

## Without org admins 
image

## With org admins
image

## Without selecting yes/no
image

## The updated new group page
image

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -1,5 +1,5 @@
 class GroupsController < WebController
-  before_action :set_group, except: %i[index new create confirm_new]
+  before_action :set_group, except: %i[index new create confirm_new confirm_new_submit]
   after_action :verify_authorized, except: :index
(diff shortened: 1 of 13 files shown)


lfdebrux:
This looks good to me from a dev perspective

lfdebrux (approved):

william-castoncook:
Looks good to me!

PR #2776, opened 2026-05-14
Don't allow goto page to be before routing page for show/edit routes page

lfdebrux:

Trello card: https://trello.com/c/58uS0Pah/3087-restrict-the-options-for-routes-next-page 

On the page to edit routes for a form, we want each select field to only show pages after the page it is routing from.


thomasiles (dismissed):
nice!

thomasiles (approved):

PR #2782, opened 2026-05-15
Add documentation  for the v2 API

theseanything:
This outline what requests can be made and the data model. Having documentation makes it far easier to see what the API looks like and understand its behaviour.


theseanything:
We're going to building a new API soon, closing for now

PR #2784, opened 2026-05-15
Limit branching to only lists with 10 or fewer options (for now)

lfdebrux:

Trello card: https://trello.com/c/pOOto1kO/3098-limit-branching-to-only-lists-with-10-or-fewer-options-for-now 

We've decided for now to avoid presenting form creators with a very long routing page if they have a selection question with a long list of options.


thomasiles (approved):
nicely done 🏅

PR #2785, opened 2026-05-15
Add message to routes#show when not enough pages

thomasiles:
### Stop showing an empty form on the routes page

Trello card: https://trello.com/c/bOtP0QNS/3086-remove-the-save-and-continue-button-from-the-routes-page-when-there-are-less-than-2-pages-in-the-form

Instead of showing the routes form when there are not enough pages, show a warning and a link to the question pages.

We show the message when there is 0 or 1 page.

### DIFF
--- a/app/views/routes/show.html.erb
+++ b/app/views/routes/show.html.erb
@@ -9,46 +9,50 @@
       <%= t("page_titles.routes") %>
     </h1>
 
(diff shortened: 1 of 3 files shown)


lfdebrux (dismissed):

hannahkc:
Content looks grand!

lfdebrux (approved):

PR #2791, opened 2026-05-19
Change live/archived pages view for multiple branches

SamJamCul:
When a form has the multiple_branches_enabled flag set, the live and archived show pages has a different format to display the routes for a form.

https://trello.com/c/LtIlAUBk/3072-show-new-routes-in-live-archived-view

The live and archived form view shows multiple branch routes appropriately.

---
### Single conditional route
image

---
### Multiple conditional routes
image

---
### Unconditionally go to a page
image

--- 
### Unconditionally go to the end of the form
image

---

### Fixing the none_of_the_above bug for the current forms as well
before: 
image

after: 
image

### DIFF
--- a/app/controllers/forms/archived_controller.rb
+++ b/app/controllers/forms/archived_controller.rb
@@ -15,7 +15,7 @@ module Forms
       return redirect_to live_form_pages_path if current_form.is_live?
       raise NotFoundError unless current_form.is_archived?
 
-      render :show_pages, locals: { form_document: current_archived_form, welsh_form_document: }
+      render :show_pages, locals: { form_document: current_archived_form, welsh_form_document:, multiple_branches_enabled: FeatureService.new(group: current_form.group).enabled?(:multiple_branches) }
     end
 
   private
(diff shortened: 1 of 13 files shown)


hannahkc:
Content in the screenshots looks perfect! thank you.

app/controllers/forms/archived_controller.rb
```diff
@@ -15,7 +15,7 @@ def show_pages
       return redirect_to live_form_pages_path if current_form.is_live?
       raise NotFoundError unless current_form.is_archived?
 
-      render :show_pages, locals: { form_document: current_archived_form, welsh_form_document: }
+      render :show_pages, locals: { form_document: current_archived_form, welsh_form_document:, current_form: }
```
lfdebrux: I think if we just need the multiple branches feature flag value it might be easier to just pass that through.

Or if we do need the whole form object, can we call it `form_metadata` to match how the current form is passed through for the `show_form` action?

SamJamCul: I think passing the flag through is clearer why we're doing it, I'll change to that.

lfdebrux (approved):

PR #2797, opened 2026-05-20
Fix content for none of the above option when adding and editing routes

lfdebrux:

While working on multiple branches, we noticed that the existing routing implementation is missing content for the situation when the answer value for a condition is the none of the above option for an optional selection question; the pages rendered the answer value as `none_of_the_above`.

This PR adds the missing content.

#### Showing a route

Screenshot of http://localhost:3000/forms/8/pages/31/routes

#### Deleting a route

Screenshot of http://localhost:3000/forms/8/pages/31/conditions/11/delete

#### Adding a secondary skip to a route

Screenshot of http://localhost:3000/forms/8/pages/31/routes/any-other-answer/questions-to-skip/new

### DIFF
--- a/app/components/page_list_component/view.rb
+++ b/app/components/page_list_component/view.rb
@@ -53,7 +53,7 @@ module PageListComponent
 
     def answer_value_text_for_condition(condition)
       if condition.answer_value.present?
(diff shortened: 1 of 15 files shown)


hannahkc:
Content looks grand - thank so much for doing this!

SamJamCul (approved):
Great to get this stuff cleaned up

PR #2810, opened 2026-05-27
Add tasks for making individual languages live

DavidBiddle:

Trello card: https://trello.com/c/nE8FUqW4/2892-separate-make-your-english-welsh-form-live-task-on-task-page

Adds new tasks to the task list, which will appear when the form:

- has any Welsh content, and
- does not have a live Welsh version

These replace the existing make live task for Welsh forms.

### Screenshots
A form with Welsh content and an unpublished English version:
image

A form with Welsh content and unpublished English changes:
image

A form with Welsh content and a live English version:
image

A draft form with no Welsh content:
image

A live form with a live Welsh version:
image

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -232,6 +232,16 @@ class Form < ApplicationRecord
     can_make_welsh_version_live? if language == "cy"
   end
 
+  def changed_from_live_version?(language:)
+    live_document = language == "cy" ? live_welsh_form_document : live_form_document
+    return false if live_document.blank?
+
+    ignored_keys = %w[live_at available_languages updated_at]
+    return false if live_document.content.except(*ignored_keys) == as_form_document(language:).except(*ignored_keys)
+
+    true
+  end
+
 private
 
   def set_external_id
@@ -296,7 +306,15 @@ private
   end
 
   def can_make_welsh_version_live?
-    has_live_version && all_ready_for_live? && welsh_completed? && live_form_document.present? && live_welsh_form_document.blank?
+    english_version_has_been_made_live? && !changed_from_live_version?(language: "en") && welsh_version_ready? && live_welsh_form_document.blank?
+  end
+
+  def english_version_has_been_made_live?
+    has_live_version && live_form_document.present?
+  end
+
+  def welsh_version_ready?
+    all_ready_for_live? && welsh_completed?
   end
 
   def after_archive
--- a/app/services/form_task_list_service.rb
+++ b/app/services/form_task_list_service.rb
@@ -224,9 +224,12 @@ private
   def make_only_english_live_task
     status = @task_statuses[:make_only_english_live_status]
     can_make_form_live = status == :not_started
+    english_changes_to_make_live = @form.is_live? && @form.changed_from_live_version?(language: "en")
+
+    task_name = english_changes_to_make_live ? I18n.t("forms.task_list_edit.make_form_live_section.make_english_form_live") : I18n.t("forms.task_list_create.make_form_live_section.make_english_form_live")
 
     {
-      task_name: I18n.t("forms.task_list_create.make_form_live_section.make_english_form_live"),
+      task_name:,
       path: can_make_form_live ? make_language_live_path(@form.id, language: "en") : "",
       status:,
       active: can_make_form_live,
--- a/app/services/form_task_list_service.rb
+++ b/app/services/form_task_list_service.rb
@@ -193,14 +193,20 @@ private
 
   def make_form_live_section_tasks
     [
-      {
-        task_name: share_preview_task_name,
-        path: share_preview_path(@form.id),
-        status: @task_statuses[:share_preview_status],
-        active: @form.pages.any?,
-      },
-      make_live_task,
-    ]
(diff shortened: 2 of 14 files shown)


app/services/form_task_list_service.rb
```diff
@@ -215,6 +221,34 @@ def make_live_task
     }
   end
 
+  def make_only_english_live_task
+    status = @task_statuses[:make_live_status]
```
stephencdaly: Should this be:

```suggestion
    status = @task_statuses[:make_only_english_live_status]
```

app/models/form.rb
```diff
@@ -296,7 +296,15 @@ def can_make_english_version_live?
   end
 
   def can_make_welsh_version_live?
-    has_live_version && all_ready_for_live? && welsh_completed? && live_form_document.present? && live_welsh_form_document.blank?
+    english_version_has_been_made_live? && welsh_version_ready? && live_welsh_form_document.blank?
+  end
+
+  def english_version_has_been_made_live?
+    !has_draft_version && has_live_version && live_form_document.present?
```
stephencdaly: If we make the English version live, but have pending Welsh translations, it seems the form status transitions to "live" and the button text is "Create a draft to edit" even though I have some pending changes to the Welsh translation in my form.

If I then finish adding my Welsh, I need to make the English form live again before I can make the Welsh for live. I'm not sure how else it should work, but does this seem a bit confusing?

DavidBiddle: I think I've managed to rework this so that now:

- the form will stay in `live_with_draft` if there are Welsh translations in the draft
- the task statuses now check whether the English form has changed from the live form document

stephencdaly: Nice, I think this works more how I would expect

stephencdaly (dismissed):
It's comprehensively tested, and works for me locally

DavidBiddle:
@StephenGill please could you review the content for this journey? You should be able to see it working on this form on the review apps: https://pr-2810.admin.review.forms.service.gov.uk/forms/5

StephenGill:
> @StephenGill please could you review the content for this journey? You should be able to see it working on this form on the review apps: https://pr-2810.admin.review.forms.service.gov.uk/forms/5

This is my first time looking at this feature , so I struggled to find the relevant part of the Mural to compare it with. And I couldn't access the relevant screens on the review app - I must have selected the wrong user profile, and it won't let me log in using a different profile (despite clearing my cookies).

But based on the screenshots in the PR - the link text used in screenshot 1 is 'Make your English form live' - rather than 'Make the changes to your English form live', which would be more consistent with the other screenshots. This might well have been a conscious choice, in which case it's fine - just mentioning it in case it's a mistake.

DavidBiddle:
@StephenGill not a conscious choice on my part - I've been going off this bit of the Mural board: https://app.mural.co/t/gaap0347/m/gaap0347/1757516003408/d58a0d5502b18b5eea0c9cc782b2d014a7bccfcb?wid=0-1764321012841

The Mural board doesn't really account (as far as I can tell) for making changes, just the initial make live journey - I think elsewhere we use 'Make your form live' the first time a form goes live and 'Make your changes live' on subsequent changes. Do you reckon we should add extra content here to make that consistent before merging this?

StephenGill:
@DavidBiddle I can see how people might read 'Make your form live' as suggesting that the form isn't already live - so might cause confusion. So personally I think it's worth doing, provided it doesn't hold up shipping significantly.

So if there's no Welsh version, would be 'Make your form live' / 'Make changes to your form live'.

Otherwise would be 'Make changes to your English form live' or 'Make changes to your Welsh form live'.

chao-xian (dismissed):

app/services/form_task_list_service.rb
```diff
@@ -215,6 +221,35 @@ def make_live_task
     }
   end
 
+  def make_only_english_live_task
+    status = @task_statuses[:make_only_english_live_status]
+    can_make_form_live = status == :not_started
+    task_name = @form.is_live? && can_make_form_live ? I18n.t("forms.task_list_edit.make_form_live_section.make_english_form_live") : I18n.t("forms.task_list_create.make_form_live_section.make_english_form_live")
```
stephencdaly:
```suggestion
    task_name = @form.is_live? ? I18n.t("forms.task_list_edit.make_form_live_section.make_english_form_live") : I18n.t("forms.task_list_create.make_form_live_section.make_english_form_live")
```

Even if this task is disabled, should the content just be dependent on whether the English version is already live or not?

DavidBiddle: This came out of the content review - we think it's potentially a bit confusing in some cases if it says `Make changes to your English form live - Completed` when you haven't actually made any English changes.

So instead we're going to show `Make your English form live - Completed` if the form has no English changes to make live, and `Make changes to your English form live - Not started` if there are changes that can be made live.
 
I'll update this bit of the code to make it more explicit what's going on here.

stephencdaly: Ah sorry that makes sense

stephencdaly (approved):

PR #2811, opened 2026-05-27
Show routes welsh live archived pages

SamJamCul:

https://trello.com/c/LtIlAUBk/3072-show-new-routes-in-live-archived-view

For groups with multiple branches enabled, forms that have a live/archived Welsh version will show new table format. 

--- 

image

---

Additionally fixed some issues with how we were displaying `None of the above` in these tables. In fact, that page just didn't work!

**Before**

image

**After**

image

### DIFF
--- a/app/services/step_summary_table_service.rb
+++ b/app/services/step_summary_table_service.rb
@@ -24,6 +24,19 @@ class StepSummaryTableService
     rows
   end
 
+  def values_with_welsh_content2
+    rows = []
+
+    rows.push(page_heading_row) if @step.respond_to?(:page_heading) && @step.page_heading.present?
+    rows.push(guidance_markdown_row) if @step.respond_to?(:guidance_markdown) && @step.guidance_markdown.present?
+    rows.push(question_text_row) if @step.page_heading.present?
+    rows.push(hint_row) if @step.hint_text.present?
+    rows.concat(selection_rows) if @step.answer_type == "selection"
+    rows.concat(route_rows) if @step.routing_conditions.present?
+
+    rows
+  end
+
   def untranslated_content
     return text_row if @step.answer_type == "text"
     return date_row if @step.answer_type == "date"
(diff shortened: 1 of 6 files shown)


app/services/step_summary_table_service.rb
```diff
@@ -293,6 +419,16 @@ def print_check_page(condition, steps)
   def welsh_answer_value(welsh_condition)
     return nil if welsh_condition.answer_value.blank?
 
+    return I18n.t("step_summary_card.selection_type.none_of_the_above.cy") if welsh_condition.answer_value == "none_of_the_above"
```
lfdebrux: Ideally we would use the `NONE_OF_THE_ABOVE` constant for this:


```suggestion
    return I18n.t("step_summary_card.selection_type.none_of_the_above.cy") if welsh_condition.answer_value == Condition::NONE_OF_THE_ABOVE
```

app/services/step_summary_table_service.rb
```diff
@@ -293,6 +419,16 @@ def print_check_page(condition, steps)
...
+
     @welsh_steps.find { |step| step.id == welsh_condition.check_page_id }.answer_settings.selection_options.find { |option| option.value == welsh_condition.answer_value }.name
   end
+
+  def welsh_answer_value2(welsh_condition)
+    return nil if welsh_condition.answer_value.blank?
+
+    return I18n.t("step_summary_card.selection_type.none_of_the_above.cy") if welsh_condition.answer_value == "none_of_the_above"
```
lfdebrux:
```suggestion
    return I18n.t("step_summary_card.selection_type.none_of_the_above.cy") if welsh_condition.answer_value == Condition::NONE_OF_THE_ABOVE
```

app/services/step_summary_table_service.rb
SamJamCul: oh yes, should have remembered - fixing it

christophercameron-ixd:
It looks right to me - from a design perspective :)

lfdebrux (approved):

PR #2812, opened 2026-05-28
Fix row order for Welsh form CSV

SamJamCul:
https://trello.com/c/JYUtlVzE/3117-content-for-guidance-questions-not-in-same-order-on-welsh-csv-as-on-add-welsh-content-page

The correct order for a question's details in a CSV is
 - Page heading
 - Guidance text
 - Question text
 - Hint text (if relevant)
 - List of options (if relevant)

This matches how the translations are displayed on the Welsh Translations page

The test for this has switched from using `match_array` to `eq` so that the ordering is taken into account.

### DIFF
--- a/app/services/welsh_csv_service.rb
+++ b/app/services/welsh_csv_service.rb
@@ -38,12 +38,12 @@ private
 
   def add_page_content(csv)
     form.pages.each do |page|
+      add_page_heading(csv, page)
+      add_guidance_text(csv, page)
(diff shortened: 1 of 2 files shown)


lfdebrux (approved):

lfdebrux:
Slick! Thanks for figuring that out.

hannahkc:
looks good to me!

PR #2821, opened 2026-06-04
Publish form count KPI metrics to CloudWatch

theseanything:

Trello card: https://trello.com/c/b3cNvzu5/3136-record-kpis-in-cloudwatch

This adds automated form count metrics to CloudWatch so we can track how the platform is growing over time and measure the impact of our work. A new service publishes counts grouped by organisation and state (live, archived, draft), and a metrics:export_form_counts rake task will run on a scheduled cron job to publish those metrics on a regular basis. This lays the foundation for KPI reporting without manual updates.

### DIFF
--- a/app/services/metrics/form_count_service.rb
+++ b/app/services/metrics/form_count_service.rb
@@ -1,0 +1,95 @@
+module Metrics
+  class FormCountService
+    class ExportError < StandardError; end
+
+    METRIC_NAME = "FormCount".freeze
+    METER_NAME = "forms-admin".freeze
+    METER_VERSION = "1.0".freeze
+    UNKNOWN_ORG = "Unknown".freeze
+    METRIC_STATES = %w[draft live archived].freeze
+
+    def publish_form_counts
+      metric_count = 0
+
+      form_counts_by_org_and_state.each do |(org, state), count|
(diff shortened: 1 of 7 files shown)


app/services/metrics/form_count_service.rb
```diff
@@ -0,0 +1,74 @@
...
+      Sentry.capture_exception(e)
+      raise
+    end
+
+  private
+
+    def form_counts_by_org_and_state
+      counts_by_org_and_state = Form
+        .where.not(state: :deleted)
+        .left_joins(group_form: { group: :organisation })
```
lfdebrux, line 39: In the forms-admin database we have a column `internal` on the organisation table that we use to mark some organisations as being excluded from reports and analytics; see https://github.com/govuk-forms/forms-admin/commit/f40573790b869bd674622ee751fa9ea8f8e4a51a.

Could we use the same here?

theseanything: We can do the filtering at monitoring tooling level - let's just get all our metrics there.

theseanything: I've gone ahead and filtered out the internal orgs here

lfdebrux (dismissed):

theseanything:
Need to re-think the best place to store these metrics.

PR #2823, opened 2026-06-05
Add NormaliseConditionsService to delete conditions that go to the next page

lfdebrux:

Trello card: https://trello.com/c/MWY6jwTQ/3113-add-code-to-delete-conditions-that-go-to-the-next-page 

When the multiple branches feature is enabled, we want to stop reporting conditions that go to the next page as a routing error to the user, and instead silently delete those conditions for them.

This PR adds a new `NormaliseConditionsService` to delete those conditions, and uses it to delete conditions to route to the next page when the form creator finishes editing the questions for their forms.

It also fixes a bug in the edit routes page that occurs when a condition is set to route to the next page.

### DIFF
--- a/spec/views/routes/show.html.erb_spec.rb
+++ b/spec/views/routes/show.html.erb_spec.rb
@@ -111,6 +111,37 @@ describe "routes/show.html.erb" do
       end
     end
 
+    context "when the route goes to the next page" do
+      let(:pages) do
+        [
+          build_stubbed(
+            :page,
+            id: 101,
+            routing_conditions: [
+              build_stubbed(
+                :condition,
+                routing_page_id: 101,
+                goto_page_id: 102,
+                answer_value: nil,
+              ),
+            ],
+          ),
+          build_stubbed(:page, id: 102),
+          build_stubbed(:page, id: 103),
+        ]
+      end
+
+      it "shows the selected goto page for the route" do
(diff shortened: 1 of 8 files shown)


spec/views/routes/show.html.erb_spec.rb
```diff
@@ -111,6 +111,38 @@ def select_options(select_field)
...
+          build_stubbed(:page, id: 102),
+          build_stubbed(:page, id: 103),
+        ]
+      end
+
+      it "shows the selected goto page for the route" do
+        render_page
+
+        expect(rendered).to have_css('.govuk-select[name="forms_routes_input[routes_attributes][0][goto]"]') do |field|
+          expect(field).to have_css("option", count: 5)
```
thomasiles: There is a failing test for this PR and I think this count should always be 3, not 5 - default (page 102), page 103, end of form?

thomasiles:
Looks good to me apart from the one test.

spec/views/routes/show.html.erb_spec.rb
lfdebrux: Oh yeah, no idea where I got 5 from... That's annoying, I thought I was writing code to make that test pass, turns out it was always broken 😅

thomasiles (approved):

PR #2825, opened 2026-06-08
Change all goto page inputs on edit routes page to be the same width

lfdebrux:

The width of a select input depends on the longest option in the input; on the edit routes page which has lots of them each with different options the last select input on the page can end up much less wide than the others, which is visually distracting.

This commit fixes that issue by using a width override to make each input take up the full available space of the grid column. This doesn't take up the full width of the viewport, because of the grid, but is still a bit wider than strictly necessary; however all the other width overrides are not wide enough for a question of reasonable length.

Thanks @lorna-howes for spotting the issue!

#### Before

Screenshot of the edit routes page before this change, showing multiple select inputs with the last being much less wide than the ones before

#### After

Screenshot of the edit routes page after this change, showing multiple select inputs all the same width

### DIFF
--- a/app/views/routes/show.html.erb
+++ b/app/views/routes/show.html.erb
@@ -33,7 +33,13 @@
                           route_form.hidden_field(:id) +
                             route_form.hidden_field(:page_id) +
                             route_form.hidden_field(:answer_value) +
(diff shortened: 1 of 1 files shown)


thomasiles (dismissed):
nice!

thomasiles (approved):

PR #2828, opened 2026-06-09
Fix stale_conditions same answer value conditions

SamJamCul:
When creating a new set of routes for a form using the Routes::SyncService, a bug occurs when two conditions for different questions share the same answer value. For example, two selection option questions might both have Yes and No as their selection options, so conditions based on either of those questions would have identical-seeming answer values.

The check to see whether a condition can be deleted because it is stale (i.e. it points to the next question, so doesn't need to exist as a condition) was using the answer values from every condition for the form to identify stale conditions by routing_page_id and answer_value.

This meant that conditions which had a routing_page_id matching a route in the list of default_routes, and an answer_value matching an answer_value in any other route in the entire form's default_routes would be considered to be stale, and destroyed.

This change makes it so that stale conditions are identified using the list of default_routes, but only using the answer_value based on the specific default route's, rather than across every default route.

### DIFF
--- a/app/services/routes/sync_service.rb
+++ b/app/services/routes/sync_service.rb
@@ -30,6 +30,7 @@ private
       condition.assign_attributes(
         route.condition_attributes,
       )
+
       condition.save!
     end
   end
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
nice 👍

PR #2838, opened 2026-06-11
Backward route validate

thomasiles:
### Add validations to the routes page for backwards routes

Trello card: https://trello.com/c/mwYaCg6h/3137-show-errors-for-backwards-routes-when-multiple-branches-are-enabled

This PR adds backward route validation to the multiple branches routes and page list.

In the routes code, we use new validation. In the page list, we use the Conditions. I think in the long term, the validation should all happen in the routes input which should be used in the page list too. While the feature is still in development, I think using the condition code for now makes sense.

## The routes view:

image

## The page list view:

image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -365,6 +365,9 @@ en:
       goto_page_doesnt_exist: The question you’re taking the person to no longer exists - edit question %{question_number}’s route
       route_number_for_any_other_answer: for any other answer
     prefix: 'Error:'
+    routes:
+      cannot_route_backwards_from_generic_page: The route from question %{question_number} cannot go to a previous question - edit this route
+      cannot_route_backwards_from_selection_page: The route from question %{question_number}, option %{option_number}, cannot go to a previous question - edit this route
   exit_page:
     no_content_added_html: "<p>No content added</p>"
   feedback_link:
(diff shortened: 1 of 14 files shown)


config/locales/en.yml
```diff
@@ -365,6 +365,14 @@ en:
       goto_page_doesnt_exist: The question you’re taking the person to no longer exists - edit question %{question_number}’s route
       route_number_for_any_other_answer: for any other answer
     prefix: 'Error:'
+    routes:
+      cannot_route_backwards_from_generic_page: The route from question %{question_number} cannot go to a previous question - edit this route
+      cannot_route_backwards_from_selection_page: The route from question %{question_number}, option %{option_number}, cannot go to a previous question - edit this route
```
hannahkc:
```suggestion
      cannot_route_backwards_from_selection_page: The route from question %{question_number}, option %{option_number} cannot go to a previous question - edit this route
```

hannahkc:
Seems to work correctly and content looks good except for an extra comma. thanks!

SamJamCul:
The anchoring from the error summary to the route selection is a bit funny - it anchors to the selection box itself, rather than the label for the selection. This means the label gets cut off when you jump to it. I don't know if that's intentional or not, as I'm not sure what the convention is!

chrome-capture-2026-06-12

SamJamCul (approved):
Looks good, just a question about the link anchoring for the error summary message on the routes page

thomasiles:
> The anchoring from the error summary to the route selection is a bit funny - it anchors to the selection box itself, rather than the label for the selection. This means the label gets cut off when you jump to it. I don't know if that's intentional or not, as I'm not sure what the convention is!

This is a good point. I double checked the design system page and errors do always link to an input box (or radio/checkbox). I think the idea is you don't need to read the error message near the input because you've just come straight from the summary anyway.

I think it breaks down a bit when you need to change multiple fields but I think it makes sense.

I appreciate the screen capture showing it!

PR #2839, opened 2026-06-11
Rake task for changing the state of a Form

chao-xian:

We have some forms that were fiddled with in the database to remove them from service. This was prior to having the Archive functionality we now have. As a consequence some of our metrics data is counting wrong. 

Example:

XL Bully - the "live" version was deleted, but as it was once live it should be counted as being published some time ago.

This PR adds a method and accompanying rake task for switching a Form's state programatically, in the shortest number of steps.

In the case of XL Bully it would make it live and then immediately archive it. As it's done programatically, it would be much quicker than one of the admins clicking around to do the same thing. It also decreases the likelihood of a user finding the live form.

Trello card: https://trello.com/c/44SCirtT/3150-live-and-archived-forms-report-is-missing-2-forms

### DIFF
--- a/lib/tasks/forms.rake
+++ b/lib/tasks/forms.rake
@@ -25,6 +25,35 @@ namespace :forms do
     end
   end
 
+  desc "set the state for a form by transitioning through the form state machine"
+  task :set_state, %i[form_id state] => :environment do |_, args|
+    usage_message = "usage: rake forms:set_state[<form_id>, <state>]".freeze
+    abort usage_message if args[:form_id].blank? || args[:state].blank?
+    abort "state must be one of #{Form.states.keys.join(', ')}" unless Form.states.key?(args[:state])
+
+    form = Form.find(args[:form_id])
+
+    # the make_live event guard checks the form's task statuses through a
+    # service that is normally injected by the controller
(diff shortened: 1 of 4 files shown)


lib/tasks/forms.rake
```diff
@@ -25,6 +25,30 @@ namespace :forms do
...
+
+    form = Form.find(args[:form_id])
+
+    # the make_live event guard checks the form's task statuses through a
+    # service that is normally injected by the controller
+    form.set_task_status_service(TaskStatusService.new(form:, current_user: nil))
+
+    events = Form.event_path(from: form.aasm.current_state, to: args[:state].to_sym)
+
+    abort "cannot transition form from \'#{form.state}\' to \'#{args[:state]}\'" if events.nil?
```
DavidBiddle, line 42: It might make sense to explicitly log something if `events` is empty - I think currently if the task is run on a form that's already in the target state, events will be `[]` and the task won't output anything which could be a little confusing

DavidBiddle (dismissed):
Looks good, well-organised and tested and it works on my local machine. Have raised one optional suggestion

DavidBiddle (approved):

PR #2840, opened 2026-06-12
Fix routes showing by goto page order

SamJamCul:
We were grouping conditions by goto_page_id, and sorting them by this as well. This meant that when page ids were not in numerical order, they would be sorted out of sequence relating to their position.

This change ensures that page positions are used when sorting the final condition groups.

### DIFF
--- a/app/components/page_list_component/view.rb
+++ b/app/components/page_list_component/view.rb
@@ -106,7 +106,7 @@ module PageListComponent
       page.routing_conditions.to_a
         .in_order_of(:answer_value, answer_order, filter: false)
         .group_by(&:goto_page_id)
-        .sort_by { |goto_page_id, _| goto_page_id || Float::INFINITY }
+        .sort_by { |goto_page_id, _| goto_page_id ? @pages.ids.index(goto_page_id) : Float::INFINITY }
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/components/page_list_component/view.rb
```diff
@@ -102,11 +102,14 @@ def skip_condition_route_page_text(condition)
 
     def answer_value_groups(page)
       answer_order = page.answer_settings&.selection_options&.map(&:value) || []
+      page_ids_by_position = @pages.where(id: page.routing_conditions.map(&:goto_page_id).compact)
+                              .order(:position)
+                              .ids
```
thomasiles: Could this be simplified to just `@pages.ids`? It would include all page ids for the whole form but doesn't require querying the conditions.

I'm not sure it makes any difference to performance at all so might not be worth it.

thomasiles (dismissed):
Looks good to me! I've left a comment about the page_ids_by_position array but I'm not sure it's an improvement

app/components/page_list_component/view.rb
SamJamCul: Very true! I completely forgot we'd already ordered them

SamJamCul: I've changed it, it's much simpler this way

thomasiles (approved):

PR #2851, opened 2026-06-18
Show backwars routing error for final page

SamJamCul:
When there is a backwards route on the final page of a form, the selection options for routing will be displayed on the routes page, so that it can be corrected.

https://trello.com/c/OGjKv7oo/3140-show-route-selection-drop-downs-on-the-last-question-if-it-has-any-backwards-routes

When there is backwards routing on the final page of a form (probably because there have been pages moved around) we want to show the selection options for that page on the routes page. Normally, we don't give the options for the final page, as they always lead to the end of the form. 

---

**Multiple selection backwards route**

image

---

**Unconditional backwards route**

image
 

### DIFF
--- a/app/services/routes/build_service.rb
+++ b/app/services/routes/build_service.rb
@@ -28,7 +28,18 @@ class Routes::BuildService
   def options_for_goto_page(page, selected = nil)
     next_page = form.next_page_after(page)
 
-    return [] unless next_page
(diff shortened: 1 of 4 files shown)


hannahkc:
Content looks good and works as expected. thanks!

lfdebrux (approved):

PR #2853, opened 2026-06-19
Warn delete options multiple branches

thomasiles:
### Delete conditions which reference options which don't exist and add banner to options page about routes

Trello card: https://trello.com/c/vaePaWB5/3141-delete-routes-when-changing-selection-options-makes-it-invalid-when-multiple-branches-is-enabled

This PR adds a banner to the selection options page to warn the user that options are being used in routes. The banner changes the wording based on how many options are involved in routes and whether the "none of the above" is activated for the page.

When options are changed or removed, it removes any conditions which have answer_values which no longer have matching options.

### DIFF
--- a/app/helpers/pages_helper.rb
+++ b/app/helpers/pages_helper.rb
@@ -23,4 +23,37 @@ module PagesHelper
   def draft_question_selection_options(draft_question)
     draft_question.answer_settings[:selection_options]
   end
+
+  # option_indexes is either :number or :answer_value
+  def selection_options_in_routes_banner(draft_question, selection_options, include_none_of_the_above, option_indexes: :number)
+    return unless draft_question.form&.group&.multiple_branches_enabled?
+
+    answer_values_from_options = selection_options.pluck(:value) + [Condition::NONE_OF_THE_ABOVE]
+    options_in_routes = draft_question.form.conditions.where(routing_page_id: draft_question.page_id, answer_value: answer_values_from_options).select(:answer_value)
+
+    return if options_in_routes.empty?
+
+    edit_routes_link = govuk_link_to("View your question routes", routes_path(draft_question.form_id))
+
+    if options_in_routes.count == 1 && options_in_routes.first.answer_value == Condition::NONE_OF_THE_ABOVE
+      return { heading: "There is a route from ‘None of the above’", text: "If you remove ‘None of the above’, the route will be deleted. #{edit_routes_link}" }
+    end
+
+    if options_in_routes.count == 1 && option_indexes == :number
+      option_index = selection_options.index { |option| option[:value] == options_in_routes.first.answer_value } + 1
+      return { heading: "There is a route from option #{option_index}", text: "If you remove or change option #{option_index}, the route will be deleted. #{edit_routes_link}" }
+    end
+
+    if options_in_routes.count == 1 && option_indexes == :answer_value
+      option_index = options_in_routes.first.answer_value
+      return { heading: "There is a route from ‘#{option_index}’", text: "If you remove or change this option, the route will be deleted. #{edit_routes_link}" }
+    end
+
+    total_selection_options_count = selection_options.length + (include_none_of_the_above ? 1 : 0)
+    if options_in_routes.count == total_selection_options_count
+      return { heading: "There are routes from these options", text: "If you remove or change an option with a route, the route will be deleted. #{edit_routes_link}" }
+    end
+
+    { heading: "There are routes from some of these options", text: "If you remove or change an option with a route, the route will be deleted. #{edit_routes_link}" }
+  end
 end
--- a/app/input_objects/pages/question_input.rb
+++ b/app/input_objects/pages/question_input.rb
@@ -61,7 +61,10 @@ class Pages::QuestionInput < BaseInput
 
     page.assign_attributes(**attrs)
 
-    page.save_and_update_form
+    ActiveRecord::Base.transaction do
+      remove_conditions_without_valid_answer_values(page) if draft_question.form&.group&.multiple_branches_enabled?
+      page.save_and_update_form
+    end
   end
 
   def update_draft_question!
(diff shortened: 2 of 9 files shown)


app/input_objects/pages/question_input.rb
```diff
@@ -133,4 +136,16 @@ def answer_settings_cy(page = nil)
 
     answer_settings_cloned
   end
+
+  def remove_conditions_without_valid_answer_values(page)
+    return unless answer_type == "selection"
+
+    valid_answer_values = answer_settings[:selection_options].map { it[:name] }
```
lfdebrux: Shouldn't we be using the `:value` rather than the `:name` here?

app/input_objects/pages/question_input.rb
```diff
@@ -133,4 +136,16 @@ def answer_settings_cy(page = nil)
...
+
+    if answer_settings[:none_of_the_above_question].present?
```
lfdebrux: What if the selection option allows "none of the above" as an answer but doesn't ask for a different answer?

app/helpers/pages_helper.rb
```diff
@@ -23,4 +23,34 @@ def has_more_than_30_options?(draft_question)
   def draft_question_selection_options(draft_question)
     draft_question.answer_settings[:selection_options]
   end
+
+  # option_indexes is either :number or :answer_value
+  def selection_options_in_routes_banner(draft_question, selection_options, include_none_of_the_above, option_indexes: :number)
```
lfdebrux, line 28: I wonder if it would be easier to use this method if if was on the input object rather than a standalone helper... did you consider that?

app/input_objects/pages/question_input.rb
thomasiles: Yes! good point

app/input_objects/pages/question_input.rb
thomasiles: yes, I don't know how I didn't trigger that when testing. I've also added a guard clause to only change it when it's being treated as a selection option as the other cases are handled elsewhere or in other tickets.

app/helpers/pages_helper.rb
thomasiles: I did consider this and it's how I started off. I didn't want to bloat the input object with something so dedicated to the presentation. I'm not super keen on helper methods but it seemed the right place.

I'll happily change it back to the input object if you think thats a better place for it. It would make it easier to call and possibly easier to test too.

lfdebrux: Yeah fair. Maybe a presenter would be in order? This seems fine for now, but something to consider.

app/helpers/pages_helper.rb
```diff
@@ -23,4 +23,34 @@ def has_more_than_30_options?(draft_question)
...
+    answer_values_from_options = selection_options.pluck(:name) + [Condition::NONE_OF_THE_ABOVE]
```
lfdebrux:
```suggestion
    answer_values_from_options = selection_options.pluck(:value) + [Condition::NONE_OF_THE_ABOVE]
```

app/helpers/pages_helper.rb
```diff
@@ -23,4 +23,34 @@ def has_more_than_30_options?(draft_question)
...
+    return if options_in_routes.empty?
+
+    edit_routes_link = govuk_link_to("View your question routes", routes_path(draft_question.form_id))
+
+    if options_in_routes.count == 1 && options_in_routes.first.answer_value == Condition::NONE_OF_THE_ABOVE
+      return { heading: "There is a route from ‘None of the above’", text: "If you remove ‘None of the above’, the route will be deleted. #{edit_routes_link}" }
+    end
+
+    if options_in_routes.count == 1 && option_indexes == :number
+      option_index = selection_options.index { |option| option[:name] == options_in_routes.first.answer_value } + 1
```
lfdebrux:
```suggestion
      option_index = selection_options.index { |option| option[:value] == options_in_routes.first.answer_value } + 1
```

app/helpers/pages_helper.rb
```diff
@@ -23,4 +23,34 @@ def has_more_than_30_options?(draft_question)
...
+      return { heading: "There is a route from option #{option_index}", text: "If you remove or change option #{option_index}, the route will be deleted. #{edit_routes_link}" }
+    end
+
+    if options_in_routes.count == 1 && option_indexes == :answer_value
+      option_index = options_in_routes.first.answer_value
+      return { heading: "There is a route from ‘#{option_index}’", text: "If you remove or change this option, the route will be deleted. #{edit_routes_link}" }
+    end
+
+    if selection_options.length == options_in_routes.count + (include_none_of_the_above ? 1 : 0)
```
lfdebrux: I couldn't get the "There are routes fro these options" banner to show with a question which doesn't include a 'none of the above' option, I think this line might need looking at again?

spec/helpers/pages_helper_spec.rb
```diff
@@ -87,4 +87,83 @@
       end
     end
   end
+
+  describe "#selection_options_in_routes_banner" do
+    let(:form) { create :form, :ready_for_routing }
+    let(:draft_question) { build :selection_draft_question, form_id: form.id, page_id: form.pages.first.id }
+    let(:selection_options) { (1..3).to_a.map { |i| { name: i.to_s } } }
+    let(:include_none_of_the_above) { false }
```
lfdebrux, line 96: `Pages::Selection::BaseOptionsInput#selected_none_of_the_above_option` returns a string, not a boolean.

app/helpers/pages_helper.rb
thomasiles: 👍 I've changed this to use the selection_options_with_value method and the :value attribute rather than :name

app/helpers/pages_helper.rb
thomasiles: Yes, I'd mucked up the order of this. I've fixed it and given it a name to hopefully make it clearer.

spec/helpers/pages_helper_spec.rb
thomasiles: good catch, I've changed the value being passed in to is_optional, which is a boolean.

thomasiles:
I've made some changes and checked everything works as designed.

I did look at moving the helper into the input object because they depend on input object methods. It didn't come out much clearer though because it still depended on path helpers and code to generate the link. It might be better as a presenter/service but I'm leaving it here as I'm not confident I'm making things much better by more fiddling.

lfdebrux (approved):

PR #2867, opened 2026-06-25
Ensure conditions match question type

thomasiles:
### Make conditions match the question type, if changed

Trello card: https://trello.com/c/lDyxaMH5/3145-if-a-question-with-an-unconditional-route-is-changed-to-a-select-one-list-question-set-all-the-options-to-go-to-the-same-target

When a question is changed to selection type which supports routing from multiple options, we need to make sure the conditions are updated to match.

This commit adds a new method which is called when a question is changed. It updates conditions to based on the selection options for the new question.

### DIFF
--- a/app/views/pages/type_of_answer.html.erb
+++ b/app/views/pages/type_of_answer.html.erb
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% if @page.present? && @page.routing_conditions.any?  %>
+    <% if !current_form&.group&.multiple_branches_enabled? && @page.present? && @page.routing_conditions.any?  %>
       <%= govuk_notification_banner(title_text: t("banner.default.title")) do |banner| %>
         <% banner.with_heading(text: t("type_of_answer.routing_warning_about_change_answer_type_heading"), tag: "h3") %>
         <%= t("type_of_answer.routing_warning_about_change_answer_type_html", pages_link_url: form_pages_path(current_form.id)) %>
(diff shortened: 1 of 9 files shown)


app/views/pages/type_of_answer.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% if @page.present? && @page.routing_conditions.any?  %>
+    <% if !current_form&.group&.multiple_branches_enabled? && @page.present? && @page.routing_conditions.any?  %>
```
lfdebrux: If we use the `FeatureService`  then we can simplify the tests:

```suggestion
    <% if !FeatureService.new(group: current_form&.group).enabled?(:multiple_branches) && @page.present? && @page.routing_conditions.any?  %>
```

spec/views/pages/type_of_answer.html.erb_spec.rb
```diff
@@ -77,6 +77,15 @@
                         .text(normalize_ws: true))
     end
 
+    context "when multiple branches are enabled" do
+      let(:form) { create(:form, :with_group, group:) }
+      let(:group) { create(:group, multiple_branches_enabled: true) }
+
```
lfdebrux: Simpler tests when using `FeatureService`:

```suggestion
    context "when multiple branches are enabled", :feature_multiple_branches do
```

app/views/pages/type_of_answer.html.erb
thomasiles: good point!

hannahkc:
This works perfectly. Thanks Tom!

lfdebrux (approved):

PR #2879, opened 2026-07-01
Update banner notifications when changing a select-one question with routes from answers

SamJamCul:

Trello card: https://trello.com/c/P4dayADV/3143-update-banner-notifications-when-changing-a-select-one-question-with-routes-from-answers

Updates the banner that shows when changing the answer type from a selection question, with only one option. The banner no longer shows if it's a selection question with multiple options. 

The banner says `route` or `routes` depending on the number of routes from that question. 

image

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -54,7 +54,7 @@ class Page < ApplicationRecord
 
     save!
     form.save_question_changes!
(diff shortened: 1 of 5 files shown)


lfdebrux (approved):

william-castoncook:
Checked content  (plural and singular) and it looks good!

PR #2880, opened 2026-07-01
Add exit pages

thomasiles:
### Add new ExitPage table and model

Trello card: https://trello.com/c/E1qseSQu/3163-add-model-for-exit-pages-to-forms-admin

This PR adds a new model for ExitPages. The model isn't used anywhere yet. The `"exit_pages": []` entry will appear in formdocument's steps. 

### DIFF
--- a/app/models/page.rb
+++ b/app/models/page.rb
@@ -6,6 +6,8 @@ class Page < ApplicationRecord
   has_many :routing_conditions, class_name: "Condition", foreign_key: "routing_page_id", dependent: :destroy
   has_many :check_conditions, class_name: "Condition", foreign_key: "check_page_id", dependent: :destroy
   has_many :goto_conditions, class_name: "Condition", foreign_key: "goto_page_id", dependent: :destroy
+  has_many :exit_pages, foreign_key: :question_page_id, dependent: :destroy
+
   acts_as_list scope: :form
 
   translates :question_text,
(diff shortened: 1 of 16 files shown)


app/models/page.rb
```diff
@@ -54,10 +54,15 @@ def destroy_and_update_form!
...
-    check_conditions.destroy_all if answer_type_changed_from_selection
-    check_conditions.destroy_all if answer_settings_changed_from_only_one_option
+    ActiveRecord::Base.transaction do
+      save!
+      form.save_question_changes!
+
+      if answer_type_changed_from_selection || answer_settings_changed_from_only_one_option
+        exit_pages.destroy_all
+        check_conditions.destroy_all
+      end
```
SamJamCul: I think my PR mangled the `answer_type_changed_from_selection` into something that tried to detect both the answer type and settings being changed. I think I prefer your version though, it's more readable!

SamJamCul (dismissed):

app/models/page.rb
lfdebrux: I think we do need your changes though @SamJamCul because if the answer type is changed from one or more options selection question we shouldn't delete the conditions... 🤔

SamJamCul: Yeah we need something there, I just made the method very unwieldy.

lfdebrux (approved):

PR #2881, opened 2026-07-02
Allow all Welsh government, Scottish government, and MOD subdomains in email validation

DavidBiddle:

Trello card: Mitigates https://trello.com/c/NQj6zljA/3449-make-email-domain-validation-check-against-allowed-domains-rather-than-just-govuk-and-the-users-current-domain

In our domain allowlist, we include a number of non .gov.uk email domains. Most of these match the entire domain - e.g. `@hs2.org.uk` - but a small number allow any subdomains - e.g. `.gov.scot`.

In this validator we allow any `.gov.uk `email address, as well as any domain that matches the current user's domain, which we implemented to avoid duplicating the entire allowlist in this app. However, this sometimes causes issues when editors from a parent organisation attempt to use an email address from one of their suborganisations.

This commit explicitly allows any subdomain of the `.gov.scot`,`.gov.wales`, and `.mod.uk` domains on the logic that:
- they are likely to account for a lot of the parent/child organisation mismatches that cause validation issues
- they allow us to significantly increase the number of valid emails without having to duplicate the whole allowlist here.

This should mitigate the domain mismatch validation issue until we have a permanent fix in place.

### DIFF
--- a/app/validators/allowed_email_domain_validator.rb
+++ b/app/validators/allowed_email_domain_validator.rb
@@ -1,7 +1,13 @@
 class AllowedEmailDomainValidator < ActiveModel::EachValidator
+  ALLOWED_SUBDOMAIN_REGEXES = [/\.gov\.uk\z/i,
+                               /\.gov\.scot\z/i,
+                               /\.gov\.wales\z/i,
+                               /\.mod\.uk\z/i].freeze
+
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):
Seems sensible

PR #2882, opened 2026-07-02
Add organisation_domains table

stephencdaly:

Trello card: https://trello.com/c/lIDJ2j9T/3176-store-email-domains-per-organisation-in-forms-admin

Add `organisation_domains` table so we can store the email domains associated with each organisation. Add rake tasks to add/remove/list domains for an organisation.


DavidBiddle (approved):
Code looks good, nicely organised, and the rake tasks work on my machine :tada:

PR #2885, opened 2026-07-03
Add rake task to populate domains from existing users

stephencdaly:

Trello card: https://trello.com/c/CELIu9oE

This rake tasks looks at existing users per organisation and adds an OrganisationDomain per unique email domain. This is intended to be run as a one-off task to populate the OrganisationDomain table with existing data. We'll review this data after we've run the task and remove any domains we don't think should be associated with organisations.


DavidBiddle (approved):
Looks good, and the rake task works for my local users 🎉

PR #2886, opened 2026-07-06
Add exit pages multiple branches

thomasiles:
### WIP: Use ExitPage model when available in Condition model

Trello card: https://trello.com/c/gZGRBU8Y/3172-change-add-edit-delete-exit-page-to-work-with-new-style-exit-page-model

First parts of using ExitPage model instead of condition attributes to model exit pages.


thomasiles:
This work was completed in another branch

PR #2892, opened 2026-07-07
Add smoke test form to database seed

lfdebrux:

In each environment in AWS we have a smoke test form which is used by the tests in the forms-e2e-tests repo. This commit adds a form with the same setup as what is in the production environment to our database seed. We also add a smoke tests group to the database seed, again matching what we have in production.

Note that the forms-e2e-tests repo can't use the "All question types form" for the smoke tests, as it currently doesn't know how to answer all question types.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -633,11 +633,11 @@ if (HostingEnvironment.local_development? || HostingEnvironment.review?) && User
   copy_of_answers_form.make_live!
 
   # add forms to groups
-  GroupForm.create! group: end_to_end_group, form_id: all_question_types_form.id # All question types form
-  GroupForm.create! group: end_to_end_group, form_id: e2e_s3_forms.id # s3 submission test form
-  GroupForm.create! group: test_group, form_id: branch_route_form.id # Branch routing form
-  GroupForm.create! group: test_group, form_id: none_of_the_above_form.id # None of the above form
-  GroupForm.create! group: test_group, form_id: welsh_form.id # Welsh form
+  GroupForm.create! group: end_to_end_group, form_id: all_question_types_form.id
+  GroupForm.create! group: end_to_end_group, form_id: e2e_s3_forms.id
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -145,6 +146,46 @@
 
   submission_email = ENV["EMAIL"].presence || `git config --get user.email`.strip.presence || "example@example.com"
 
+  smoke_test_form = Form.create!(
```
stephencdaly, line 150: We'll now need to add a `DeliveryConfiguration` to the form so that it will send submissions via email. I think you need to rebase this branch to get the latest code with this added to the other seeds.

lfdebrux: Ah, thanks, I'll do that

lfdebrux: I think I've added the correct `DeliveryConfiguration`, does it look right to you?

stephencdaly (dismissed):

db/seeds.rb
```diff
@@ -667,8 +681,9 @@
   copy_of_answers_form.make_live!
 
   # add forms to groups
-  GroupForm.create! group: end_to_end_group, form_id: all_question_types_form.id # All question types form
-  GroupForm.create! group: end_to_end_group, form_id: e2e_s3_forms.id # s3 submission test form
+  GroupForm.create! group: smoke_test_group, form_id: smoke_test_form.id # s3 submission test form
```
stephencdaly: Is this comment wrong? do we need these comments as the variable names say which form it is?

db/seeds.rb
stephencdaly: Yep

db/seeds.rb
lfdebrux: Oh woops yes it is wrong. I also agree we don't need the comments, I'll remove them.

lfdebrux: Fixed, thank you for catching that!

stephencdaly (approved):

PR #2893, opened 2026-07-07
Make rds connections able to do verified ssl

whi-tw:

We want to verify the RDS server's identity when connecting to it (so that we can enforce SSL).

~We have chosen to have a copy of the cert bundle in the repo (instead of downloading) so that we have more control over it.~

The file name is the default for postgres https://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES

Trello card: https://trello.com/c/Iy8wEgFh/811-enable-tls-for-rds-database-connections

This is only adding files to use later. The real change will happen when we merge https://github.com/govuk-forms/forms-deploy/pull/2190 and re-deploy the apps

We've tested in dev


Dockerfile
```diff
@@ -71,6 +71,10 @@ RUN chmod 0755 bin/*
 COPY --chown=ruby:ruby --from=build /usr/local/bundle /usr/local/bundle
 COPY --chown=ruby:ruby --from=build /app /app
 
+# Add the RDS CA certificate to the container so that we can verify the RDS server's identity when connecting to it.
+# Cert bundle from https://truststore.pki.rds.amazonaws.com/eu-west-2/eu-west-2-bundle.pem (https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.SSL.html#UsingWithRDS.SSL.CertificatesDownload)
+COPY --chown=ruby:ruby ./docker/eu-west-2-bundle.pem /home/ruby/.postgresql/root.crt
```
theseanything: Any reason we don't just pull that from AWS directly e.g. like what GOV.UK do

https://github.com/alphagov/govuk-ruby-images/blob/e65377c3aaa675585d8c7c0d2f711d32a880cd32/base.Dockerfile#L101

cadmiumcat: That was going to be our approach initially. We thought they're likely to change very infrequently, so we decided to not have to trust and download every time.

In doing this we are trusting that AWS will notify us if we need to change it, which could definitely be a mistake. And a maintenance burden

Happy to change

theseanything: I think risk here is low, we have to trust "s3.amazonaws.com" for a lot we do anyways - if someone could compromise the s3 domain or cert authority, we've got bigger problems. Also if AWS is down there's likely to be other things that break in the deployment pipeline.

Think there's greater risk of the CA silently going stale (causing an outage), and trusting that a human pulls the CA from a trusted source.

whi-tw: We're doing cert pinning for security reasons. If we're just blindly pulling the bundle down at build time, we completely lose any control over the certificates we're trusting.

We pin all dependency versions, and have version numbers committed to the repository, along with their respective SHAs. It's not possible to do that for this certificate, so we have to do the next best thing: using a specific file that's committed to the repository.

If Amazon change / invalidate the certificates they're using for RDS, there will be people with *far* bigger issues than us. This will be announced.

As a compromise, I'm willing to create a cron action to regularly check for updates to the certificate and create a PR if necessary. But I'm not happy with pulling it down dynamically at build time.

theseanything: I disagree - but I think the risks are all round are low here. That's its not worth blocking this.

theseanything (approved):

PR #2903, opened 2026-07-08
Add custom brand task list item for a form

theseanything:
This adds an optional "Brand" task so creators can associate their form with a brand, which forms-runner then uses to style the live form. A form with no brand continues to use the default GOV.UK branding, so existing forms are unaffected.

This adds a "brand_id" attribute to the form document, which can be used by forms-runner.

Brands are defined in configuration (config/settings.yml), the initial list is Cheshire East Council and South Gloucestershire Council.

The feature is behind the custom_branding feature flag, so has to be enabled per group. Otherwise the task list item is hidden and the brand configuration pages are inaccessible.

### DIFF
--- a/app/controllers/forms/brand_controller.rb
+++ b/app/controllers/forms/brand_controller.rb
@@ -1,12 +1,13 @@
 module Forms
   class BrandController < FormsController
+    before_action :check_user_has_permission
+    before_action :check_feature_flag
+
     def new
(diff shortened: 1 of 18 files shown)


SamJamCul (dismissed):
Looks good to me!

SamJamCul (approved):

PR #2908, opened 2026-07-10
Add brands admin UI

theseanything:
Brands that a form can be styled with were defined in the Settings config, so seeing which brands exist or changing them required a code change and deployment.

This moves brands into the database and adds a super-admin (read-only) for them. This is work to make sure managing brands doesn't become a dev only task. Pages for creating and editing brands will be added in later PRs.

This is the foundation work for providing API, so we can move brand configuration fulling into Forms Admin.

image
image


theseanything:
I don't what happened here - but I can't fix the git history. Going to raise two separate PRs anyways, one to move brands to the db and another for the UI work.

PR #2909, opened 2026-07-10
Let super admins enable group feature flags

theseanything:
Some features are rolled out per group through boolean columns on the groups table, but the only way to set those columns was a rake task that had to be run by an engineer. There was also no easy way to see which flags were already enabled for a group.

This adds a page, linked from the group page for super admins, that shows a checkbox for each group feature flag and saves the selection. The flags are derived from the features in settings.yml marked `enabled_by_group` that have a matching column on the groups table, so new flags show up without further changes.

Flags can only be switched on: a feature can change a form or group in ways that need manual work to undo, so flags that are already on are shown disabled and are never written back off.

image

image

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -108,19 +108,17 @@ class GroupsController < WebController
   # GET /groups/1/feature-flags
   def feature_flags
     authorize @group, :manage_feature_flags?
+
+    @feature_flags_input = Groups::FeatureFlagsInput.new(group: @group).assign_group_values
   end
 
   # POST /groups/1/feature-flags
   def update_feature_flags
     authorize @group, :manage_feature_flags?
 
-    # Feature flags can only be switched on, never off. Enabling a feature can change
-    # a group's forms or data in ways that would need to be manually reversed before
-    # it is safe to disable, so we only ever turn flags on here.
-    flags_to_enable = feature_flag_params.select { |_attr, value| ActiveModel::Type::Boolean.new.cast(value) }
-    @group.assign_attributes(flags_to_enable.transform_values { true })
+    @feature_flags_input = Groups::FeatureFlagsInput.new(feature_flags_input_params)
 
-    if @group.save
+    if @feature_flags_input.submit
       redirect_to @group, success: t("groups.success_messages.feature_flags"), status: :see_other
     else
       render :feature_flags, status: :unprocessable_content
@@ -223,8 +221,10 @@ private
     params.require(:group).permit(:name, :organisation_id)
   end
 
-  def feature_flag_params
-    params.require(:group).permit(*Group.feature_flag_attributes)
+  def feature_flags_input_params
+    # When every flag is already enabled the form has no enabled inputs, so the
+    # input params may be missing entirely.
+    params.fetch(:groups_feature_flags_input, {}).permit(*Group.feature_flag_attributes).merge(group: @group)
   end
 
   def confirm_upgrade_input_params
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -105,6 +105,28 @@ class GroupsController < WebController
     render :move
   end
 
+  # GET /groups/1/feature-flags
+  def feature_flags
+    authorize @group, :manage_feature_flags?
+  end
+
+  # POST /groups/1/feature-flags
+  def update_feature_flags
+    authorize @group, :manage_feature_flags?
+
+    # Feature flags can only be switched on, never off. Enabling a feature can change
+    # a group's forms or data in ways that would need to be manually reversed before
+    # it is safe to disable, so we only ever turn flags on here.
+    flags_to_enable = feature_flag_params.select { |_attr, value| ActiveModel::Type::Boolean.new.cast(value) }
+    @group.assign_attributes(flags_to_enable.transform_values { true })
+
+    if @group.save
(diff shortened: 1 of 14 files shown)


app/controllers/groups_controller.rb
```diff
@@ -105,6 +105,28 @@ def move
...
+  end
+
+  # POST /groups/1/feature-flags
+  def update_feature_flags
+    authorize @group, :manage_feature_flags?
+
+    # Feature flags can only be switched on, never off. Enabling a feature can change
+    # a group's forms or data in ways that would need to be manually reversed before
+    # it is safe to disable, so we only ever turn flags on here.
+    flags_to_enable = feature_flag_params.select { |_attr, value| ActiveModel::Type::Boolean.new.cast(value) }
```
SamJamCul: We'd usually have an input object for submitting a form like this. I know this is a very simple form, but I don't know if this justifies breaking that pattern?

app/views/groups/show.html.erb
```diff
@@ -66,6 +66,7 @@
     <li><%= govuk_link_to(t(".edit_group_name"), edit_group_path(@group)) %></li>
     <li><%= govuk_link_to(t(".edit_group_members"), group_members_path(@group)) %></li>
     <% if policy(@group).move? %><li><%= govuk_link_to(t(".move"), move_group_path(@group)) %></li><% end %>
+    <% if policy(@group).manage_feature_flags? %><li><%= govuk_link_to(t(".feature_flags"), feature_flags_group_path(@group)) %></li><% end %>
```
SamJamCul, line 69: I'm a bit nervous about how this blends in to other actions that can be taken on the group. Because it's an admin only and somewhat irreversible change that can take place, I think it might be good to indicate this somewhere.

config/locales/en.yml
```diff
@@ -743,6 +743,13 @@ en:
       success: The group, ‘%{group_name}’, has been deleted
     edit:
       title: Change the name of this group
+    feature_flags:
+      flags:
+        custom_branding_enabled: Custom branding
+        multiple_branches_enabled: Multiple branches and skip conditions
```
SamJamCul: I think just `Multiple branches`, skip conditions are a bit of a loaded term

config/locales/en.yml
```diff
@@ -743,6 +743,13 @@ en:
...
+        send_filler_answers_enabled: Send a copy of completed answers to form fillers
+      hint: You can turn a feature flag on, but not off. A feature can change how a form or group works, which might take extra steps to reverse. Contact the team if you need to switch a feature off.
```
SamJamCul: We can be more specific here than `the team` as it'll need to be a developer who's probably working on the feature itself.

app/controllers/groups_controller.rb
```diff
@@ -105,6 +105,28 @@ def move
...
+    @group.assign_attributes(flags_to_enable.transform_values { true })
+
+    if @group.save
+      redirect_to @group, success: t("groups.success_messages.feature_flags"), status: :see_other
```
SamJamCul: The success message appears even when no change has been made, so the wording `This group's feature flags have been updated` might be misleading

spec/requests/groups_controller_spec.rb
```diff
@@ -448,6 +448,86 @@
...
+    end
+  end
+
+  describe "POST /feature-flags" do
+    let(:feature_flag) { Group.feature_flag_attributes.first }
+    let(:feature_name) { feature_flag.delete_suffix("_enabled") }
+    let(:group) { create(:group, feature_flag => false) }
+
+    before do
+      skip "no group feature flags are configured" if Group.feature_flag_attributes.empty?
```
SamJamCul, line 482: Not sure I understand what's going on here?

app/views/groups/show.html.erb
theseanything: Do you mean indicate in the code (e.g. code comment) or in the actual UI? Could you give an example of the scenario you are worried about?

spec/requests/groups_controller_spec.rb
theseanything: This is to prevent running the tests (and causing failures) if there are no feature flags configured. It's hard to mock out that scenario in tests as it depends on a db migration (e.g. dropping all feature_flag_enabled columns).

I didn't want to spend to much effort rigorously testing this, because its just an internal feature, pretty low risk if the interface doesn't work when there's no feature flags to begin with.

app/views/groups/show.html.erb
SamJamCul: It's a bit contrived, but someone on support who is trying to make changes to a group may think that turning the flag on will resolve an issue, and then not be able to reverse it. If I didn't know that the feature flag option was only available to super admins, I might think that it being visible in that list means it would be an option available to org or group admins too. 

I guess something somewhere to indicate that this action is for Super Admins only, so the reason you're seeing it is because you're a super admin.

spec/requests/groups_controller_spec.rb
SamJamCul: Ah gotcha, thanks for explaining

app/views/groups/show.html.erb
theseanything: So I see what you're saying - I think we can defer figuring that out if becomes a problem.

theseanything:
@SamJamCul address your comments (see new commits)

SamJamCul (approved):
Awesome!

PR #2911, opened 2026-07-13
Show and filter organisations by agreement type

theseanything:
The organisations list showed a yes/no "MOU signed" column and filter, however want to replicate the information available in the MOUs page.

This replaces the "MOU signed" column with an "Agreement type" column showing which agreements each organisation has signed ("Crown MOU", "Non-crown agreement", or "None"), reusing the existing agreement type translations. The filter is updated to match, offering all organisations, Crown MOU, non-crown agreement, or no agreement signed.

image


thomasiles (approved):
LGTM

PR #2912, opened 2026-07-13
Remove the dedicated MOUs list page

theseanything:
## What

Removes the standalone MOUs list page at `/mous` in favour of the MOUs and agreements shown on the organisation pages.

- Removes the `mou_signatures#index` route, action and view (the MOU signing flow is unchanged)
- Removes the "MOUs" header navigation item — the "Organisations" item covers this now
- Updates the sitemap to link to the organisations page instead of the MOUs list, keeping the link to the MOU signing page
- Deletes the now-unused `MouSignaturePolicy` and locale strings
- Rewrites the MOU feature specs to view signed MOUs via the organisation pages

## Why

The organisations pages list each organisation's signed MOUs and agreements (and the organisations index shows MOU-signed status with filtering), so the separate list page is redundant.


theseanything:
Superseded by #2919

PR #2913, opened 2026-07-13
Add Brand model

theseanything:

Trello card: 

Adds a `Brand` model backed by a new `brands` table, as a first step towards storing branding configuration in the database rather than in code.

Each brand has a `name` and a unique `slug`, enforced both by model validations and a unique database index. Seed data creates a couple of example brands in local development and review environments so the model can be exercised without manual setup.


theseanything:
superseded by https://github.com/govuk-forms/forms-admin/pull/2914

PR #2914, opened 2026-07-13
Add Brand model

theseanything:
Brands are currently configured in application code, so changing them requires a deployment.

This adds a Brand model with a name and unique slug, and seeds example brands in local development and review environments. Slug uniqueness is enforced with a database index as well as a model validation.

This lays the groundwork for moving brand configuration into the database — in later work we can add an admin UI for managing brands and an API to serve the configuration to forms-runner.

Will remove brand config from Setting YAML - once the existing brands are created in prod.


SamJamCul (approved):
Looks good!

PR #2918, opened 2026-07-13
Fix some flaky tests which rely on Db order

thomasiles:
### Fix some flaky tests which rely on the order of results from the database

Similar to https://github.com/govuk-forms/forms-admin/pull/2710, we try and fix some tests which have been tagged with `:flaky`.

Some tests fail because they depend on the order of results from the database, when it isn't defined. 

To address those, I added the [unreliable gem](https://github.com/jamiemccarthy/unreliable) which hooks into rails SQL calls and applies a random order to results which haven't already been ordered.

This helps show flaky tests which need fixing.

This PR starts off by adding the gem, tagging failing tests (there were a few false positives where the gem got confused about more complicated SQL), then fixing the issues by either changing the tests to not rely on order, or making the application code return results in a set order.

At the end, we remove the gem. I think it could be a useful addition to help prevent issues in the future. I'm not going to add another dependency in this PR though.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -135,6 +135,9 @@ group :test do
   # Enable running the specs in parallel (and optionally with Spring)
   gem "parallel_rspec"
   gem "spring-prspec"
+
+  # Look for tests which rely on implicit ordering of DB queries
+  gem "unreliable"
 end
 
 gem "mobility", "~> 1.3"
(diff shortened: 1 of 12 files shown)


DavidBiddle (approved):
Nice!

PR #2921, opened 2026-07-14
Add pages for creating and editing brands

theseanything:
This PR adds pages for super admins to create a brand and to edit a brands, following the same patterns as the groups section. The brands list page gets a button for creating a brand, and each brand's page gets a link to edit it.

This is a stepping stone towards making branding self service - to start allow management by other's in the team and not just a dev task.

image
image
image
image

### DIFF
--- a/app/controllers/brands_controller.rb
+++ b/app/controllers/brands_controller.rb
@@ -1,4 +1,5 @@
 class BrandsController < WebController
+  before_action :set_brand, only: %i[show edit update]
   after_action :verify_authorized
 
   def index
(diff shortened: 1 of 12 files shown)


SamJamCul (dismissed):
Looks good! We could add a custom error message when the slug is already used, but given it's internal it's not a big issue

Image

theseanything:
@SamJamCul good catch - done some other tidy ups. (Added 4 new small scoped commits)

- Validation on the slugs themselves (should only be kebab case)
- Some code tidy ups - de-dupes of permission setting and setting brands, shared specs
- Improved the error messages https://github.com/govuk-forms/forms-admin/pull/2921/changes/6bcb3581ae424143262639c88a4fc067de3dc1ee

SamJamCul (approved):

PR #2922, opened 2026-07-14
Improve factories

thomasiles:
### Improvements to our test Factories

Trello card: https://trello.com/c/L37Bdby6/260-improve-the-tests-in-forms-admin

Our FactoryBot tests have become a bit tangled. I'm trying to uncouple them a bit and stop using the database when it's not needed and stop hardcoding database ids, which, I think, causes reliability problems when tests are run in certain orders or at the same time.

This PR tackles some of the issues which stop us using build, which doesn't use the database instead of create. It doesn't switch any tests yet, it's just preparing the groundwork.

One of the biggest issues is how we treat organisations. I've tried to keep the commits small but one commit makes small changes to a lot of specs which relied on orgs being the same.

You can use the rails console to test the factories with `Factorybot.create` or `FactoryBot.build` enabling SQL logging is also useful:

```
ActiveRecord::Base.logger = Logger.new(STDOUT)
ActiveRecord::LogSubscriber.attach_to :active_record
```

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -159,6 +159,6 @@ private
   end
 
   def requires_organisation?
-    organisation_id_was.present? || role_changed?(to: :organisation_admin)
+    organisation.blank? && (organisation_id_was.present? || role_changed?(to: :organisation_admin))
   end
 end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -424,7 +424,7 @@ describe User, type: :model do
       context "and the user's organisation is not the same as given" do
         it "returns false" do
           user = create(:organisation_admin_user)
-          other_org = build(:organisation, id: 2)
+          other_org = build(:organisation)
 
           expect(user.is_organisations_admin?(other_org)).to be(false)
         end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -58,6 +58,16 @@ describe User, type: :model do
       end
     end
 
+    context "when reassigning to a new, unsaved organisation" do
+      let(:user) { create(:user, organisation: create(:organisation)) }
+
+      it "is valid even though the new organisation is not yet persisted" do
+        user.organisation = build(:organisation)
+
+        expect(user).to be_valid
+      end
+    end
+
     context "when updating name" do
       let(:user) { create :user, :with_no_name }
 
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -128,7 +128,7 @@ describe User, type: :model do
   describe "associations" do
     it "destroys associated memberships" do
       user = create(:user)
-      group = create(:group)
+      group = create(:group, organisation: user.organisation)
 
       membership = create(:membership, user:, group:)
       user.destroy!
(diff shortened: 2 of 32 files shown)


spec/models/user_spec.rb
theseanything: We've got the hardcoded org id, which could clash now?

spec/services/mailchimp/list_sync_service_spec.rb
```diff
@@ -69,17 +72,19 @@
...
-    let(:user_with_access_admin) { create(:user, email: "admin_user@example.com") }
-    let(:user_without_access) { create(:user, email: "inactive_user@example.com", has_access: false) }
-    let(:user_access_and_admin_with_mou) { create(:user, email: "admin_mou_user@example.com") }
-
+    let(:signed_mou_org) { create(:organisation) }
+    let(:user_with_access_and_mou) { create(:user, email: "mou_user@example.com", organisation: signed_mou_org) }
+    let(:user_with_access_admin) { create(:user, email: "admin_user@example.com", organisation: signed_mou_org) }
+    let(:user_without_access) { create(:user, email: "inactive_user@example.com", has_access: false, organisation: signed_mou_org) }
+    let(:user_access_and_admin_with_mou) { create(:user, email: "admin_mou_user@example.com", organisation: signed_mou_org) }
     let(:non_crown_org) { create(:organisation, id: 2, slug: "non-crown-org") }
```
theseanything: Drop the hard coded org id here? as will now collide with signed_mou_org?

app/models/user.rb
```diff
@@ -159,6 +159,6 @@ def requires_name?
   end
 
   def requires_organisation?
-    organisation_id_was.present? || role_changed?(to: :organisation_admin)
+    organisation.blank? && (organisation_id_was.present? || role_changed?(to: :organisation_admin))
```
theseanything, line 162: this is going to add extra db query every time, this function is called - is that okay?

spec/models/user_spec.rb
thomasiles: Good spot. This doesn't cause a clash because using build now doesn't create anything in the database where the conflicts happen.

I've removed it because it's not needed and it's just confusing to have it.

spec/services/mailchimp/list_sync_service_spec.rb
thomasiles: yes good spot, I've changed it.

app/models/user.rb
thomasiles: Your right, I've really struggled to fix this.

I think rails by default does a query based because organisation is an association but I'm not certain. I haven't found a way to check the value of current object and what it was without issuing a query.

I think the bigger fix is rethinking this validation.

What the current code is aiming for is:
- organisation_admins must have an organisation
- user's can only move between organisations once set

I think this would be better covered by separating them into two. Org admins must have a organisation is straightforward. 

I'm less certain about only allowing users to change, not remove organisation. It's there because when users first sign up they won't have an organisation. I think we could handle it better by either having two states somehow, incomplete and complete user, or by just dropping it totally.

I think I'd rather leave this code as it is and look at tackling it in another PR. I'm open to ideas though!

theseanything: Yah, I think you can check the id first (which just checks the attribute on user) first, then do the db if you need. 
e.g. 
```suggestion
    organisation_id.blank? && organisation.blank? && (organisation_id_was.present? || role_changed?(to: :organisation_admin))
```

Tbh, just looking through the code, this might be called that much anyways - it's only user saves and signing in. So probably minimal impact on perf.

theseanything (approved):

PR #2923, opened 2026-07-14
Add allowed brands for organisations

theseanything:
Brands are a global list, but each organisation should only be able to use brands it has been granted. This can help make sure that brands that each organisation see's is manageable list and relevant (i.e. Croydon council is unlikely going to use Hackney council branding). It also prevents user creating or impersonating organisations for which they're not associated with.

The management is only available via the organisations page and can only be done by a super admin.

The UI uses the autocomplete select, in case the list of brands becomes very long (potentially, e.g. each local council has its own brand).


SamJamCul (approved):
Looks good! We could also show all the organisations that a Brand is linked to on the Brands show page, but that can be later

PR #2924, opened 2026-07-14
Show result counts on the users list page

theseanything:
This adds a count above the table matching the one on the organisations list: "Showing x to y of n users" when the results span multiple pages, or just the total number of users when they fit on one page. (Same as organisation page)

Helpful to see total number of users, and number of users when filtered.


stephencdaly (approved):
Nice

PR #2933, opened 2026-07-15
Add DeliveryConfigurations for forms to allow sending submissions by multiple methods

stephencdaly:

Trello card: https://trello.com/c/WMnjLlJV/245-allow-both-email-and-s3-submissions-simultaneously

Move the settings that decide where we send a form's submissions to a `delivery_configurations` table, which has a many to one relationship with forms.

This will allow for submissions to be sent by both email and to an s3 bucket for the same form.

Also add `DeliveryConfiguration` records for batch submissions, which are identified by the `frequency` column.

This PR ensures that appropriate DeliveryConfigurations are created when a form is updated, and adds a data migration rake task to create records for existing forms.

We do not yet consume the DeliveryConfigurations, for example to determine the content to display on the live form page but will do in a subsequent PR once the data migration has been run.


thomasiles (commented):
This works for me, nice one 🏅

thomasiles (dismissed):

app/models/form.rb
```diff
@@ -17,6 +17,7 @@ class Form < ApplicationRecord
   has_one :draft_welsh_form_document, -> { where tag: "draft", language: :cy }, class_name: "FormDocument"
   has_one :draft_form_document, -> { where tag: "draft", language: :en }, class_name: "FormDocument"
   has_many :conditions, through: :pages, source: :routing_conditions
+  has_many :delivery_configurations
```
theseanything: When a form is deleted, we probably just want to delete the delivery configs

```suggestion
  has_many :delivery_configurations, dependent: :destroy
```

db/migrate/20260715095436_create_delivery_configurations.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDeliveryConfigurations < ActiveRecord::Migration[8.1]
+  def change
+    create_table :delivery_configurations do |t|
+      t.references :form, null: false, foreign_key: true
```
theseanything: again to delete delivery configs when the a form is deleted
```suggestion
      t.references :form, null: false, foreign_key: { on_delete: :cascade }
```

app/models/form.rb
theseanything: You'll need to run the rake task soon after deploy, as this is also going to make `changed_from_live_version?` true, because the "delivery_configurations" will be missing from form_documents. Which will make it look like there are changes to the English for to go live in the Admin UI.

lib/tasks/forms.rake
```diff
@@ -85,7 +85,32 @@ namespace :forms do
...
+
+      delivery_configuration = form.delivery_configurations.find_or_initialize_by(delivery_method: :email, frequency: :immediate)
+      delivery_configuration.formats = submission_format
+      delivery_configuration.save!
+
+      form.delivery_configurations.where(delivery_method: :s3).destroy_all
+      form.save!
+
+      if form.is_live?
+        form_document = form.live_form_document
```
theseanything, line 103: Does this update both the English and Welsh? as `live_welsh_form_document` is different from `live_form_document`

app/input_objects/forms/batch_submissions_input.rb
```diff
@@ -2,8 +2,20 @@ class Forms::BatchSubmissionsInput < BaseInput
...
+    form.send_daily_submission_batch = selected_frequencies.include?("daily")
+    form.send_weekly_submission_batch = selected_frequencies.include?("weekly")
+
+    %w[daily weekly].each do |frequency|
+      if selected_frequencies.include?(frequency)
+        form.delivery_configurations.find_or_create_by!(delivery_method: "email", frequency:, formats: %w[csv])
+      else
+        form.delivery_configurations.where(delivery_method: "email", frequency:).destroy_all
+      end
+    end
```
theseanything, line 16: What about when a draft is discarded? e.g. RevertDraftFormService

app/input_objects/forms/submission_email_input.rb
```diff
@@ -43,6 +43,7 @@ def confirm_confirmation_code
 
     # Update the submission email in the form
     form.submission_email = temporary_submission_email
+    form.delivery_configurations.find_or_create_by!(delivery_method: :email, frequency: :immediate)
```
theseanything: Does this behave okay for a user changing a submission email address, when they have other formats already enabled? e.g. will it find the existing one with say json enabled?

stephencdaly: Yep, there's a test that checks that it doesn't create a new DeliveryConfiguration when there's already one which has formats enabled.

stephencdaly:
@theseanything thanks for the review, I'll make all those changes.

app/models/form.rb
stephencdaly: Good spot, I can update `changed_from_live_version?` to ignore the delivery_configurations for now and update it again in the next PR

stephencdaly: Hmm actually, this method seems fairly tricky. Generally we don't want to need to update the live form document for all migrations - instead we should add backwards compatibility as agreed in the ADR on versioning.

In this case I'm migrating live form documents because it will be a lot easier to.

I think we need to rethink the comparison that this method does and whether we need to do it or whether there's another way to determine which content to show.

stephencdaly: I think maybe I'll just leave this as it is on second thoughts. There will be other changes that cause the user to have to make the English version live when we've made changes behind the scenes and it will only affect a small number of users. It's not the end of the world if sometimes users needs to make the English version live again even if they don't think they've made changes.

app/input_objects/forms/batch_submissions_input.rb
stephencdaly: Updated RevertDraftFormService to reset the DeliveryConfigurations

lib/tasks/forms.rake
stephencdaly: Updated this rake task to update all live form documents

db/migrate/20260715095436_create_delivery_configurations.rb
stephencdaly: done

app/models/form.rb
stephencdaly: done

thomasiles (approved):

PR #2940, opened 2026-07-16
Add rake task to retrieve live and draft for docs

SamJamCul:

As part of investigating issues with a draft form document, this task will make it easier to compare the form's current state in the db. I've kept this as a task under the `forms` namespace as it's easy to work out a form's ID, and get the form documents for that form using it. 

### DIFF
--- a/lib/tasks/forms.rake
+++ b/lib/tasks/forms.rake
@@ -174,6 +174,24 @@ namespace :forms do
       Rails.logger.info "Form #{form.id} (\"#{form.name}\") created by #{creator&.name || 'No creator'} with organisation #{creator&.organisation&.name || 'N/A'}"
     end
   end
+
+  desc "Show a form's form_document as JSON"
+  task :show_form_document, %i[form_id tag language] => :environment do |_, args|
+    usage_message = "usage: rake forms:show_form_document[<form_id>, <tag>, <language>]".freeze
+
+    abort usage_message if args[:form_id].blank? || args[:tag].blank?
+    language = args[:language].presence || "en"
+
+    abort "tag must be one of draft, live or archived" unless %w[draft live archived].include?(args[:tag])
+    abort "language must be en or cy" unless %w[en cy].include?(language)
+
+    form = Form.find(args[:form_id])
+
+    form_document = form.form_documents.find_by(tag: args[:tag], language:)
+    abort "#{fmt_form(form)} does not have a #{args[:tag]} #{language} form document" if form_document.blank?
+
+    puts JSON.pretty_generate(form_document.as_json)
+  end
 end
 
 def move_forms(form_ids, group_id)
--- a/spec/lib/tasks/forms.rake_spec.rb
+++ b/spec/lib/tasks/forms.rake_spec.rb
@@ -807,4 +807,59 @@ RSpec.describe "forms.rake", type: :task do
       end
     end
   end
+
+  describe "forms:show_form_document" do
+    subject(:task) do
+      Rake::Task["forms:show_form_document"]
+    end
+
+    let(:form) { create(:form) }
+
+    it "prints the requested form document as JSON" do
+      expect { task.invoke(form.id, "draft", "en") }
+        .to output(/"id": #{form.draft_form_document.id}/).to_stdout
+    end
+
+    it "prints the requested English form document when no language is given" do
+      expect { task.invoke(form.id, "draft") }
+        .to output(/"id": #{form.draft_form_document.id}/).to_stdout
+    end
+
+    it "aborts with a usage message when arguments are missing" do
+      expect {
+        task.invoke(form.id)
+      }.to raise_error(SystemExit)
+         .and output(/usage: rake forms:show_form_document\[<form_id>, <tag>, <language>\]/).to_stderr
+    end
+
+    it "aborts when the tag is invalid" do
+      expect {
+        task.invoke(form.id, "invalid", "en")
+      }.to raise_error(SystemExit)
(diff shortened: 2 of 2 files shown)


lib/tasks/forms.rake
```diff
@@ -138,6 +138,24 @@ namespace :forms do
       Rails.logger.info "Form #{form.id} (\"#{form.name}\") created by #{creator&.name || 'No creator'} with organisation #{creator&.organisation&.name || 'N/A'}"
     end
   end
+
+  desc "Show a form's live_form_document"
+  task :show_live_form_document, %i[form_id] => :environment do |_, args|
```
stephencdaly: Can we have a single task that takes a `tag`, and then we can also look at archived? Would also be good to be able to view the Welsh form document, so could it either take a language or show both english and welsh if there are both?

stephencdaly (dismissed):
Added a suggestion but nothing to stop this going in as it is to use for the support task

lib/tasks/forms.rake
SamJamCul: Can do, I'll adjust it

lib/tasks/forms.rake
```diff
@@ -138,6 +138,27 @@ namespace :forms do
...
+    abort "language must be one of english or welsh" unless %w[english welsh].include?(args[:language])
+
+    form = Form.find(args[:form_id])
+
+    abort "#{fmt_form(form)} is not live" if args[:tag] == "live" && !form.is_live?
+    abort "#{fmt_form(form)} is not archived" if args[:tag] == "archived" && !form.is_archived?
+    abort "#{fmt_form(form)} does not have a Welsh translation" if args[:language] == "welsh" && !form.has_welsh_translation?
+
+    form_document_method = "#{args[:tag]}_#{args[:language] == 'english' ? 'form_document' : 'welsh_form_document'}"
+    form_document = form.public_send(form_document_method)
```
SamJamCul: Not sure about this, it relies on our naming convention for the form document methods to be consistent, which they currently are - but I can imagine it getting out of sync. The alternative was a bit of a chunky if/case statement that spelled out each method based on the tag/language combination

lib/tasks/forms.rake
SamJamCul: I've updated it to take a tag and language

lib/tasks/forms.rake
```diff
@@ -138,6 +138,27 @@ namespace :forms do
...
     end
   end
+
+  desc "Show a form's form_document as JSON"
+  task :show_form_document, %i[form_id tag language] => :environment do |_, args|
+    usage_message = "usage: rake forms:show_form_document[<form_id>, <tag>, <locale>]".freeze
+
+    abort usage_message if args.to_a.length != 3
+    abort "tag must be one of draft, live or archived" unless %w[draft live archived].include?(args[:tag])
+    abort "language must be one of english or welsh" unless %w[english welsh].include?(args[:language])
```
stephencdaly: Can this be `en` or `cy` to match how we refer to locales elsewhere? Does it also make sense to allow this not to be provided and default to returning the english?

lib/tasks/forms.rake
stephencdaly: Can we just do `form.form_documents.find_by(tag:, language:)`?

lib/tasks/forms.rake
```diff
@@ -138,6 +138,27 @@ namespace :forms do
...
+
+    form = Form.find(args[:form_id])
+
+    abort "#{fmt_form(form)} is not live" if args[:tag] == "live" && !form.is_live?
+    abort "#{fmt_form(form)} is not archived" if args[:tag] == "archived" && !form.is_archived?
+    abort "#{fmt_form(form)} does not have a Welsh translation" if args[:language] == "welsh" && !form.has_welsh_translation?
```
stephencdaly: We probably don't need all these checks - just try to find the document using the `.find_by` and if no document is returned, let the user know that no document exists with that tag and locale?

stephencdaly (commented):
We can simplify by using the `find_by` ActiveRecord query method

lib/tasks/forms.rake
SamJamCul: I've swapped over to using `en` and `cy`, and it'll use `en` if no language is given as a third arg

lib/tasks/forms.rake
SamJamCul: Changed it to this

lib/tasks/forms.rake
SamJamCul: I've changed it to the `find_by` version now

lib/tasks/forms.rake
```diff
@@ -138,6 +138,24 @@ namespace :forms do
       Rails.logger.info "Form #{form.id} (\"#{form.name}\") created by #{creator&.name || 'No creator'} with organisation #{creator&.organisation&.name || 'N/A'}"
     end
   end
+
+  desc "Show a form's form_document as JSON"
+  task :show_form_document, %i[form_id tag language] => :environment do |_, args|
+    usage_message = "usage: rake forms:show_form_document[<form_id>, <tag>, <language>]".freeze
+
+    abort usage_message if args[:form_id].blank? || args[:tag].blank?
+    args[:locale] = "en" if args[:language].blank?
```
stephencdaly:
```suggestion
    args[:language] = "en" if args[:language].blank?
```

SamJamCul: Fixed, also found a bug with how it assigned the default `en` and wrote a test for it.

stephencdaly (dismissed):

stephencdaly (approved):

PR #2942, opened 2026-07-17
Use DeliveryConfigurations instead of old form attributes

stephencdaly:

Trello card: https://trello.com/c/WMnjLlJV

The following columns have been replaced by the delivery_configurations table, and are no longer used by forms-runner
- submission_type
- submission_format
- send_daily_submission_batch
- send_weekly_submission_batch

Replace places where we were reading and updating these attributes with only reading and updating the DeliveryConfigurations.

This is in preparation to ignore and drop these columns. 

This PR does not update the rake tasks for configuring whether S3 submissions are enabled, this will be done in the next PR.


DavidBiddle (approved):
Commits are well organised, code change makes sense and I can see the change working in the report, task list and the live form page 🎉

PR #2954, opened 2026-07-21
Update rake tasks for changing delivery configurations

stephencdaly:

Trello card: https://trello.com/c/WMnjLlJV

Update the rake tasks for enabling/disabling S3 and email submissions to have the following tasks:
- enable_s3
- disable_s3
- enable_email
- disable_email

These tasks will now only edit the delivery_configurations and not edit the submission_type and submission_format which are deprecated and unused.

The new tasks make it possible to have both S3 and email submissions enabled at the same time.

### Review notes

As the tasks and tests have changed quite significantly, it might be easier to review using the split diff view or by checking to code out locally to look at the new tasks.


DavidBiddle (approved):
Code looks good and I've tested the tasks locally :tada:

You're probably already aware but we'll need to update this wiki page with the new tasks once this is merged: https://github.com/govuk-forms/forms-team/wiki/How-to-set-up-sending-submissions-to-a-department%E2%80%99s-AWS-S3-bucket

PR #2985, opened 2026-07-29
Change exit page model

SamJamCul:

Trello card: https://trello.com/c/gZGRBU8Y/3172-change-add-edit-delete-exit-page-to-work-with-new-style-exit-page-model

An ExitPage will be created and destroyed in conjunction with Conditions having their exit page markdown and headings modified. 

This means that any exit page changes that take place in the app from this point forward will synchronise the Exit Page table with changes to exit pages. We still need to create a migration to update all existing exit page Conditions so that they have a corresponding ExitPage. 

This also holds for translations for exit pages, which will also be mapped over to the Exit Page table. 

Areas of the app that we've tested this with - CRUDding exit pages, CRUDding exit page translations, copying a form, deleting the draft of a live/archived form 

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -52,9 +52,8 @@ class Condition < ApplicationRecord
   end
 
   def destroy_and_update_form!
-    exit_page_to_delete = exit_page
+    exit_page.presence&.destroy!
     destroy! && form.save_question_changes!
-    exit_page_to_delete.presence&.destroy! # unless FeatureService.new(group: form.group).enabled?(:multiple_branches)
   end
 
   def validation_errors
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -10,7 +10,6 @@ class Condition < ApplicationRecord
 
   has_one :form, through: :routing_page
   belongs_to :exit_page, optional: true, autosave: true
-  validates_associated :exit_page
 
   before_destroy :destroy_postconditions
   before_validation :sync_exit_page
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -9,15 +9,40 @@ class Condition < ApplicationRecord
   belongs_to :goto_page, class_name: "Page", optional: true
 
   has_one :form, through: :routing_page
-  belongs_to :exit_page, optional: true
+  belongs_to :exit_page, optional: true, autosave: true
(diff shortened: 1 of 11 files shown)


app/models/condition.rb
```diff
@@ -28,7 +52,9 @@ def save_and_update_form
   end
 
   def destroy_and_update_form!
+    exit_page_to_delete = exit_page
     destroy! && form.save_question_changes!
+    exit_page_to_delete.presence&.destroy! # unless FeatureService.new(group: form.group).enabled?(:multiple_branches)
```
theseanything: If a user deletes an exit page route, I think the draft form document is saved before the exit page is destroyed, so the document still shows an exit page that no route points to. Until someone edits it and regenerates it. Also if we get an exception between the two steps, the ExitPage is orphaned for good. 

I think we just need to destroying the exit page before form.save_question_changes!?

lfdebrux (dismissed):
I tested this branch locally, it works for me.

app/services/revert_draft_form_service.rb
```diff
@@ -188,6 +188,7 @@ def clear_welsh_translations
     form.translations.where(locale: :cy).delete_all
     Page::Translation.where(locale: :cy, page_id: form.page_ids).delete_all
     Condition::Translation.where(locale: :cy, condition_id: form.condition_ids).delete_all
+    ExitPage::Translation.where(locale: :cy, exit_page_id: form.exit_page_ids).delete_all
```
theseanything, line 193: Are we removing the exit page itself? Or just the translations?

thomasiles: Just the translations here. The exit page itself is removed by setting the condition attributes and saving the object. We don't need anything extra in that case. We need to remove the translations directly when Welsh isn't needed though.

app/models/condition.rb
thomasiles: Yes, good spot.

We ended up with this before we had an association to condition on exit page, which meant rails wasn't removing the reference in the table before deleting it, causing an exception.

Now we have the association we can go back to something more sensible.

It's a candidate for wrapping in a transaction. I'm hesitant to do that as it calls a chain of methods. I've added a new commit with your suggestion.

lfdebrux (approved):

PR #2986, opened 2026-07-29
Handle one match or no matching organisations for the user

stephencdaly:

Trello card: https://trello.com/c/LH68oEwk

- When only one organisation has a domain matching the user's email address, display a different page with "Yes" and "No" radio buttons for the user to confirm whether it is their organisation.
- When no organisations match the user's email address, display the page telling them to contact support.

The page shown when there is only one matching organisation:

Screenshot 2026-07-29 at 12 22 43

If "No" is selected, we display the same page as we show if there were no organisations found matching the user's domain:

Screenshot 2026-07-29 at 12 22 52

### DIFF
--- a/app/views/account/organisations/confirm.html.erb
+++ b/app/views/account/organisations/confirm.html.erb
@@ -1,0 +1,22 @@
+<% set_page_title(title_with_error_prefix(t('page_titles.confirm_account_organisation'), @confirm_organisation_input.errors.any?)) %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with(model: @confirm_organisation_input, url: confirm_account_organisation_path) do |f| %>
+      <% if @confirm_organisation_input&.errors.any? %>
+        <%= f.govuk_error_summary %>
+      <% end %>
+
+      <h1 class="govuk-heading-l">
+        <%= t('page_titles.confirm_account_organisation') %>
+      </h1>
+
+      <%= t('.body_html', organisation: @confirm_organisation_input.organisation.name_with_abbreviation) %>
+
+      <%= f.govuk_collection_radio_buttons :confirm,
+        @confirm_organisation_input.values, ->(option) { option },
+        legend: { text: t('.radios_legend', organisation: @confirm_organisation_input.organisation.name_with_abbreviation), size: 'm' } %>
+
+      <%= f.govuk_submit t("save_and_continue") %>
+      <% end %>
+  </div>
+</div>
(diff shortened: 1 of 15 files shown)


StephenGill:
@stephencdaly screenshots look good to me 👍

app/views/account/organisations/confirm.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      <% end %>
+
+      <h1 class="govuk-heading-l">
+        <%= t('page_titles.confirm_account_organisation') %>
+      </h1>
+
+      <%= t('.body_html', organisation: @confirm_organisation_input.organisation.name_with_abbreviation) %>
+
+      <%= f.govuk_collection_radio_buttons :confirm,
+        @confirm_organisation_input.values, ->(option) { option }, ->(option) { t('helpers.label.account/confirm_organisation_input.options.' + "#{option}") },
```
DavidBiddle: You can simplify this by using by moving the translation keys to `account_confirm_organisation_input.confirm_options`

```suggestion
        @confirm_organisation_input.values, ->(option) { option },
```

config/locales/en.yml
```diff
@@ -1,13 +1,21 @@
...
-    organisation:
-      details_summary: If your organisation is not listed
-      not_listed_html: |
-        <p>If you’re from a central government organisation that’s not listed, %{contact_link}.</p>
-        <p>At the moment, GOV.UK Forms is only available to central government organisations. If you’re from an organisation outside central government, you can <a href="https://www.forms.service.gov.uk/forthcoming-features">read about future plans for GOV.UK Forms</a>.</p>
-      not_listed_option: The organisation I work for is not listed
-      you_created_account_with: You created your account using an @%{email_domain} email address.
+    organisations:
+      confirm:
+        body_html: "<p>Based on the email address you used to create your account, we think your organisation is <strong>%{organisation}</strong>.</p>\n"
```
DavidBiddle: Is the `\n` here needed? I don't think it should affect how the HTML is rendered but I'm curious about whether removing it breaks something

stephencdaly: Oh strange! No, this looks like it must have been some kinda strange autocomplete suggestion from the IDE that I didn't notice, will remove.

DavidBiddle (commented):
Raised a couple of suggestions but looks good overall and works as I'd expect locally

app/views/account/organisations/confirm.html.erb
stephencdaly: Nice, thanks. I've updated

stephencdaly:
Thanks for the review @DavidBiddle I've made those changes

DavidBiddle (approved):

PR #2991, opened 2026-07-30
Use organisation domains for email address validation

stephencdaly:

Trello card: https://trello.com/c/NQj6zljA

For submission and contact email addresses, we were were allowing email addresses on known government domains or with a domain matching the current user's domain.

Instead of looking at the current user's domain, we now look at the domains associated with the organisation the form belongs to.

Only allow non-government email addresses with domains exactly matching the organisation's domains rather than sub-domains. If we need to allow a sub-domain, we can add it to the organisation's domains. We can revisit this if we think it's too restrictive.

### DIFF
--- a/app/controllers/forms/contact_details_controller.rb
+++ b/app/controllers/forms/contact_details_controller.rb
@@ -19,7 +19,7 @@ module Forms
   private
 
     def contact_details_input_params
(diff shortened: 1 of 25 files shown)


DavidBiddle (approved):
Code looks good, and validation works as expected in Review :tada:

PR #2992, opened 2026-07-31
Add rake task for populating domains from a list

DavidBiddle:

Trello card: https://trello.com/c/KTHYph7i/3195-get-latest-organisation-email-domains-data-from-notify

Adds a new rake task for uploading the domain data from Notify.

This takes a list of arguments in the form `An organisation name:example.com\,example.gov.uk`, e.g.:
```zsh
bin/rake "organisations:domains:populate_from_list[Government Digital Service:digital.cabinet-office.gov.uk\,dsit.gov.uk,Department for Science\, Innovation and Technology:dsit.gov.uk]"
```

This uses a list of arguments which is a bit unwieldy - ideally we'd use CSV for this, but that makes it hard to run this task in production without adding the full CSV to the repo. When running this in production it probably makes sense to run this in a few smaller batches.

I've tested locally with the real Notify data and the latest organisations from the GOV.UK organisations API.

### DIFF
--- a/lib/tasks/organisations.rake
+++ b/lib/tasks/organisations.rake
@@ -158,6 +158,44 @@ namespace :organisations do
         Rails.logger.info("Added domains for #{organisation.name}: #{domains.join(', ')}")
       end
     end
+
+    desc "Populate organisation domains from a list of organisations and domains"
+    task :populate_from_list, [] => :environment do |_, args|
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):
Looks like it does what we need and works for me as expected

PR #2997, opened 2026-08-04
Add exit page migration rake task

SamJamCul:
Adds the forms:sync_all_exit_pages task, which will identify all Conditions that have exit page content, but no exit page objects associated - and then sync them up.

Conditions will be touched by this change, as they need to have their exit_page_id updated when a new exit page is created and associated with them.

https://trello.com/c/gZGRBU8Y/3172-change-add-edit-delete-exit-page-to-work-with-new-style-exit-page-model

### DIFF
--- a/lib/tasks/forms.rake
+++ b/lib/tasks/forms.rake
@@ -256,6 +256,20 @@ namespace :forms do
 
     puts JSON.pretty_generate(form_document.as_json)
   end
+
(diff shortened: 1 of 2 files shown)


thomasiles (approved):
Looks good, it might be a bit slow when run in production but I don't think that matters?

PR #3006, opened 2026-08-05
Move exit pages to condition

thomasiles:
### Move the existing exit page controllers and input objects out the way

Trello card: https://trello.com/c/gecvdskV/3173-change-add-edit-delete-exit-page-to-work-with-multiple-branches

We are changing how exit pages are created, updated and deleted for the multiple branch feature.

We can't delete the existing exit pages code because they are still being used in production until we switch to multiple branch routing.

To add the new exit page routes for multiple branches we need to move the existing code out of the way. We are moving it instead of adapting it because there are differences in the new implementation, such as not depending on conditions.

When the new routes for exit pages have been added and the multiple branches feature is released, we will be able to remove the exit page code under conditions.

This PR shouldn't create any difference in how the product behaves. It's a code refactor to make implementing exit pages for multiple branches easier.

## Details

This is split into commits. The main changes are moving the controller under Condition, instead of Page and moving input objects.

The URLs for exit pages don't change but the path helpers have. We go through the code and tests to update them to the new values.

To move the input objects, we need to change the locales to match and check the controllers expect the correct params. Both depend on the name of the input object.

Whilst this PR is big, most of the changes are just altering the names of things.

### DIFF
--- a/app/controllers/pages/conditions_controller.rb
+++ b/app/controllers/pages/conditions_controller.rb
@@ -100,7 +100,7 @@ class Pages::ConditionsController < PagesController
 
   def confirm_delete_exit_page
     condition = page.routing_conditions.find(params[:condition_id])
-    delete_exit_page_input = Pages::DeleteExitPageInput.new
+    delete_exit_page_input = Conditions::DeleteExitPageInput.new
 
     render template: "pages/conditions/confirm_delete_exit_page", locals: {
       answer_value: params.require(:answer_value),
@@ -115,7 +115,7 @@ class Pages::ConditionsController < PagesController
 
     return redirect_to form_pages_path(current_form.id) unless condition.exit_page?
 
-    delete_exit_page_input = Pages::DeleteExitPageInput.new(delete_exit_page_params)
+    delete_exit_page_input = Conditions::DeleteExitPageInput.new(delete_exit_page_params)
 
     unless delete_exit_page_input.valid?
       return render template: "pages/conditions/confirm_delete_exit_page", locals: {
--- a/app/controllers/pages/conditions_controller.rb
+++ b/app/controllers/pages/conditions_controller.rb
@@ -28,7 +28,7 @@ class Pages::ConditionsController < PagesController
 
     if condition_input.submit
       if condition_input.create_exit_page?
-        redirect_to new_exit_page_path(current_form.id, page.id, answer_value: condition_input.answer_value)
+        redirect_to conditions_exit_page_new_path(current_form.id, page.id, answer_value: condition_input.answer_value)
       else
         # TODO: Route number is hardcoded whilst we can only have one value for it
         # https://trello.com/c/BfkZEIgM/3446-set-route-count-dynamically-instead-of-hard-coding-it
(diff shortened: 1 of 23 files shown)


lfdebrux (dismissed):
Changes makes sense to me, and work on my machine

app/views/conditions/exit_page/edit.html.erb
lfdebrux:
```suggestion
label: t("helpers.label.conditions_exit_page_input.exit_page_markdown"),
hint: t("helpers.hint.conditions_exit_page_input.exit_page_markdown"),
```

app/views/conditions/exit_page/new.html.erb
```diff
@@ -16,19 +16,19 @@
 
       <%= render MarkdownEditorComponent::View.new(:exit_page_markdown,
         form_builder: f,
-        render_preview_path: exit_page_render_preview_path(form_id: exit_page_input.form.id, page_id: exit_page_input.page.id, check_preview_validation:),
+        render_preview_path: conditions_exit_page_render_preview_path(form_id: exit_page_input.form.id, page_id: exit_page_input.page.id, check_preview_validation:),
         preview_html: preview_html,
         form_model: exit_page_input,
         label: t("helpers.label.pages_exit_page_input.exit_page_markdown"),
         hint: t("helpers.hint.pages_exit_page_input.exit_page_markdown"),
```
lfdebrux:
```suggestion
label: t("helpers.label.conditions_exit_page_input.exit_page_markdown"),
hint: t("helpers.hint.conditions_exit_page_input.exit_page_markdown"),
```

app/views/pages/conditions/confirm_delete_exit_page.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_with(model: delete_exit_page_input, url: update_change_exit_page_path(condition_id: exit_page.id, form_id: @current_form.id, page_id: @page.id, params: { answer_value:, goto_page_id:})) do |f| %>
+    <%= form_with(model: delete_exit_page_input, url: conditions_exit_page_update_path(condition_id: exit_page.id, form_id: @current_form.id, page_id: @page.id, params: { answer_value:, goto_page_id:})) do |f| %>
```
lfdebrux: I think this needs to stay unchanged, because it's the special path for when you're deleting an exit page implicitly by changing the goto page for a condition from the exit page to something else, so it needs to go to the `pages/conditions_controller#update_change_exit_page` action:

https://github.com/govuk-forms/forms-admin/blob/f018457953860a3c09387bb40cb6bc87b265d81c/config/routes.rb#L147

For more context see https://github.com/govuk-forms/forms-admin/blob/f018457953860a3c09387bb40cb6bc87b265d81c/app/controllers/pages/conditions_controller.rb#L59-L62


```suggestion
    <%= form_with(model: delete_exit_page_input, url: update_change_exit_page_path(condition_id: exit_page.id, form_id: @current_form.id, page_id: @page.id, params: { answer_value:, goto_page_id:})) do |f| %>
```

app/controllers/pages/conditions_controller.rb
lfdebrux:
```suggestion
    params.require(:conditions_delete_exit_page_input).permit(:confirm)
```

spec/requests/pages/conditions_controller_spec.rb
lfdebrux:
```suggestion
        params: { conditions_delete_exit_page_input: { confirm: } },
```

lfdebrux (commented):
Another hidden issue, sorry

lfdebrux (approved):

PR #3007, opened 2026-08-05
Add Save and Return feature flag

DavidBiddle:

Trello card: https://trello.com/c/FS2dSAM4/3225-add-save-and-return-feature-flag-in-forms-admin

Adds a new `save_and_return` feature flag.


theseanything (approved):
Thanks for generalising some of this behaviour!

PR #3008, opened 2026-08-06
Fix intermittent Selenium::WebDriver::Error::UnknownError

DavidBiddle:

Trello card: 

We've had a longstanding issue with our Capybara tests intermittently failing with message:
```
Selenium::WebDriver::Error::UnknownError:
       unknown error: unhandled inspector error: {"code":-32000,"message":"Node with given id does not belong to the document"}
```
when using recent versions of Chrome.

This is an instance of https://github.com/teamcapybara/capybara/issues/2800

Previously we resolved this by pinning our CI to an older version of Chrome, then in some repos including this one we added a patch to retry tests when they hit this error.

Subsequently the root cause was identified on the Capybara issue (I believe thanks to an explanation of the error and a test case from @lfdebrux 🙌 ).

This PR:
- adds a flag to the Chrome options which should resolve the issue
- removes the retry patch, since it should no longer be needed


theseanything (approved):

DavidBiddle:
I've tested this locally by running the content of [this feature test](https://github.com/govuk-forms/forms-admin/blob/9b2feda0177de9cf707fd322eefca5e70e476694/spec/features/form/add_or_edit_questions/add_a_question_with_guidance_spec.rb#L14-L22) 300 times without seeing the error (200 headless, 100 with the GUI)

PR #3009, opened 2026-08-06
Generalise feature flag rake tasks and specs

theseanything:

Adding a group-scoped feature flag currently requires a copy-pasted rake task and copy-pasted spec blocks on top of the migration, settings entry and locale label. This PR replaces the per-flag rake tasks with a single parameterised task and the per-flag spec assertions with consistency checks, so a new flag only needs the migration, a `config/settings.yml` entry and an `en.yml` label.

### DIFF
--- a/lib/tasks/groups.rake
+++ b/lib/tasks/groups.rake
@@ -35,25 +35,20 @@ namespace :groups do
     end
   end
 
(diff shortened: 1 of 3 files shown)


DavidBiddle (approved):
All works for me locally and the error/usage messages are all nice and clear 🎉

PR #3010, opened 2026-08-06
Add multiple branch exit pages

thomasiles:

Trello card: https://trello.com/c/gecvdskV/3173-change-add-edit-delete-exit-page-to-work-with-multiple-branches

This PR adds a new exit page controller, input object and view to add exit pages under the multiple branches flag.

Currently there are no guards to adding exit pages to pages. I think we probably want to prevent exit pages based on the question type. I think we can add that later though. It also doesn't add update and delete which I'll do in other PRs.

There is no link from the new routes page (or anywhere else). To check the new page use the URL directly, for example:
http://localhost:3000/forms/6/pages/23/exit-pages/new

image


hannahkc:
content looks good to me

SamJamCul (approved):

PR #3012, opened 2026-08-10
Exit pages on page list view

SamJamCul:

When conditions lead to exit pages, this is displayed on the page list
view.

The order of conditions being displayed is:
 - Conditions that go to another page, in order of the pages they go to
 - Conditions that skip to the end
 - Conditions that go to an exit page, in order of the exit pages'
   created_at values

Within each condition grouping, the answer values are displayed in the
order that they appear in the question.

https://trello.com/c/ul4Ot2Sn/3174-change-add-and-edit-your-questions-page-to-show-multiple-exit-pages

The image below shows a question which has:
 - A route to another page
 - A route that skips to the end
 - A route that goes to the exit page "I have a special passport"
 - Two routes that both go to the exit page "You should get a passport"

image

### DIFF
--- a/app/components/page_list_component/view.html.erb
+++ b/app/components/page_list_component/view.html.erb
@@ -49,10 +49,10 @@
               <% end %>
 
                 <% if page.routing_conditions.first.answer_value.present? %>
-                  <% answer_value_groups(page).each do |_, group| %>
+                  <% answer_value_groups(page).each do |group| %>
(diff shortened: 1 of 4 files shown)


hannahkc:
Content looks good on the screenshot.

thomasiles (approved):
looks good to me, nice one

PR #3014, opened 2026-08-11
Add version to form documents

stephencdaly:

Add a version column to FormDocuments to support having multiple immutable versions of FormDocuments. Create new live form documents with a version number of 1, and add a rake task to backfill existing live/archived form documents to be version 1, as we can still currently only have one live/archived version per language per form.

A future PR will make it so that we can have multiple versions and we increment the version number when a draft is made live.

Also adds a `latest_form_document` foreign key relation to forms to make it easiest to get the latest version for constructing feature reports.


lfdebrux (approved):
Works for me, I've tested this locally.

PR #3017, opened 2026-08-12
Add exit page position methods

SamJamCul:
We need to occasionally list exit pages in their created at order, and rather than having to work out each time which position the exit page occupies in that order, this method can do that.

Trello card: 

### DIFF
--- a/app/models/exit_page.rb
+++ b/app/models/exit_page.rb
@@ -15,4 +15,14 @@ class ExitPage < ApplicationRecord
       "markdown" => markdown,
     }
   end
+
+  def position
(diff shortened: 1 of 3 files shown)


app/models/exit_page.rb
```diff
@@ -15,4 +15,14 @@ def as_form_document_exit_page
       "markdown" => markdown,
     }
   end
+
+  def exit_page_position
```
lfdebrux: This can probably just be called `position`?


```suggestion
  def position
```

SamJamCul: Good point!

thomasiles (approved):

PR #3020, opened 2026-08-13
Return latest form document version from the API

stephencdaly:

Change the indexes on the form_documents table to allow for multiple versions of a form to exist.

Make the forms API consumed by forms-runner handle a form having multiple form document versions. It works as follows:

- If the tag param is “draft”, return the only draft form document
- If the tag param is “live” return the form document if the latest version is live. Return a 404 if the latest version is archived
- If the tag param is “archived” return the form document if the latest version is archived. Return a 404 if the latest version is live

It will return the latest form document for the requested language - even if there is a higher version number for another language. This should handle the situation where a Welsh form document has been archived and a new version of the English form has been created. Forms runner handles this by rendering a page saying the form was archived with a link to the English form.

### DIFF
--- a/app/controllers/api/form_documents_controller.rb
+++ b/app/controllers/api/form_documents_controller.rb
@@ -18,6 +18,8 @@ private
 
     raise NotFoundError unless latest_form_document&.tag == tag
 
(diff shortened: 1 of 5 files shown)


lfdebrux (approved):
Looks good to me

PR #3021, opened 2026-08-13
Change unique indexes for form documents to allow for multiple versions

stephencdaly:

Update the unique indexes for form documents so that:
- There can only be form document with a given version per language per form
- There can only be one form document with a null version per language per form, which accounts for draft form documents

A migration has already been run on all environments to set the version for live and archived form documents to 1


lfdebrux (dismissed):

stephencdaly:
@theseanything I've updated to not add the indexes concurrently and instead use a transaction, as discussed

theseanything (approved):

PR #3028, opened 2026-08-18
Replace form document associations

stephencdaly:

Replace the associations on the Form model to get live/archived the FormDocuments by tag to instead get the FormDocument with the highest version number.

### DIFF
--- a/app/controllers/forms/submission_email_controller.rb
+++ b/app/controllers/forms/submission_email_controller.rb
@@ -59,11 +59,7 @@ module Forms
     def live_submission_email_updated?
       return false unless current_form.is_live?
 
-      current_live_form.submission_email != current_form.submission_email
-    end
-
-    def current_live_form
-      @current_live_form ||= FormDocument::Content.from_form_document(current_form.latest_form_document)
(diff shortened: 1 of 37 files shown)


app/controllers/forms/submission_email_controller.rb
```diff
@@ -59,11 +59,11 @@ def submission_email_input
     def live_submission_email_updated?
       return false unless current_form.is_live?
 
-      current_live_form.submission_email != current_form.submission_email
+      current_live_or_archived_form.submission_email != current_form.submission_email
     end
 
-    def current_live_form
-      @current_live_form ||= FormDocument::Content.from_form_document(current_form.live_form_document)
+    def current_live_or_archived_form
```
theseanything: Isn't this already inherited from FormsController?

stephencdaly: Oh yeh good spot

stephencdaly: updated

theseanything (approved):

PR #3043, opened 2026-08-19
Refactor RouteInput#goto

thomasiles:
### Refactor RouteInput to make adding exit page routes simpler

Trello card: https://trello.com/c/GQ6Qdn8N/3211-add-exit-pages-to-the-edit-question-routes-page-new-style-exit-pages-only
 Currently, routes can goto the following destinations:
- A page, represented by the page_id
- The end of form, using "end_of_form"
- The default, which is the next page in the form, represented with "default"

We currently use the goto attribute of RouteInput, which is a string.

We are adding exit pages as destinations for routes. We'll need a way to distinguish routes to exit pages from pages. My first approach was to mark exit pages with a prefix. This would mean they appear in select boxes as "exit_123" where 123 is the id of the exit page.

This works but leaves us matching strings in different parts of the code to check what goto is referring to. We need to do this when creating the list of options for a route, parsing the form and any other checks where we need to check what kind of route it is. 

This refactor tries to push the parsing/serialising logic to the edges of the application with an activemodel type. Using value objects for the GotoValue allows us to deal only with value objects in most parts of the code and let Rails handle the parsing for us.

We are using case statements and .is_a, which can be a code smell. Because we are using Data.define and Value objects, I think it's ok. The deafult and of end_of_form values don't hold any data. Should they be singleton classes so we don't need to call .new?

This PR is broken into commits. It's mainly changing the format in different parts of the code which relied on it.

### DIFF
--- a/app/input_objects/forms/route_input.rb
+++ b/app/input_objects/forms/route_input.rb
@@ -1,12 +1,9 @@
 class Forms::RouteInput < BaseInput
   include ActiveModel::Attributes
 
-  END_OF_FORM_VALUE = "end_of_form".freeze
-  DEFAULT_VALUE = "default".freeze
-
   attribute :id # id of the Condition
   attribute :page_id, :integer
-  attribute :goto
+  attribute :goto, GotoValueType.new
   attribute :answer_value
 
   attr_accessor :page, :goto_page, :goto_options
@@ -14,20 +11,25 @@ class Forms::RouteInput < BaseInput
   validate :route_is_not_backwards
 
   def goes_to_default_next_page?
-    goto == DEFAULT_VALUE
+    goto.is_a?(GotoValue::DefaultValue)
   end
 
   def goes_to_end_of_form?
-    goto == END_OF_FORM_VALUE
+    goto.is_a?(GotoValue::EndOfFormValue)
+  end
+
+  def goes_to_page?
+    goto.is_a?(GotoValue::Page)
   end
 
   def condition_attributes
-    if goes_to_end_of_form?
-      { goto_page_id: nil, skip_to_end: true, check_page_id: page.id }
-    elsif goes_to_default_next_page?
+    case goto
+    when GotoValue::DefaultValue
       nil
-    else
-      { goto_page_id: goto, skip_to_end: false, check_page_id: page.id }
+    when GotoValue::EndOfFormValue
+      { goto_page_id: nil, skip_to_end: true, check_page_id: page.id }
+    when GotoValue::Page
+      { goto_page_id: goto.page_id, skip_to_end: false, check_page_id: page.id }
     end
   end
 
--- a/app/input_objects/forms/route_input.rb
+++ b/app/input_objects/forms/route_input.rb
@@ -45,15 +45,11 @@ class Forms::RouteInput < BaseInput
 private
 
   def route_is_not_backwards
-    return if skippable_for_backwards_validation?
+    return if !goes_to_page? || goto_page.nil?
 
     return if goto_page.position > page.position
 
     errors.add(:goto, :backwards_route, message: error_message_for_backwards_route)
-  end
-
-  def skippable_for_backwards_validation?
-    goes_to_default_next_page? || goes_to_end_of_form? || goto_page.nil?
   end
 
   def error_message_for_backwards_route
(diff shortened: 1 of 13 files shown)


app/models/goto_value.rb
```diff
@@ -0,0 +1,30 @@
...
+      case self
+      when DefaultValue
+        DEFAULT_STRING
+      when EndOfFormValue
+        END_OF_FORM_STRING
+      when Page
+        "#{PAGE_PREFIX}#{page_id}"
+      end
+    end
+  end
```
lfdebrux: I like the `GotoValue` idea and the use of `Data.define` very much, however the way `#to_s` is defined here using types that are defined later (that then include the definition of `#to_s`) seems a bit recursive... Why not define the `#to_s` method on each data class [[1](https://docs.ruby-lang.org/en/3.4/Data.html#method-c-define:~:text=%3A%3Adefine%20method%20accepts%20an%20optional%20block%20and%20evaluates%20it%20in%20the%20context%20of%20the%20newly%20defined%20class)]? Maybe you were going for a bit more of a functional style rather than an object-oriented style...?

app/input_objects/forms/route_input.rb
```diff
@@ -1,33 +1,35 @@
...
+    goto.is_a?(GotoValue::Page)
   end
 
   def condition_attributes
-    if goes_to_end_of_form?
-      { goto_page_id: nil, skip_to_end: true, check_page_id: page.id }
-    elsif goes_to_default_next_page?
+    case goto
+    when GotoValue::DefaultValue
       nil
```
lfdebrux: It might be worth making the case when `goto` is `nil` explicit?


```suggestion
    when nil, GotoValue::DefaultValue
      nil
```

app/models/goto_value.rb
```diff
@@ -0,0 +1,30 @@
+module GotoValue
```
lfdebrux, line 1: Is it worth putting the `GotoValue` code in the `Routes` namespace? I'm not sure whether it's currently super obvious what it's for.

app/services/routes/build_service.rb
```diff
@@ -31,13 +31,14 @@ def options_for_goto_page(page, selected = nil)
     # If there's no next page then it's the last page of the form.
     # We only need options for this page if there's an error that needs correcting.
     if next_page.nil?
-      selected_page = if selected && selected != Forms::RouteInput::DEFAULT_VALUE
-                        option_for_select(form.pages.find { |page| page.id == selected })
+      selected_page = if selected.is_a?(GotoValue::Page)
+                        matching_page = form.pages.find { |p| selected.page_id == p.id }
+                        option_for_select(matching_page) if matching_page
```
lfdebrux: Previously if the page wasn't found this `option_for_select` would raise an exception (because it would try and do `nil.position`), but now the block will fail silently and selected_page will be `nil`. Is that okay?

app/services/routes/build_service.rb
```diff
@@ -93,20 +94,21 @@ def build_route_for_generic_page(page, conditions_by_key)
     key = [page.id, nil]
     condition = conditions_by_key[key]
 
+    selected = GotoValue::Page.new(condition&.goto_page_id) if condition&.goto_page_id.present?
```
lfdebrux: Could we use the `EndOfFormValue` here to be more explicit?

app/services/routes/build_service.rb
lfdebrux: I think this refactor makes sense, but it took me a while to work it out. I might have been worth having it in a separate commit.

app/models/goto_value.rb
thomasiles: You are right! I shifted the code around, it was a separate method before. I think it makes sense to get rid of the module and includes and also move the parsing code into a class method. That keeps it all together and is a bit less mysterious.

lfdebrux: That makes sense to me. Thanks!

lfdebrux (approved):
I like the refactor, and the changes work on my machine. One small additional comment about naming, but otherwise looks good to me.

app/models/goto_value.rb
thomasiles: Yes, I think that does make sense. I think the route inputs can be put there too so I'll do it in a later PR.

PR #3045, opened 2026-08-20
Add exit page routes

thomasiles:

Trello card: https://trello.com/c/GQ6Qdn8N/3211-add-exit-pages-to-the-edit-question-routes-page-new-style-exit-pages-only

This PR adds exit pages as an option to send routes to.

To do that, we add a new GotoValue to represent exit pages, define how to convert them from and to conditions and add them to the list of options for pages.

We also add a new exit page button and a list of links to edit the existing exit pages for a page.

### DIFF
--- a/app/services/routes/build_service.rb
+++ b/app/services/routes/build_service.rb
@@ -36,10 +36,10 @@ class Routes::BuildService
                         option_for_select(matching_page) if matching_page
                       end
 
-      return [
+      return (exit_pages(page) + [
         selected_page,
         [END_OF_FORM_OPTION.first, GotoValue::DefaultValue.new],
-      ].compact
+      ]).compact
     end
 
     # Don't include the current page or pages before in the options,
@@ -49,7 +49,7 @@ class Routes::BuildService
     next_page_id = next_page.id
     drop = true
 
-    all_goto_options.filter_map do |option|
+    options = all_goto_options.filter_map do |option|
       _, value = option
 
       if drop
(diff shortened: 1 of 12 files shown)


app/services/routes/build_service.rb
```diff
@@ -64,6 +64,14 @@ def options_for_goto_page(page, selected = nil)
         option
       end
     end
+
+    exit_pages(page) + options
+  end
+
+  def exit_pages(page)
+    page.exit_pages.map do |exit_page|
+      ["Exit page #{exit_page.position}. #{exit_page.heading}", GotoValue::ExitPage.new(exit_page.id)]
```
lfdebrux: In the designs the exit page number is separated from the heading by a colon:

Image

https://app.mural.co/t/gaap0347/m/gaap0347/1777554146285/b4db73fa35cee565fab61bb4ca050a371393e52f?wid=0-1782227724132

app/services/routes/build_service.rb
```diff
@@ -64,6 +64,14 @@ def options_for_goto_page(page, selected = nil)
         option
       end
     end
+
+    exit_pages(page) + options
+  end
+
+  def exit_pages(page)
+    page.exit_pages.map do |exit_page|
```
lfdebrux: It might be better to use `ExitPage.positions_for_page` here to avoid N + 1 database queries.

app/services/routes/build_service.rb
lfdebrux: (It's the same format as for the `.exit_page_list_entry` translation key)

thomasiles: good spot!, I've changed it over

app/services/routes/build_service.rb
thomasiles: yes, that makes sense. i've switched it over and made sure they use the same ordering too

lfdebrux (approved):

PR #3049, opened 2026-08-21
Add UI for adding/removing Organisation Domains

chao-xian:

Trello card: https://trello.com/c/BFiR2guc

Adds UI for managing Organisation Domains. We only need to manage adding and deleting domains.

localhost_3000_organisations_4

localhost_3000_organisations_4_domains_new

localhost_3000_organisations_4 (1)

### DIFF
--- a/app/controllers/organisation_domains_controller.rb
+++ b/app/controllers/organisation_domains_controller.rb
@@ -19,13 +19,47 @@ class OrganisationDomainsController < WebController
     end
   end
 
+  def delete
+    authorize organisation, :can_manage_organisation_domains?
+
+    @organisation_domain = organisation_domain
+    @delete_confirmation_input = Organisations::DeleteConfirmationInput.new
+  end
+
+  def destroy
+    authorize organisation, :can_manage_organisation_domains?
+
+    @organisation_domain = organisation_domain
+    @delete_confirmation_input = Organisations::DeleteConfirmationInput.new(delete_confirmation_input_params)
+
+    unless @delete_confirmation_input.valid?
+      return render :delete, status: :unprocessable_entity
+    end
+
+    unless @delete_confirmation_input.confirmed?
+      return redirect_to organisation_path(organisation), status: :see_other
+    end
+
+    @organisation_domain.destroy!
+
+    redirect_to organisation_path(organisation), success: "#{@organisation_domain.domain} has been removed from this organisation", status: :see_other
+  end
+
 private
 
   def organisation
     @organisation ||= Organisation.find(params[:organisation_id])
   end
 
+  def organisation_domain
+    @organisation_domain ||= organisation.organisation_domains.find(params[:id])
+  end
+
   def organisation_input_params
     params.require(:organisations_domain_input).permit(:domain).merge(organisation:)
+  end
+
+  def delete_confirmation_input_params
+    params.require(:organisations_delete_confirmation_input).permit(:confirm)
   end
 end
--- a/app/controllers/organisation_domains_controller.rb
+++ b/app/controllers/organisation_domains_controller.rb
@@ -1,0 +1,31 @@
+class OrganisationDomainsController < WebController
+  after_action :verify_authorized
+
+  def new
+    authorize organisation, :can_manage_organisation_domains?
+
+    @domain_input = Organisations::DomainInput.new(organisation:)
+  end
+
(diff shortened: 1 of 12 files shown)


app/input_objects/organisations/domain_input.rb
```diff
@@ -0,0 +1,14 @@
+module Organisations
+  class DomainInput < BaseInput
+    attr_accessor :organisation, :domain
+
+    validates :organisation, presence: true
+    validates :domain, presence: true
```
stephencdaly: Can we validate it's a valid domain to show the error summary if it's invalid?

```suggestion
    validates :domain, presence: true, domain: true
```

app/controllers/organisation_domains_controller.rb
```diff
@@ -0,0 +1,29 @@
+class OrganisationDomainsController < WebController
```
stephencdaly, line 1: can we add 

```
after_action :verify_authorized
```

to verify all actions are authorised?

app/views/organisation_domains/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<% set_page_title(title_with_error_prefix(t(".title"), @domain_input.errors.any?)) %>
+<% content_for :back_link, govuk_back_link_to(organisation_path(@domain_input.organisation)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with(model: @domain_input, url: organisation_domains_path(@domain_input.organisation)) do |f| %>
+      <% if @domain_input.errors.any? %>
+        <%= f.govuk_errors_summary %>
```
stephencdaly: This is raising an exception when there is an error

```suggestion
        <%= f.govuk_error_summary %>
```

app/views/organisation_domains/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+      <% end %>
+
+      <h1 class="govuk-heading-l" id="heading">
+        <span class="govuk-caption-l"><%= @domain_input.organisation.name_with_abbreviation %></span>
+        <span class="govuk-visually-hidden"> - </span>
+        <%= t(".title") %>
+      </h1>
+
+      <%= f.govuk_text_field :domain, label: { text: t('.label'), size: "m" } %>
```
stephencdaly, line 17: I wonder whether it's useful to have a hint about what it should look like - that it shouldn't have the `@` sign in it? Although maybe the error message will be clear enough when we've added validation to the input.

app/views/organisations/show.html.erb
```diff
@@ -102,11 +102,27 @@
...
+              row.with_cell(header: true, text: t("organisations.show.domains.table_headings.actions"))
+            end
+          end %>
+
+          <%= table.with_body do |body| %>
+            <% @organisation.organisation_domains.each do |organisation_domain| %>
+              <%= body.with_row do |row| %>
+                <%= row.with_cell(text: organisation_domain.domain) %>
+                <%= row.with_cell do %>
+                  <%= govuk_button_link_to(t("organisations.show.domains.remove"), delete_organisation_domain_path(@organisation, organisation_domain), secondary: true, visually_hidden_suffix: organisation_domain.domain, class: "govuk-!-margin-bottom-0") %>
```
stephencdaly: I wonder whether this should be styled as a normal link because it's linking to another page rather than instantly performing the action?

stephencdaly (changes requested):
Looks good and mainly works as expected, just spotted a few things

app/views/organisation_domains/new.html.erb
chao-xian: @stephencdaly  I think the error message would suffice, if I amended the current `Enter a domain name in the correct format, like subdomain.gov.uk` to something like `Enter a domain name in the correct format, like subdomain.gov.uk without a leading @`?

chao-xian: Screenshot 2026-08-25 at 11 27 47

chao-xian: @stephencdaly I think the existing example is good enough...

stephencdaly: Yep looks good

stephencdaly (approved):

PR #3052, opened 2026-08-25
Multiple exit page translations

SamJamCul:

https://trello.com/c/Qe1U2bBO/3181-change-welsh-translation-feature-to-work-with-multiple-exit-pages

Adds the alternative way to render Exit Pages on the Welsh translations page when the multiple branches feature flag is enabled. This covers the creation/deletion of multiple exit page translations. 

You can also delete the Welsh translation for a form's exit pages, either by removing the translation and saving the change, or deleting the entire Welsh translation for the form. 

For the technical implementation, I've created some parallel input objects, which have a '2' after their name. At some point in the future we'll want to remove the old input objects and collapse these into being the only input objects - but for now this was a lot easier to do, as it meant only checking the feature flag's state on the top level. 

___

#### A single exit page is listed

image

___
#### Multiple exit pages are listed in sequence, and translations can be added to individual fields independently

image

___

#### It should show errors when invalid fields are submitted

image

### DIFF
--- a/app/input_objects/forms/welsh_page_translation_input2.rb
+++ b/app/input_objects/forms/welsh_page_translation_input2.rb
@@ -1,0 +1,276 @@
+class Forms::WelshPageTranslationInput2 < BaseInput
+  include TextInputHelper
+  include ActionView::Helpers::FormTagHelper
+  include ActiveModel::Attributes
+
+  attr_accessor :exit_page_translations, :selection_options_cy
+  attr_reader :page
+
+  attribute :id
+  attribute :question_text_cy
+  attribute :hint_text_cy
+  attribute :page_heading_cy
+  attribute :guidance_markdown_cy
+  attribute :none_of_the_above_question_cy
+
+  validate :question_text_cy_present?, on: :mark_complete
+  validate :question_text_cy_length, if: -> { question_text_cy.present? }
+
+  validate :hint_text_cy_present?, on: :mark_complete
+  validate :hint_text_cy_length, if: -> { hint_text_cy.present? }
+
+  validate :page_heading_cy_present?, on: :mark_complete
+  validate :page_heading_cy_length, if: -> { page_heading_cy.present? }
+
+  validate :guidance_markdown_cy_present?, on: :mark_complete
+  validate :guidance_markdown_cy_length_and_tags, if: -> { guidance_markdown_cy.present? }
+
+  validate :none_of_the_above_question_cy_present?, on: :mark_complete
+  validate :none_of_the_above_question_cy_length, if: -> { none_of_the_above_question_cy.present? }
+
+  validate :exit_page_translations_valid?
+  validate :selection_options_valid?
+
+  def initialize(attributes = {})
+    @page = attributes.delete(:page) if attributes.key?(:page)
+    super
+    self.id ||= @page&.id
+    @exit_page_translations ||= []
+    @selection_options_cy ||= []
+  end
+
+  def submit
+    return false if invalid?
+
+    page.question_text_cy = question_text_cy
+    page.hint_text_cy = page_has_hint_text? ? hint_text_cy : nil
+    page.page_heading_cy = page_has_page_heading_and_guidance_markdown? ? page_heading_cy : nil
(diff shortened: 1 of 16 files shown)


hannahkc:
I've reviewed the content in the screenshots against the designs and it all looks good to me.

app/input_objects/forms/welsh_page_translation_input2.rb
```diff
@@ -0,0 +1,276 @@
+class Forms::WelshPageTranslationInput2 < BaseInput
```
stephencdaly, line 1: How long do you think we'll need both inputs for? It looks like there's only smallish changes between the 2 versions that could use  conditionals - how much more effort would it be to do that?

I just ask because we're making a few changes to this page too and I'm worried about the overhead of having to make sure we've updated both versions when we're making changes that aren't to do with exit pages.

SamJamCul: We should only need it until the feature launches and the flag is on for everyone. At that point we can disregard the unflagged input objects. 

If that's going to be too far in the future then I can rework this into single input objects

stephencdaly: Ok, just focus on getting this in if the reviewer is happy with it and we'll work around it. We can think again later if its getting too difficult.

config/locales/input_objects/welsh_exit_page_translation.yml
```diff
@@ -0,0 +1,22 @@
...
+              too_long: The exit page content for question %{question_number} cannot be longer than %{count} characters
+              unsupported_markdown_syntax: Exit page content for question %{question_number} can only contain formatting for links, subheadings (##), bulleted lists (*), or numbered lists (1.)
+  helpers:
+    hints:
+      forms_welsh_exit_page_translation_input:
+        markdown_cy: Enter Welsh exit page content
+    label:
+      forms_welsh_exit_page_translation_input:
+        heading_cy_html: Enter Welsh exit page %{exit_page_number} heading<span class="govuk-visually-hidden"> for question %{question_number}</span>
+        markdown_cy_html: Enter Welsh exit page %{exit_page_number} content<span class="govuk-visually-hidden"> for question %{question_number}</span>
```
lfdebrux: Looking at Mural, I think the hidden text should include the exit page number?


```suggestion
        heading_cy_html: Enter Welsh exit page %{exit_page_number} heading<span class="govuk-visually-hidden"> for question %{question_number}’s exit page %{exit_page_number}</span>
        markdown_cy_html: Enter Welsh exit page %{exit_page_number} content<span class="govuk-visually-hidden"> for question %{question_number}’s exit page %{exit_page_number}</span>
```

app/views/forms/welsh_translation/new.html.erb
```diff
@@ -151,50 +151,101 @@
...
-                    label: { text: t("helpers.label.forms_welsh_condition_translation_input.exit_page_heading_cy_html", question_number: page_form.object.page.position), for: condition_form.object.form_field_id(:exit_page_heading_cy) },
-                    lang: "cy"
-                  %>
+                            <%= exit_page_form.govuk_text_field :heading_cy,
+                              id: exit_page_form.object.form_field_id(:heading_cy),
+                              label:  { 
+                                text: t("helpers.label.forms_welsh_exit_page_translation_input.heading_cy_html", exit_page_number: exit_page_form.object.position, question_number: page_form.object.page.position), 
+                                for: exit_page_form.object.form_field_id(:heading_cy),
+                                lang: "cy",
+                              }
```
lfdebrux: I think the `lang` attribute should be on the text field rather than the label for the text field.

```suggestion
                              },
                              lang: "cy",
```

lfdebrux (commented):
How does this work with the old exit page attributes on the conditions?

config/locales/input_objects/welsh_exit_page_translation.yml
SamJamCul: I'll add it in, I wasn't sure as the exit page number is already in the unhidden text

app/views/forms/welsh_translation/new.html.erb
SamJamCul: Good spot, fixed

thomasiles (approved):
This looks good to me 🏅

PR #3067, opened 2026-08-26
Fix markdown length validation for exit pages

thomasiles:
When the char count for the markdown field is too long, the error cannot be displayed because the translations assumed the count field was available.

As the error is checked in the markdown gem and added to the model in the our validator, it doesn't set the count field.

I looked at different ways of making the count available. I don't think it's worth making wider code changes though. Instead, I've hardcoded the 5000 character limit in the translations.

Trello card: 

### DIFF
--- a/config/locales/input_objects/exit_pages.yml
+++ b/config/locales/input_objects/exit_pages.yml
@@ -10,7 +10,7 @@ en:
               too_long: Page heading must be %{count} characters or less
             markdown:
               blank: Enter page content
-              too_long: Page content must be 5000 characters or less
+              too_long: Page content must be 4,999 characters or less
               unsupported_markdown_syntax: Page content can only contain formatting for links, subheadings (##), bulleted lists (*), or numbered lists (1.)
   helpers:
     hint:
--- a/config/locales/input_objects/exit_pages.yml
+++ b/config/locales/input_objects/exit_pages.yml
@@ -10,7 +10,7 @@ en:
               too_long: Page heading must be %{count} characters or less
             markdown:
               blank: Enter page content
-              too_long: Page content must be %{count} characters or less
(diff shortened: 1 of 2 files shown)


config/locales/input_objects/exit_pages.yml
```diff
@@ -10,7 +10,7 @@ en:
               too_long: Page heading must be %{count} characters or less
             markdown:
               blank: Enter page content
-              too_long: Page content must be %{count} characters or less
+              too_long: Page content must be 5000 characters or less
```
DavidBiddle: The [validation in govuk-forms-markdown will fail if there are exactly 5000 characters ](https://github.com/govuk-forms/govuk-forms-markdown/blob/96fc4aa4706a5dfb6598e3fef659152d45790ca1/lib/govuk-forms-markdown/validator.rb#L24) so I think this actually ought to be 4,999 - at least that's what we're using in the various markdown validation fields on the Welsh translation page:

```suggestion
              too_long: Page content must be 4,999 characters or less
```

thomasiles: Thats a really good point! Thanks, I've applied your suggestion.

config/locales/input_objects/update_exit_pages.yml
```diff
@@ -10,7 +10,7 @@ en:
               too_long: Page heading must be %{count} characters or less
             markdown:
               blank: Enter page content
-              too_long: Page content must be %{count} characters or less
+              too_long: Page content must be 5000 characters or less
```
DavidBiddle: Sorry, same for this one I think:
```suggestion
              too_long: Page content must be 4,999 characters or less
```

DavidBiddle (approved):

PR #3068, opened 2026-08-27
Send user research invitation email automatically

theseanything:
When a new user consents to being contacted about user research during onboarding, send them the research sign-up invitation email automatically via GOV.UK Notify. This email is currently sent manually by the user researcher.

The email is only sent when the user's status changes to consented, so re-submitting the form cannot send it twice. Delivery failures are reported to Sentry rather than raised, so a Notify problem does not block the user from completing onboarding.

Tested in dev.


DavidBiddle (approved):
Email sends locally and the content all makes sense to me 🎉

PR #3070, opened 2026-08-28
Update tasks to make merging organisations easier

DavidBiddle:

Trello card: 

Makes a few updates to make merging organisations easier:
- updates the `organisations:merge` task to include moving email domains, which we added to organisations after we originally wrote this task
- adds a new `merge_mous` flag to the `organisations:merge` task, so that we can optionally merge organisations together even if one of them doesn't have an MOU signed (this will be mainly for cases where an organisation that has signed the MOU gets renamed in a Machinery of Government (MoG) change)
- adds new `organisations:open` and `organisations:close` tasks for setting the `closed` status of an organisation manually (this is useful if an organisation is in the middle of a MoG change and both the old organisation and new organisation are still open in the GOV.UK Organisations API - it lets us close the old organisation to stop new users being spread across both organisations).

### DIFF
--- a/lib/tasks/organisations.rake
+++ b/lib/tasks/organisations.rake
@@ -55,13 +55,13 @@ namespace :organisations do
   end
 
   desc "Merge an organisation into another"
-  task :merge, %i[source_organisation_slug target_organisation_slug] => :environment do |task, args|
+  task :merge, %i[source_organisation_slug target_organisation_slug merge_mous] => :environment do |task, args|
     merge_organisations(**args, task:, dry_run: false)
   end
 
   namespace :merge do
     desc "Merge an organisation into another - dry run"
-    task :dry_run, %i[source_organisation_slug target_organisation_slug] => :environment do |task, args|
+    task :dry_run, %i[source_organisation_slug target_organisation_slug merge_mous] => :environment do |task, args|
       merge_organisations(**args, task:, dry_run: true)
     end
   end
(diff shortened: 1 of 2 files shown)


lib/tasks/organisations.rake
```diff
@@ -225,25 +251,38 @@ def merge_organisations(task:, source_organisation_slug: nil, target_organisatio
...
+      Rails.logger.info("#{task.name}: Would move #{users.count} users, #{groups.count} groups, and #{domains.count} domains from #{source_organisation.name} to #{target_organisation.name}") unless merge_mous_boolean
+      Rails.logger.info("#{task.name}: Would move #{users.count} users, #{groups.count} groups, #{domains.count} domains and #{mou_signatures.count} MOU signatures from #{source_organisation.name} to #{target_organisation.name}") if merge_mous_boolean
+
       return
     end
 
-    Rails.logger.info("#{task.name}: Moving #{users.count} users and #{groups.count} groups from #{source_organisation.name} to #{target_organisation.name}")
+    Rails.logger.info("#{task.name}: Moving #{users.count} users, #{groups.count} groups, and #{domains.count} domains from #{source_organisation.name} to #{target_organisation.name}")
+
+    Rails.logger.info("#{task.name}: Moving #{users.count} users, #{groups.count} groups, and #{domains.count} domains from #{source_organisation.name} to #{target_organisation.name}") unless merge_mous_boolean
```
lfdebrux: Is this message meant to be duplicated?

lfdebrux (dismissed):

lib/tasks/organisations.rake
DavidBiddle: No - good spot!

lfdebrux (approved):

PR #3071, opened 2026-08-28
Show exit pages on live pages view

SamJamCul:
This currently shows routes on the live pages view in all cases, for goto pages, skip to end and exit pages.

It also shows the exit pages for those routes for the live form. All exit pages are shown, even if there isn't a corresponding route. 

This does not include live forms with a Welsh version, as that's coming separately.

Trello card: https://trello.com/c/32d9j0e1/3192-change-live-archived-form-questions-pages-to-show-multiple-exit-pages

#### Routes including ones that go to exit pages

image

#### Exit pages section

image

### DIFF
--- a/app/models/form_document/exit_page.rb
+++ b/app/models/form_document/exit_page.rb
@@ -1,0 +1,16 @@
+class FormDocument::ExitPage
+  include ActiveModel::API
+  include ActiveModel::Attributes
+
+  attribute :id, :integer
+  attribute :created_at, :datetime
+  attribute :updated_at, :datetime
+  attribute :question_page_id, :string
+  attribute :heading, :string
+  attribute :markdown, :string
+
+  def initialize(attributes = {})
+    attributes.slice!(*self.class.attribute_names)
+    super
+  end
+end
--- a/app/models/form_document/step.rb
+++ b/app/models/form_document/step.rb
@@ -2,7 +2,7 @@ class FormDocument::Step
   include ActiveModel::API
   include ActiveModel::Attributes
 
-  attr_reader :routing_conditions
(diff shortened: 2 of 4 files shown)


app/views/forms/_made_live_form_pages.html.erb
```diff
@@ -14,7 +14,7 @@
 
   <div class="govuk-grid-column-full">
     <% form_document.steps.each_with_index do |step, index| %>
-      <%= govuk_summary_card(**StepSummaryCardPresenter.call(step:, steps: form_document.steps, welsh_steps: welsh_form_document&.steps).build_card) do %>
+      <%= govuk_summary_card(**StepSummaryCardPresenter.call(step:, steps: form_document.steps, welsh_steps: welsh_form_document&.steps).build_card, classes: "app-summary-card--exit-pages") do %>
```
lfdebrux: We need to make sure that the `app-summary-card` class is on the element, otherwise the text sizes will be inconsistent.

Image

Normally it's added by `**StepSummaryCardPresenter`, but because you're providing the `classes` keyword argument here that is overridden.

You could add `app-summary-card` to the string you provide, but it might be worth adding a `classes` argument to `**StepSummaryCardPresenter` to make it easier to add classes like this.

SamJamCul: Interesting, I must have changed something at some point that's done this, I didn't notice it locally until just now! I'll fix it up

SamJamCul: Ah it's part of me messing around with the CSS, which I ended up not doing yet. I've removed it.

lfdebrux (approved):

PR #3079, opened 2026-09-02
Show existing routes for long selection questions

lfdebrux:

Trello card: https://trello.com/c/R6lOpwZC/3214-show-route-inputs-on-edit-question-routes-page-for-selection-questions-with-more-than-10-options 

Currently, when the multiple branches feature flag is enabled, for selection questions with more than 10 options we don't allow form creators to add routes for those selection options, we just treat it as a generic question. This is because the edit question routes page could get impractically long otherwise.

However, when the multiple feature is not enabled (the default in production), we do allow form creators to add routes to a long selection question, so there are already some forms in production with routes for a selection question with more than 10 options.

To make sure form creators are still able to edit these forms (and that these forms won't be broken by the edit question routes page), we've decided to show the route inputs for existing routes only for a long selection question. This means affected form creators won't be able to add new routes once the feature flag is enabled for their forms, but they will be able to see and edit existing ones.

We hope to iterate on the multiple branches feature in future to allow form creators to add routes to long selection questions without the edit question routes page becoming more difficult to use.

This PR adds a new method `Forms::RoutesInput.routes_type`, that tells the `Routes::BuildService` how to create route inputs for a form. `Routes::BuildService#build_routes` will now create selection option route inputs for a long selection question, but only for existing conditions. The `.routes_for_selection_options?` method is updated to treat either selection questions with 10 options or less or selection questions with more than 10 options and existing conditions (for a selection option) as a route for selection options, so other places where we care about whether the route is for a selection option or not is still correct.

The logic for `.routes_type` is quite complex, because we need to make sure that if a form creator has a long selection question without routes, and then sets the generic catch-all route for it, it doesn't then get treated as a long selection question with selection options routes. I tried different ways of writing out the conditionals, I settled on this being the most readable, but am definitely open to other suggestions!

### DIFF
--- a/app/controllers/pages/exit_pages_controller.rb
+++ b/app/controllers/pages/exit_pages_controller.rb
@@ -91,7 +91,7 @@ private
   end
 
   def check_page_can_have_exit_pages
-    return if Forms::RoutesInput.route_with_selection_options?(page)
+    return if Forms::RoutesInput.can_have_exit_pages?(page)
 
     render "errors/not_found", status: :not_found, formats: :html
   end
--- a/app/input_objects/forms/route_input.rb
+++ b/app/input_objects/forms/route_input.rb
@@ -40,7 +40,7 @@ class Forms::RouteInput < BaseInput
   end
 
   def label_text
-    if Forms::RoutesInput.route_with_selection_options?(page)
+    if selection_options_route?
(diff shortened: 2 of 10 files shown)


thomasiles (dismissed):
Really nicely considered, works well for me locally.

thomasiles (approved):

PR #3084, opened 2026-09-03
Update Welsh CSV service for multiple exit pages

lfdebrux:

Trello card: https://trello.com/c/Qe1U2bBO/3181-change-welsh-translation-feature-to-work-with-multiple-exit-pages 

When the multiple branches feature is released, the Welsh translation CSV should include the number for each exit page in the row heading, so that exit pages can be uniquely identified for a form when there's a page with more than one exit page.


thomasiles (approved):
nice 🏅

PR #3086, opened 2026-09-03
Fix other parts for exit pages

thomasiles:
### Update exit page related code across the app

Trello card: https://trello.com/c/O0iVW7hX/3213-update-code-not-related-to-routing-to-work-with-new-style-exit-pages

This PR modifies a few different parts of the code which need to be modified because of the changes exit pages.

The main reason the code has to change is because pages can now have exit pages which are not connected to conditions.

The database seed has not been updated. This was in the trello card. The current seed has examples for the new style of exit pages and the existing, condition linked, style. I think this makes sense while both are still possible (the multiple branches work is still behind a feature flag). I've created a new trello card to track tasks to complete when the feature is released and the flag remove.

### DIFF
--- a/app/services/form_copy_service.rb
+++ b/app/services/form_copy_service.rb
@@ -14,6 +14,7 @@ class FormCopyService
     @form = form
     @copied_form = Form.new
     @logged_in_user = logged_in_user
+    @exit_page_id_mapping = {}
   end
 
   def copy(tag: "draft")
@@ -28,6 +29,7 @@ class FormCopyService
       prepend_name_for_language(:en)
 
       copy_pages(content["steps"])
+      copy_exit_pages(content["steps"])
       copy_routing_conditions(content["steps"])
 
       @copied_form.copied_from_id = @form.id
@@ -68,6 +70,23 @@ private
       page = @copied_form.pages.build
       copy_page_attributes(page, step)
       page.save!
+    end
+  end
+
+  def copy_exit_pages(steps)
+    return if steps.blank?
+
+    steps.each_with_index do |step, index|
+      page = @copied_form.pages[index]
+
+      (step["exit_pages"] || []).each do |ep_data|
+        exit_page = page.exit_pages.build(
+          heading: ep_data["heading"],
+          markdown: ep_data["markdown"],
+        )
+        exit_page.save!
+        @exit_page_id_mapping[ep_data["id"]] = exit_page
+      end
     end
   end
 
@@ -139,6 +158,10 @@ private
       exit_page_markdown: condition_data["exit_page_markdown"],
     )
 
+    if condition_data["exit_page_id"].present?
+      condition.exit_page = @exit_page_id_mapping[condition_data["exit_page_id"]]
+    end
+
     condition.save!
   end
 
@@ -161,6 +184,7 @@ private
         page.send("#{attr}=", data[attr]) if data[attr].present?
       end
       copy_welsh_exit_page_conditions(step, page)
+      copy_welsh_exit_pages(step)
       page.save!(validate: false)
     end
   end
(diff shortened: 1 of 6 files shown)


app/services/revert_draft_form_service.rb
```diff
@@ -117,10 +139,18 @@ def revert_routing_conditions(steps_data)
   def assign_condition_attributes(condition, condition_data)
     condition.answer_value = condition_data["answer_value"]
     condition.routing_page = Page.find_by!(external_id: condition_data["routing_page_id"]) if condition_data["routing_page_id"]
```
lfdebrux: Might be worth reformatting this line to match the following two which you've also reformatted.

spec/services/revert_draft_form_service_spec.rb
```diff
@@ -200,6 +200,122 @@ def revert_draft(tag)
...
+          check_page_id: live_form.pages.first.id,
+          exit_page_id: exit_page.id,
+        )
+        live_form.latest_form_document.update!(content: live_form.as_form_document(live_at: live_form.updated_at))
+      end
+
+      context "when the exit page is removed from the condition in the draft" do
+        before do
+          condition = live_form.pages.first.routing_conditions.first
+          condition.update!(goto_page_id: live_form.pages.second.id)
```
lfdebrux: Should probably unset the `exit_page_id` as well

```suggestion
          condition.update!(goto_page_id: live_form.pages.second.id, exit_page_id: nil)
```

spec/services/revert_draft_form_service_spec.rb
```diff
@@ -200,6 +200,122 @@ def revert_draft(tag)
...
+          expect(condition.exit_page.heading).to eq("Heading")
+          expect(condition.exit_page.markdown).to eq("Markdown")
+          expect(condition.goto_page).to be_nil
+        end
+      end
+
+      context "when the exit page is removed from the condition and the exit page is removed" do
+        before do
+          condition = live_form.pages.first.routing_conditions.first
+          condition.update!(goto_page_id: live_form.pages.second.id)
```
lfdebrux:
```suggestion
          condition.update!(goto_page_id: live_form.pages.second.id, exit_page_id: nil)
```

app/services/form_copy_service.rb
```diff
@@ -71,6 +73,24 @@ def copy_pages(steps)
     end
   end
 
+  def copy_exit_pages(steps)
+    return if steps.blank?
+
+    steps.each_with_index do |step, index|
+      page = @copied_form.pages[index]
+      next if page.blank?
```
lfdebrux: `page` shouldn't ever be blank here, right? Do we need this line?

```suggestion
```

app/services/form_copy_service.rb
```diff
@@ -135,10 +155,15 @@ def copy_condition(condition_data, page_id_mapping)
...
-      exit_page_heading: condition_data["exit_page_heading"],
-      exit_page_markdown: condition_data["exit_page_markdown"],
     )
 
+    if condition_data["exit_page_id"].present?
+      condition.exit_page = @exit_page_id_mapping[condition_data["exit_page_id"]]
+    else
+      condition.exit_page_heading = condition_data["exit_page_heading"]
+      condition.exit_page_markdown = condition_data["exit_page_markdown"]
+    end
```
lfdebrux, line 163: It might be safer to always copy `exit_page_heading` and `exit_page_markdown` to the condition for now, what do you think?

app/services/form_copy_service.rb
```diff
@@ -177,4 +203,20 @@ def copy_welsh_exit_page_conditions(step, page)
...
+
+  def copy_welsh_exit_pages(step)
+    # Exit pages already linked to a condition are handled by copy_welsh_exit_page_conditions
+    # Here we are only copying exit pages that are not linked to a condition.
+    condition_exit_page_ids = (step["routing_conditions"] || []).filter_map { |c| c["exit_page_id"] }
+    standalone_exit_pages = (step["exit_pages"] || []).reject { |ep| condition_exit_page_ids.include?(ep["id"]) }
+
+    standalone_exit_pages.each do |ep_data|
+      exit_page = @exit_page_id_mapping[ep_data["id"]]
+      next unless exit_page
```
lfdebrux: It feels like `exit_page` should never be falsey here, so I don't think we want this line

```suggestion
```

app/services/form_copy_service.rb
thomasiles: yes, I added these to match other parts of the code. I don't think we should ever hit this

app/services/form_copy_service.rb
thomasiles: yes, thats a good point. I thought they needed to be separate cases but it's probably better to match the copied form as closely as possible to avoid any hard to debug differences.

lfdebrux (approved):

PR #3087, opened 2026-09-03
Add welsh translations upload

stephencdaly:

Trello card: https://trello.com/c/35HPJgw3/3247-allow-uploading-a-csv-of-welsh-translations

Adds a page to upload a CSV of Welsh translations. This accepts a CSV in the same format as the downloaded CSV, and uses the content in the "Welsh content" column to populate the fields on the Welsh translation page. These translations are only saved when the translations page is submitted so we can perform individual validation on the fields and allow the user to correct any errors.

This PR does not add a link to the page to upload the translations, but it can be accessed for testing at `/forms/:form_id/welsh-translation-upload`.

Things that still need to be done in follow-up PRs:
- Add more validation to the CSV when it is uploaded. We plan to validate the rows present in the CSV to ensure they match the current form to account for questions changing or being re-ordered to ensure we don't apply the translations in the wrong places.
- Adjust the WAF rules to allow larger files to be uploaded.

Screenshot 2026-09-07 at 10 06 17

Screenshot 2026-09-07 at 10 06 50

Screenshot 2026-09-04 at 12 41 17

### DIFF
--- a/spec/requests/forms/welsh_translation_controller_spec.rb
+++ b/spec/requests/forms/welsh_translation_controller_spec.rb
@@ -486,6 +486,7 @@ RSpec.describe Forms::WelshTranslationController, type: :request do
         it "renders the new template" do
           expect(response).to have_http_status(:ok)
           expect(response).to render_template(:new)
+          expect(response.body).to include(I18n.t("banner.success.form.welsh_translation_uploaded"))
         end
 
         it "pre-populates fields from CSV" do
@@ -531,6 +532,7 @@ RSpec.describe Forms::WelshTranslationController, type: :request do
         it "renders the new template" do
           expect(response).to have_http_status(:ok)
           expect(response).to render_template(:new)
+          expect(response.body).to include(I18n.t("banner.success.form.welsh_translation_uploaded"))
         end
 
         it "pre-populates fields from CSV" do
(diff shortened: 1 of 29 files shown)


spec/requests/forms/welsh_translation_controller_spec.rb
```diff
@@ -46,8 +46,8 @@
           expect {
             post(welsh_translation_create_path(id), params:)
           }.to change { form.reload.welsh_completed }.to(true)
-          .and change { form.pages.first.reload.question_text_cy }.to("Ydych chi'n adnewyddu trwydded?")
-          .and change { condition.reload.exit_page_heading_cy }.to("Nid ydych yn gymwys")
+                                                     .and change { form.pages.first.reload.question_text_cy }.to("Ydych chi'n adnewyddu trwydded?")
+                                                                                                             .and change { condition.reload.exit_page_heading_cy }.to("Nid ydych yn gymwys")
```
DavidBiddle: Out of interest, do you know where these spacing changes came from - were they automatic? I don't get them locally from rubocop

stephencdaly: Ah oops, it will have been from my IDE auto-formatting, I can revert any unnecessary ones. Still haven't worked out a setting to stop it

DavidBiddle: I don't think they're a problem, was mainly wondering if there was something missing in my local setup - these tests still seem perfectly readable

DavidBiddle (dismissed):

spec/requests/forms/welsh_translation_controller_spec.rb
stephencdaly: I've reverted these changes. I found a setting to make the indentation less severe in my IDE but not stop it completely.

DavidBiddle (approved):

PR #3101, opened 2026-09-10
Add save and return flag column

chao-xian:

Trello card: https://trello.com/c/6SzUYNnI/3271-add-a-database-column-for-whether-save-and-return-is-enabled-on-a-form

To support the feature we need to store when an editor denotes that the form they are creating will allow users to save and return to their submission. 

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -38,10 +38,16 @@ class Form < ApplicationRecord
     enabled: "enabled",
   }, prefix: :send_copy_of_answers
 
+  enum :save_and_return, {
+    disabled: "disabled",
+    enabled: "enabled",
+  }, prefix: :save_and_return
+
   validates :name, presence: true
   validates :payment_url, url: true, allow_blank: true
   validate :marking_complete_with_errors
   validates :send_copy_of_answers, presence: true
+  validates :save_and_return, presence: true
   validates :available_languages, presence: true, inclusion: { in: SUPPORTED_LANGUAGES }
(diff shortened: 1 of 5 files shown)


db/seeds.rb
```diff
@@ -764,6 +764,34 @@
   copy_of_answers_form.set_task_status_service(TaskStatusService.new(form: multiple_branch_form))
   copy_of_answers_form.make_live!
 
+  save_and_return_form = Form.create!(
+    name: "Saved form for returning to",
+    pages: [
+      Page.create(
```
stephencdaly, line 770: Not essential right now, but is it maybe worth adding a second question so we can better test partially filling a form and returning to it?

db/seeds.rb
```diff
@@ -764,6 +764,34 @@
...
+    question_section_completed: true,
+    declaration_markdown: "",
+    declaration_section_completed: true,
+    privacy_policy_url: "https://www.gov.uk/help/privacy-notice",
+    submission_email:,
+    support_email: "your.email+fakedata84701@gmail.com.gov.uk",
+    support_phone: "08000800",
+    what_happens_next_markdown: "Test",
+    share_preview_completed: true,
+    save_and_return: "enabled",
```
stephencdaly, line 789: I think we should we add `delivery_configurations` to this form as without this the form won't get sent to the submission email when it's submitted. See other forms in the seeds for what this looks like.

stephencdaly (dismissed):
Looks good, just some very minor comments about the seed data

stephencdaly (approved):

PR #3122, opened 2026-09-17
Add V3 endpoint for fetching delivery configurations

DavidBiddle:

Trello card: https://trello.com/c/P7bDQZTP/3248-get-submission-delivery-configuration-from-latest-live-version-of-the-form-in-forms-runner

Adds a new endpoint to the V3 API to get the draft or live/archived delivery configurations for a form.

It's at `/api/v3/forms/#{form.id}/delivery-configurations/{state}`, where #{state} is either `draft` or `current`. Calling it with `draft` will return the delivery config from the draft form document, while `current` will return the config for the most recent live or archived form document.


app/controllers/api/v3_form_documents_controller.rb
```diff
@@ -53,4 +59,20 @@ def language
   def form_id
     params.require(:form_id)
   end
+
+  def state
+    params.require(:state)
+  end
+
+  def draft_delivery_configurations
+    raise NotFoundError unless form.has_draft_version
```
stephencdaly: Is this necessary? You can still preview the draft of a form even if there are no changes so I think we should still be able to get the draft delivery configurations.

config/routes.rb
```diff
@@ -328,11 +328,15 @@
...
-      get "/:version", to: "api/v3_form_documents#show", as: :form_document_version, constraints: { version: /\d+/ }
+    scope "forms/:form_id/" do
+      scope "versions" do
+        get "/live", to: "api/v3_form_documents#live", as: :live_version
+        get "/archived", to: "api/v3_form_documents#archived", as: :archived_version
+        get "/draft", to: "api/v3_form_documents#draft", as: :draft_version
+        get "/:version", to: "api/v3_form_documents#show", as: :form_document_version, constraints: { version: /\d+/ }
+      end
+
+      get "/delivery-configurations/:state", to: "api/v3_form_documents#delivery_configurations", as: :delivery_configurations
```
stephencdaly: should we add a constraint on the `:state`, mainly for documentation purposes?

spec/requests/api/v3_form_documents_controller_spec.rb
```diff
@@ -196,4 +196,161 @@
...
+        form.save_draft!
+      end
+
+      context "when requesting the draft delivery configuration" do
+        before do
+          get "/api/v3/forms/#{form.id}/delivery-configurations/draft", headers:
+        end
+
+        it "returns the draft delivery configuration" do
+          expect(response.body).to eq "[{\"formats\":[],\"delivery_method\":\"email\",\"delivery_schedule\":\"immediate\"}]"
```
stephencdaly: very minor, but is it worth making the draft delivery configurations slightly different so this test is ensuring its returning the draft rather than the live and also the other way round?

stephencdaly (dismissed):

spec/requests/api/v3_form_documents_controller_spec.rb
DavidBiddle: I was in two minds about this - it felt weird to make them different since the draft and live documents should be the same if the form state is `live`. But on reflection I think you're right, and better to have a slightly weird/artificial test than one that doesn't test the behaviour properly. I'll update it

stephencdaly: Ah yeh I can see it's covered by the tests with the live_with_draft state etc. so no need to update

DavidBiddle: Done!

stephencdaly (approved):

